PR #31, opened 2026-05-06
feat: add local_dir backend for container-less instances

BenjaminBenetti:
## Summary
- Introduces a `Backend` strategy implementation that operates directly on a host filesystem directory — no container, VM, or remote workspace.
- Adds `Stateful() bool` to the `Backend` interface so callers can skip `Stop`/`Start` for backends with no lifecycle state. Devcontainer, Coder, and Codespaces return `true`; LocalDir returns `false`.
- Wires the new backend through the factory, TUI dialog metadata, CLI `--backend local_dir` flag, and the `create.Run` git-clone allowlist.

## Behavior

| Method | LocalDir behavior |
|---|---|
| `Up` | Validates wsDir exists; returns `ContainerID = workspaceDir` |
| `Down` / `Stop` / `Start` | No-op (Stateful=false; instanceops refuses Stop/Start) |
| `Exec` / `ExecCommand` | Runs with `cmd.Dir = workspaceDir` |
| `Stats` | `nil, nil` |
| `Logs` / `LogsCommand` | Emits "Local backend has no runtime logs." |
| `CaptureAllSessions` | Runs the shared tmux capture script on the host |
| `AgentToolProbe` | `("", true)` — no agent reported (scoped probe deferred) |
| `EditorURI` | `file:///<workspaceDir>` |
| `ResolveHostname` | `("localhost", true)` — port forwards use the in-process TCP proxy |

## TUI
- `s` (toggle) on a Local Dir instance shows "Local Dir instances cannot be stopped or started" instead of mutating status.
- `backendFor` now lazy-inits `m.backends` defensively (fixed an existing test fragility surfaced by the new code path).

## Test plan
- [ ] CI passes (`go build ./...`, `go test ./...`)
- [ ] Manual: create a Local Dir instance from the TUI, exec a shell, open VS Code via `c`, attempt `s` toggle (expect refusal)
- [ ] Manual: port-forward a local service through the in-process proxy
- [ ] Manual: confirm the existing devcontainer/coder/codespaces flows still work end-to-end

### DIFF
--- a/integration/tests/070_local_dir_up_and_ls.sh
+++ b/integration/tests/070_local_dir_up_and_ls.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# Description: `fleet up --backend local_dir` clones the repo, marks the instance running, and creates no docker container.
+set -euo pipefail
+
+source "$(dirname "$0")/../common.sh"
+itest_begin
+
+setup_test
+
+info "fleet up alpha --backend local_dir"
+"${FLEET_BIN}" up alpha --repo "${FIXTURE_REPO_URL}" --backend local_dir
+
+info "fleet ls"
+ls_out=$("${FLEET_BIN}" ls "${FIXTURE_REPO_NAME}")
+printf '%s\n' "${ls_out}"
+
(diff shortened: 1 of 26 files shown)


BenjaminBenetti:
complexity to high, no longer planned

PR #42, opened 2026-05-08
Restore saved pane groups on WSL

### LINKED ISSUE #41: Pane layout persistence broken on WSL: inner tmux sessions don't survive fleet quit
## Summary

The persistence behavior fixed in #16 regresses on WSL.
`integration/tests/280_tui_pane_layout_persist.sh` passes on bare-metal
Linux but fails reproducibly on Ubuntu-24.04 inside WSL2 (with dockerd
installed in the distro). It's currently skipped on the WSL CI job via
`FLEET_ITEST_SKIP=280_tui_pane_layout_persist` — see
`.github/workflows/integration-windows.yml`.

## Symptom

After building a multi-pane group, quitting fleet cleanly (`q`), and
relaunching, the saved group does **not** reappear under the instance.
On respawn the instance row expands but has zero session children, so
`(N panes)` never shows and the saved layout is lost.

`state.json` is correctly written on quit (asserted in the test), so
the save side works. The break is on restore.

## Root cause (best read of the code)

`internal/tui/splitpane.go:421` —
[`restoreGroupCmd`](https://github.com/BenjaminBenetti/fleet-man/blob/main/internal/tui/splitpane.go)
queries the inner tmux server for sessions matching the group prefix
and bails with `no sessions found for group %s` if the list is empty:

```go
if len(sessions) == 0 {
    return splitPaneMsg{err: fmt.Errorf("no sessions found for group %s", groupID)}
}
```

`internal/tui/app.go:216` — `pruneSavedGroupsForInstance` then deletes
the saved group from both the in-memory map and `state.json` because
the live discovery returned no live sessions for that group ID.

So the entire restart-restore path assumes the inner tmux sessions
inside the devcontainer survive fleet's quit. On Linux they do; under
dockerd-in-WSL they don't appear to — the signal cascade when the
outer tmux pane running `fleet` (or `fleet shell --group`) dies
propagates through `docker exec` and tears down the inner tmux server
along with its sessions.

## Reproduce

```bash
# In WSL (Ubuntu-24.04 with docker-ce + devcontainer CLI):
go build -o /usr/local/bin/fleet ./cmd/fleet
unset FLEET_ITEST_SKIP
FLEET_BIN=/usr/local/bin/fleet bash integration/tests/280_tui_pane_layout_persist.sh
```

Expected: pass (as on Linux).
Actual: timeout waiting for `(2 panes)` after the relaunch.

## Possible fixes (worth thinking through together)

1. **Make `restoreGroupCmd` recreate sessions from saved state when
   none are live**, instead of erroring. The saved group already knows
   the expected session names, pane count, and layout — enough to
   recreate skeleton sessions and have the user reconnect to fresh
   shells. Loses live shell history on WSL but preserves layout.
2. **Fix the signal lifecycle so inner tmux sessions persist** — e.g.
   spawn the container-side tmux server with explicit detachment so it
   survives client SIGHUP (`tmux new-session -d` already does this,
   but maybe the `attach` clients are inheriting a controlling tty
   that propagates differently under WSL's docker exec).
3. **Both** — \(2\) so live sessions actually persist, \(1\) as a
   fallback for genuinely-dead sessions.

## CI status

- Linux job (`integration.yml`): runs this test; required green.
- Windows/WSL job (`integration-windows.yml`): skips this test via
  `FLEET_ITEST_SKIP`; will re-enable once this is fixed.

Yuribane:
Fixes #41.

Summary:
- keep saved pane-group rows visible when WSL reports no live inner tmux sessions
- restore missing group sessions from saved layout state instead of failing immediately
- preserve saved group state when discovery returns an empty live group set, while still deleting saved layouts on explicit group delete
- add FLEET_DEVCONTAINER_UPDATE_REMOTE_USER_UID to pass devcontainer up --update-remote-user-uid-default for WSL/Docker Desktop UID-rewrite failures

Validation:
- go test ./...
- go vet ./...
- git diff --check
- attempted integration/tests/280_tui_pane_layout_persist.sh with FLEET_DEVCONTAINER_BUILDKIT=never and FLEET_DEVCONTAINER_UPDATE_REMOTE_USER_UID=never; local run is blocked by Docker Desktop/MCR EOF while pulling mcr.microsoft.com/devcontainers/base:debian-12

### DIFF
--- a/.github/workflows/integration-windows.yml
+++ b/.github/workflows/integration-windows.yml
@@ -128,14 +128,6 @@ jobs:
           # budget by 3 so legitimately-passing tests aren't killed by the
           # default polling deadlines. Linux job is unaffected.
           export FLEET_TUI_WAIT_SCALE=3
-          # 280_tui_pane_layout_persist exposes a real fleet-on-WSL
-          # behavior gap, not a CI flake — tracked in #41. When fleet
-          # quits under dockerd-in-WSL, the inner tmux sessions inside
-          # the devcontainer do not survive the signal cascade the way
-          # they do on bare Linux; restoreGroupCmd then sees zero live
-          # sessions and the saved layout gets pruned from state. Skip
-          # here until #41 is fixed so the CI gate stays useful.
-          export FLEET_ITEST_SKIP=280_tui_pane_layout_persist
           ./integration/run.sh
 
       - name: Dump dockerd log on failure
--- a/README.md
+++ b/README.md
@@ -102,16 +102,19 @@ npm install -g @devcontainers/cli
 
 On Docker Desktop plus WSL, if devcontainer startup fails while building the
 UID-adjustment or feature image, disable BuildKit for Fleet-managed
-devcontainers:
+devcontainers. If the failure is specifically in `updateUID.Dockerfile`,
+disable the devcontainer CLI's remote user UID rewrite as well:
 
 ```bash
 export FLEET_DEVCONTAINER_BUILDKIT=never
+export FLEET_DEVCONTAINER_UPDATE_REMOTE_USER_UID=never
 ```
 
 To persist that setting:
 
 ```bash
 echo 'export FLEET_DEVCONTAINER_BUILDKIT=never' >> ~/.bashrc
+echo 'export FLEET_DEVCONTAINER_UPDATE_REMOTE_USER_UID=never' >> ~/.bashrc
 ```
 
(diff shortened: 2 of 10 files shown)


BenjaminBenetti (changes requested):
@Yuribane  

Can you remove the skip on the test 

`280_tui_pane_layout_persist`

visible in the run summary? 

https://github.com/BenjaminBenetti/fleet-man/actions/runs/25558203961?pr=42


Your PR fixes this issue, so I want to see the CI pass for it.

Issue #104, opened 2026-06-07
Add fleet_session_list MCP tool to enumerate tmux sessions on an instance


BenjaminBenetti:
## Summary

Add a new MCP tool **`fleet_session_list`** that enumerates the tmux sessions inside a given instance.

## Motivation

The MCP currently exposes three session tools — `fleet_session_spawn`, `fleet_session_exec`, and `fleet_session_read` — but **all three require you to already know the session name**. There is no way to *discover* which sessions exist on an instance:

- `fleet_list` only goes down to the instance level (no sessions).
- The only workaround today is shelling out: `fleet_exec(instance, ["tmux", "ls"])` and parsing the text.

This makes session discovery fragile and forces agents into a raw `tmux ls` fallback.

## Proposed tool

```
fleet_session_list(fleet, instance) -> [
  { session, windows, created_at, attached }
]
```

- **session** — tmux session name (e.g. `hello-world-test~i-spy`)
- **windows** — window count
- **created_at** — session creation time
- **attached** — whether a client is currently attached

## Why a dedicated tool (instead of folding sessions into `fleet_list`)

Considered adding a `sessions[]` array to each instance in `fleet_list`, but rejected it:

1. **Cost/latency** — `fleet_list` is cheap (reads daemon state). Attaching sessions would require a `tmux ls` exec inside *every* container on *every* call (N round-trips).
2. **Failure coupling** — a single unhealthy instance (e.g. a wedged dockerd) could slow or break the entire instance roster.
3. **Cohesion / ISP** — `fleet_list` answers "what instances exist"; sessions are a sub-resource of a single instance, and most callers don't need them.

A dedicated, on-demand `fleet_session_list` keeps `fleet_list` fast and queries sessions only when wanted. (Alternative compromise: an opt-in `fleet_list(include_sessions=true)` flag — but a separate tool is cleaner.)

## Acceptance criteria

- [ ] New MCP tool `fleet_session_list(fleet, instance)` returns the session array above.
- [ ] Returns an empty array (not an error) for a running instance with no sessions.
- [ ] Clear error if the instance is not running.

PR #107, opened 2026-06-07
feat(mcp): add fleet_session_list tool to enumerate tmux sessions (#104)

BenjaminBenetti:
Adds an MCP tool fleet_session_list(fleet, instance) that lists the tmux sessions inside a running instance, returning {session, windows, created_at, attached} per session. This closes the discovery gap: the existing session tools (spawn/exec/read) all required a known session name, leaving a raw `tmux ls` exec as the only way to find sessions.

- On-demand, dedicated tool (not folded into fleet_list) so fleet_list stays a cheap daemon-state read and isn't coupled to a per-container tmux exec.
- No tmux server (no sessions) returns an empty list, not an error (`2>/dev/null || true`); a not-running instance returns a clear error via runningInstance. session_name is placed last in the -F format so a name containing ':' can't shift the fixed fields.

Drive-by fix: session_attached is tmux's CLIENT COUNT, not a 0/1 flag, so the pre-existing parseTmuxSessionsProto (which feeds the TUI's attached indicator) mis-reported any session with 2+ clients as detached. Both parsers now treat any positive count as attached.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,8 @@ An MCP client config (`mcp.json`) can then reference them directly:
 Tools mirror the non-interactive CLI: `fleet_list`, `fleet_status`,
 `fleet_version`, `fleet_logs`, `fleet_up`, `fleet_start`, `fleet_stop`,
 `fleet_down`, `fleet_destroy_fleet`, `fleet_clone`, `fleet_exec`, and the tmux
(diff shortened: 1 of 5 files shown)


BenjaminBenetti:
/qa the provided evidence link is blank! 

File not found
No live file or directory at github/BenjaminBenetti/fleet-man/qa/plans/mcp-server/run-results/2026-06-07T0258Z-b034694.md.

Issue #10, opened 2026-04-18
Build mac binaries in CI


BenjaminBenetti:
Update CI / release process to build mac binaries.

PR #116, opened 2026-06-07
feat(cache): per-fleet deb-package and docker-image cache servers

BenjaminBenetti:
## What

Adds two new per-fleet caching features, modeled on the existing **buildkit** cache — same edit-fleet **Caching** UI (toggle + inline **Delete cache** button), same proto/RPC/settings plumbing, and the same best-effort *never break the instance* contract.

1. **Deb package cache** — a shared `apt-cacher-ng` container per fleet. Each instance gets an apt http-proxy drop-in (`/etc/apt/apt.conf.d/01fleet-proxy`) so repeated `apt install` across the fleet reuse cached `.deb` packages.
2. **Docker image cache** — a shared `registry:2` pull-through mirror of Docker Hub. An instance's *own* dockerd is pointed at it (`registry-mirrors` + `insecure-registries` merged into `/etc/docker/daemon.json`, then a `SIGHUP` reload) so repeated `docker pull` of docker.io images don't re-download.

## How it differs from buildkit

Buildkit shares its server through a **bind-mounted unix socket**. These two are **HTTP services**, so instances reach them over a new **per-fleet user-defined docker network** (`internal/fleetnet`) by container-name DNS — no host-port conflicts, stable across restarts (the default bridge has no DNS). Code shape is otherwise identical to buildkit: host-side `docker` orchestration + an in-instance config write via `ExecCommand`, both best-effort.

## New packages
- `internal/fleetnet` — per-fleet docker network (ensure / connect-instance / remove), idempotent + per-fleet locked.
- `internal/debcache` — apt-cacher-ng host orchestration + in-instance apt-proxy config.
- `internal/imagecache` — registry pull-through host orchestration + in-instance dockerd config.

Wiring mirrors buildkit end to end: `FleetSettings` flags (`deb_cache_server`, `image_cache_server`) + proto, server `DeleteDebCache`/`DeleteImageCache` RPCs + TUI-connect reconcile sweeps, create/clone/start setup, fleet-destroy teardown, and three cache rows in the TUI.

## Scope / limitations (by design)
- **Deb cache** caches **HTTP** apt sources (the default Debian/Ubuntu archives — the bulk of `apt install`). HTTPS sources keep working but go direct/uncached (caching them would require rewriting the instance's apt sources, which is fragile and against the never-break contract).
- **Image cache** only helps **docker-in-docker** instances (only an in-instance dockerd can take a per-fleet `registry-mirror`) and only **docker.io** images (all `registry-mirrors` ever applies to). It silently no-ops on docker-outside-of-docker / no-docker instances, like buildkit skips images without buildx.
- On full fleet destroy the cache containers **and** the shared network are torn down **unconditionally** (idempotent), so toggling a cache off before destroy can't orphan them. The `daemon.json` merge **refuses to clobber** an unparseable file and writes **atomically** (python3, with a jq fallback).

## Testing
- Unit tests for all three new packages (host orchestration via stubbed `docker`, in-instance probe/configure, traversal guards, idempotency, delete-wipes-and-restarts) plus server RPC/reconcile/teardown tests and TUI dialog tests. `make test` (depguard import-boundary + `go test ./...`) and `make lint` pass.
- **End-to-end against a real docker daemon** (this devcontainer is DinD):
  - apt install through the proxy: cold 5.0s → warm 1.4s (3.5×), 9.1M cached, report shows hits.
  - registry pull-through: reachable by container-name DNS; fetched & cached an Alpine manifest from docker.io.
  - `configureScript` exercised in a real shell across all branches (fresh file, python merge preserving existing keys + idempotent, jq fallback) **and** the corruption cases (invalid `daemon.json` is left untouched, script exits non-zero).


BenjaminBenetti:
Thanks for the review — addressed in dea9d8f:

- **`imagecache/instance.go` — `pgrep` dependency (medium):** Fixed. DinD detection (and the SIGHUP reload) no longer rely on `pgrep`; they scan `/proc/<pid>/comm` instead, which is world-readable and needs no `procps`. Verified in a real shell: `ABSENT` with no dockerd, `PRESENT` with a `dockerd`-named process even when `pgrep` is not installed.
- **`server/jobs.go` — unconditional cache teardown (low):** Intentional, kept as-is. Toggling a cache *off* doesn't stop its running container, so gating teardown on the live setting would orphan a container/network the user disabled before destroy. The teardown helpers are idempotent (a missing container/network is a no-op), and there's a regression test (`TestDestroyFleetTearsDownCachesUnconditionally`) plus a code comment explaining it.

BenjaminBenetti:
/qa-manager Thanks — the deb-cache regression you found on `dea9d8f` is fixed in `ef210e6`. Could you re-run the deb-cache plan against it, especially:

- **#9 re-create same-named fleet after destroy:** root cause was `debcache.ensureDir`'s unconditional `os.Chmod(.aptcache, 0777)` returning EPERM after apt-cacher-ng chowned the preserved dir to its own uid. `ensureDir` now tolerates a chmod failure on a *pre-existing* dir (already world-writable from first creation) while still failing on a freshly-created one. Verified locally against real docker: dir chowned to uid 101 → `EnsureSharedServer` recovers and the warm cache is reused. The same tolerance was mirrored into `imagecache`.
- **#10 silent failure:** `state.WriteWarn` now `MkdirAll`s `~/.fleet/logs/` before writing, so a best-effort cache-setup warning is actually produced instead of being lost.

Also FYI: your earlier run on `0dd79ca` reported a 6-instance "only the first instance gets wired" failure with `fleetDebCacheEnabled` as the root cause — that diagnosis doesn't hold (`Load`/`Save`/`Update` share one mutex, so `fleetDebCacheEnabled` and `fleetImageCacheEnabled` read the same locked state and can't disagree). If you can still reproduce multi-instance/fan-out wiring gaps on `ef210e6`, please capture the per-instance `.warn` files (now that they're written) and the create-vs-clone path for the skipped instances — that'll pinpoint it.

PR #117, opened 2026-06-08
ci(gateway): build & publish fleet-gateway image on release

BenjaminBenetti:
## What

Tagging a release now also builds and publishes a **fleet-gateway** Docker image to the GitHub Container Registry, alongside the existing binary release.

The gateway isn't a separate binary — it's the same `fleet` binary run as `fleet gateway` (`internal/cli/gateway.go`, listening on `:443` public / `:8443` control, no `~/.fleet` state). So the image just ships that binary with `fleet gateway` as its entrypoint.

## Changes

- **`deploy/gateway/Dockerfile`** — multi-stage build: cross-compiles a static binary (`CGO_ENABLED=0`, `GOOS/GOARCH` from buildx, built on native `BUILDPLATFORM` to skip QEMU), stamps `internal/version.Version` from the tag exactly like the release binaries, ships on `distroless/static` (root variant so default ports `:443`/`:8443` bind out of the box), `EXPOSE 443 8443`, entrypoint `fleet gateway`.
- **`.dockerignore`** — trims the repo-root build context to what the binary build needs.
- **`.github/workflows/release.yml`** — adds `packages: write` and a `gateway-image` job (no `needs`, runs parallel to the binary release/pages jobs) that pushes multi-arch (`amd64`+`arm64`) to `ghcr.io/<repo>/gateway`. Tags via `docker/metadata-action` (`latest=auto`): stable → `X.Y.Z` / `X.Y` / `latest`; `-beta` → full version only (no `latest`), mirroring the binary release's beta handling. Go version read from `go.mod` like `qa-image.yml`.
- **`README.md`** — documents the published image under *Running a gateway*, and **resolves a stale committed git merge conflict** (the `<<<<<<< Updated upstream` / `Stashed changes` markers straddling the *Security model* / *Remote control (gRPC)* / *Environment Variables* sections). All info retained; the duplicate gRPC subsection was dropped in favor of the complete one, and *Environment Variables* moved to its proper top-level slot.

## Verification

- `CGO_ENABLED=0 go build ./cmd/fleet` succeeds and `fleet gateway --help` shows the expected flags.
- `release.yml` parses cleanly: jobs `[release, pages, gateway-image]`, `packages: write` present.
- No conflict markers remain in `README.md`; heading order is sane.

The multi-arch image build itself runs on the next tag (no push creds locally).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+# Trim the build context for the gateway image (deploy/gateway/Dockerfile, built
+# from the repo root). Only cmd/ + internal/ + go.{mod,sum} are needed to build
+# the binary; everything below is dead weight or shouldn't enter the image build.
+.git
+dist
+_site
+web
+qa
+doc
+readme-image.png
+*.md
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,8 @@ permissions:
   contents: write
   pages: write
   id-token: write
+  # Push the fleet-gateway image to the GitHub Container Registry.
+  packages: write
 
 concurrency:
   group: pages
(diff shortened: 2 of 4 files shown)


BenjaminBenetti:
Thanks for the review — addressed in e393ce8:

**1. Concurrency (medium)** — Fixed. Gave `gateway-image` its own per-ref `concurrency` group with `cancel-in-progress: false` so a cancel between build and push can't leave an incomplete image in ghcr. (The workflow-level `group: pages` already prevented cancellation of the whole run, but it's conceptually the Pages deploy, so an explicit job-scoped guard is clearer and matches `qa-image.yml`.)

**2. VERSION leading `v` (low)** — No change needed; this is correct by design. I pass `VERSION=${{ github.ref_name }}` (`v1.2.3`), and the `release` job stamps the identical value (`version="${GITHUB_REF_NAME}"` → `v1.2.3`) into `internal/version.Version`. So the image binary and the release binaries embed the same version string — which is exactly what the Dockerfile comment claims. The image *tags* strip the leading `v` (via `docker/metadata-action` semver), but the embedded `version.Version` intentionally keeps it to match the binaries.

PR #124, opened 2026-06-10
fix: unify tmux session naming between CLI and TUI

BenjaminBenetti:
## Problem

`fleet spawn-session <instance> main` creates a bare-named tmux session that doesn't follow the TUI's `<instance>~<group>` naming convention. The chain that turns this into duplicated sessions:

1. `groupSessions` (internal/tui/groups.go) wraps the non-conforming session into a pseudo-group whose ID is the session name itself (`main`).
2. On attach, the split-key binding passes that pseudo-group ID to `bindHostSplitKeys` — the attach path right above guards with `isGroupedSession`, but the bind never did.
3. Pressing the split key runs `fleet shell <instance> --group main`, which mints a brand-new session `<instance>~main~<hex>` — a *real* group named `main`, duplicating the bare session in the UI.

Root cause: the session-name construction logic was duplicated between the CLI and the TUI, and the copies disagreed.

## Fix

Make the TUI's naming convention the single source of truth (`internal/tui/groups.go`):

- Export `GroupSessionName`, and add `NewGroupRootSessionName`, `NewGroupPaneSessionName`, and `ResolveSessionName`.
- `spawn-session`, `exec-in-session`, `read-session`, and `fleet shell --session` canonicalize user-supplied names via `ResolveSessionName` (`main` → `<instance>~main`; already-conforming names pass through). The short name an agent spawns with keeps working across all the `*-session` commands.
- `fleet shell` drops its hand-rolled copy of the `~`-joined name construction in favor of the shared helpers.
- Defense in depth for pre-existing bare sessions: the split-key bind in `app.go` now mirrors the `isGroupedSession` guard the attach path already had, so a pseudo-group ID can no longer reach `fleet shell --group`.

## Testing

- New unit tests for `ResolveSessionName` (canonicalization, sanitization, pass-through, foreign-prefix handling).
- New integration test: `spawn-session` with an already-canonical name doesn't double-prefix.
- Updated `TestSpawnSession_CreatesSession` to expect the canonical name; the existing spawn → exec → read round-trip tests pass unchanged through the new resolution.
- `go test ./...` passes.

### DIFF
--- a/internal/admiralskill/SKILL.md
+++ b/internal/admiralskill/SKILL.md
@@ -88,6 +88,10 @@ fleet exec-in-session <instance> <session> "<command>"   # type a command + Ente
 fleet read-session <instance> <session> [--scrollback N] # capture the session's screen
 ```
 
+Session names are canonicalized to `<instance>~<name>` internally so the
+session appears as a regular session group in the fleet TUI — keep using the
+short name with these commands; they all resolve it the same way.
+
 `exec-in-session` sends keystrokes (it does not wait for completion), so to
 follow progress you read the session back. `read-session --scrollback N` returns
 the last N lines (`0` = just the visible screen, `-1` = full history). To put an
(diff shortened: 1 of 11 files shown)


BenjaminBenetti:
/qa-manager when `fleet` testing make sure to configure it with repos that have only simple devcontainer setups. If you pick one with docker in it, it will not work since the QA env is in docker in docker in docker so things get funky. Prefer simple setups over complex.

PR #160, opened 2026-06-12
fix/multi-tui-group-layout-desync — stop two TUIs fighting over session-group state (#158)

### LINKED ISSUE #158: Fleet Armada - Multi Pane State Desync.
When 2 fleet TUIs are connected to the same `fleetd` instance, and one TUI adds panes to a session. The state file can end up in a sort of "fight state". Where the other TUI overwrites the state with a now out of date version. 

An example is this state snipppet from the `fleetd` state file. 
```
    "gotta-have-dogwater/dog": {
      "groupID": "dog",
      "instanceName": "gotta-have-dogwater",
      "sessions": [
        "gotta-have-dogwater~dog",
        "gotta-have-dogwater~dog~3421"
      ],
      "layout": "2b01,227x53,0,0{68x53,0,0[68x26,0,0,0,68x26,0,27,7],158x53,69,0,8}",
      "paneCount": 2
    },
```

The state can be observed to flip back and forth. between the two pane counts, when interaction
with the session is invoked by one TUI or the other. Fully quitting one of the TUIs and loading it back up will re-sync the two until more split panels are added .
```
    "gotta-have-dogwater/dog": {
      "groupID": "dog",
      "instanceName": "gotta-have-dogwater",
      "sessions": [
        "gotta-have-dogwater~dog",
        "gotta-have-dogwater~dog~3421",
        "gotta-have-dogwater~dog~af01"
      ],
      "layout": "324e,227x53,0,0{68x53,0,0[68x26,0,0,0,68x13,0,27,12,68x12,0,41,13],158x53,69,0,14}",
      "paneCount": 3
    },
```

BenjaminBenetti:
## Summary

Fixes #158 — with two TUIs connected to the same `fleetd`, the persisted group layout (`paneCount`/`layout`/`sessions`) flip-flopped between each TUI's view whenever either interacted with the session.

Root cause (three compounding problems, detailed in [the issue comment](https://github.com/BenjaminBenetti/fleet-man/issues/158#issuecomment-4687768069)):

1. The per-TUI `savedGroups` snapshot cache was hydrated once at startup and never reconciled with Watch state changes, so one TUI's writes left the other permanently stale until restart.
2. Group restore trusted the stale snapshot verbatim and deliberately ignored the live session list, recreating the old pane set.
3. `saveCurrentGroupLayout` persisted whatever the local outer tmux showed, unconditionally (last-writer-wins), letting a stale view clobber newer state — hence the ping-pong.

Fix — the inner tmux session list (polled by fleetd, pushed over Watch) is the shared source of truth:

- **Stale-view guard:** `saveCurrentGroupLayout` only persists a snapshot whose session set matches the live runtime sessions for the group. A mismatch is skipped *without* updating the diff-gate cache, so legitimate local changes (whose runtime echo lags ~1s) retry on a later tick and can never clobber the other TUI's newer state.
- **Watch reconcile:** `savedGroups` now mirrors server `GroupLayouts` on every `stateChangedMsg` (replace semantics, deletions propagate), not just at startup — layouts written by another TUI show up without restarting. The group open in this TUI's split is exempt: its outer tmux is the live truth for this view.
- **Restore merge:** `restoreSessionNames` appends live group sessions missing from the saved snapshot, so a pane added by another TUI is restored on reopen, while snapshot sessions are still kept when discovery doesn't list them (recreate-if-dead semantics for fleet restarts). Also hoisted the `savedGroups` read out of the restore closure, which was reading the map off the Update goroutine.

### DIFF
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -210,7 +210,7 @@ func newModel() model {
 	// Rehydrate saved pane layouts from disk so group restores after
 	// a fleet restart use the exact geometry the user left behind,
 	// then drop any layouts whose instance no longer exists.
-	m.hydrateSavedGroups()
+	m.reconcileSavedGroups()
 	m.pruneOrphanedSavedGroups()
 
 	// Resume tracking any instances still in "creating" state from a previous session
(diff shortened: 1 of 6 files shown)


BenjaminBenetti:
/qa-manager Your pass on c9fd20f validated the write-side guard, but user testing found a read-side hole that your plan didn't cover — please add this scenario to the regression plan when re-testing the new commit (90378c5):

1. TUI-A: open a session group and split it to 3-4 panes (so extra group sessions exist, e.g. `inst~grp~xxxx`).
2. TUI-B: open the same group (so its local snapshot records the full pane set), then close the split in TUI-B.
3. TUI-A: kill 2 of the panes (inner sessions die). Persisted state correctly drops to the smaller session set.
4. TUI-B: re-open the group.

On c9fd20f, step 4 could RESURRECT the killed sessions: restore read TUI-B's stale local snapshot, and each restored name goes through `fleet shell --session` → `tmux new-session -A` (attach-or-create), recreating the dead sessions as real empty panes — after which the state legitimately jumps back to the larger paneCount (observed by the user as 2 → 4).

Fix in 90378c5: restore now reads the server's persisted layout (kept fresh over Watch) instead of the per-TUI cache, and dead snapshot sessions are only recreated when the group has NO live session left (the fleet-restart case). With any group session alive, dead entries are dropped and live extras are appended.

Expected after the fix: step 4 reopens exactly the surviving panes; persisted paneCount/sessions stay at the post-kill set. Please also keep the fleet-restart restore case in the plan (all group sessions dead → all panes recreated from the saved layout).

PR #163, opened 2026-06-13
feat/version-chain-collapse — collapse a matched version chain to '<version> xN'

BenjaminBenetti:
## Summary

Small follow-up to the control-chain version header (#162). When every hop is on the same build — the healthy, in-sync case — the arrow form is just noise:

- before: `v1.0.19-beta → v1.0.19-beta → v1.0.19-beta`
- after: `v1.0.19-beta x3`

A matched two-hop chain (remote `FLEET_SERVER`) collapses to `… x2` the same way. Any divergence keeps the full arrow form so skew still stands out — including an old gateway's `?` hop (`v2.0.0 → ? → v2.0.0` is not collapsed). The local single-version header is unchanged.

Verified live (all three hops on v2.0.0 → header shows `v2.0.0 x3`) and with unit tests covering x3, x2, and the non-collapse cases (mixed versions, and matched ends around a differing gateway).

### DIFF
--- a/internal/tui/render.go
+++ b/internal/tui/render.go
@@ -34,9 +34,9 @@ func versionChain(m *model) string {
 		if gw == "" {
 			gw = "?"
 		}
-		return versionLabel(version.Version) + versionArrow + gw + versionArrow + versionLabel(m.serverVersion)
+		return joinVersions(versionLabel(version.Version), gw, versionLabel(m.serverVersion))
 	case fleetclient.IsRemote():
-		return versionLabel(version.Version) + versionArrow + versionLabel(m.serverVersion)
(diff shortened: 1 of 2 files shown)


BenjaminBenetti:
/qa-manager infra died out for a moment there, please retest, I've fixed it.

BenjaminBenetti:
Addressed in 4854a4b — added the `len(hops) == 0` guard so `joinVersions` can't panic on an empty call, plus a `TestJoinVersionsEdges` test locking in the 0/1/2-arg behavior (empty, bare single, `x2`). Good catch; it was unreachable today but a real latent footgun.

Issue #150, opened 2026-06-12
Fleet TUI - Layout Presets


BenjaminBenetti:
## Problem statement
When working  on certain projects (fleets) I find that I have a pre-set setup for different activities. For example I usually have an "app-run" session which has the app running in the right pane, and a lazydocker window on the right under the fleet TUI. There are many such possible configurations I could imagine a user wanting. 

## What this enables 
By being able to quick restore a layout preset a user can quickly get up and running in a fresh instance, very quickly without having to re-split tmux and re-launch the programs.

## Layout Setup 

the user can setup layouts via a new section in the fleet settings dialog. "Layouts" this section will be collapsible like "Caching" and it will appear above "Custom mounts" and under "Home Dir"

Inside this section there will be a "+ Layout Preset" button. when activated by the user an inline creation process will kick off.

First the user will select from the currently active sessions across all instances in the fleet. The display will show <instance name> / <session name> so that it's clear even if multiple session have the same name.

After selecting a session to copy (thus specifying the layout)  a small layout preview will render. showing the panel layout, this will be navigable by the user using all normal navigation methods used in the TUI. Text at the bottom will say 

"[ cancel ]        Assign startup commands to all panels to continue"

The user after selecting a pane in the preview sees a box where they may type a bash command.  This is the command that will be invoked in that pane automatically when making the layout. After all panes are assigned the text at the bottom is replaced with these options:

"[ cancel ]       [ create ]"

The user can save and or cancel creation. Note that ESC/q always cancels at any time to be consistent with the TUI conventions. 

Finally this pane layout preview with command editing will be available on existing layout templates as well. The button at the bottom will simply read 

" [ cancel ]       [ save ] "

as the only major difference. 

## Usage

when creating a new session via the existing new session dialog, the user can use "Tab" to cycle through templates that are configured on the instance. This is similar to how backends can be cycled when creating a new instance. The user still names the session. When created the session will be pre-loaded with the saved layout, and the saved commands auto invoke in each pane as setup.

PR #165, opened 2026-06-13
feat(tui): layout presets — capture, edit, and apply session pane layouts (#150)

BenjaminBenetti:
## Summary

Implements **layout presets** (issue #150): per-fleet templates that pair a captured tmux pane layout with a startup command per pane, so a user can spin up a fresh session pre-split and pre-loaded with the right programs.

**Setup** — the edit-fleet dialog gains a collapsible **Layouts** section (above Custom mounts, under Home dir). `+ Layout Preset` starts an inline flow: pick a live session group across any of the fleet's instances (shown as `<instance> / <session>`), preview its pane layout in a navigable box-drawing diagram, optionally set a bash startup command per pane (an empty one is a plain shell), name it, and `[ save ]`. Navigation in the preview is spatial — arrows/`hjkl` move between panes by their on-screen geometry, plus the name row and the cancel/save buttons. Existing presets reopen the same preview to edit, and each row's `[ remove ]` button works like the Caching section's `[ Delete cache ]` (→ to focus it, Enter to arm a `[remove?]` confirm, Enter again to remove).

**Usage** — the new-session dialog `Tab`-cycles the fleet's presets as a bracketed selector (`[ none ]` → `[ name ]`, like the backend selector). Creating with a template selected mints the whole session group — one session per pane — auto-runs each saved command, and records the group layout so opening the session restores the captured geometry. A blank name defaults to the template's name.

**Data model** — `FleetSettings` grows `repeated LayoutPreset {name, layout, pane_commands}`, validated server-side in `SetFleetSettings` (`NormalizeLayoutPresets`: non-empty unique names, 1–32 panes) and round-tripped through all four proto⇄legacy converters + `state.json`. The captured `layout` is the same `window_layout` string the existing group save/restore already persists; `pane_commands` is position-ordered (top then left) like `GroupLayout.sessions`.

## Notes

- Persistence rounds entirely through fleetd (`SetFleetSettings` RPC → `state.Update` → `state.json`); an integration test (`291`) creates a preset, kills the daemon, and confirms it reloads on restart.
- Pane startup commands are sent with `tmux send-keys -l --` so a command beginning with `-` is typed literally rather than parsed as a flag.

### DIFF
--- a/fleetgrpc/domain.pb.go
+++ b/fleetgrpc/domain.pb.go
@@ -395,7 +395,12 @@ type FleetSettings struct {
 	// Code CLI — keeps its login session and settings there) plus the Auggie
 	// auto-install startup script. Plain bool like the other *Mount flags:
 	// false == "never set" == disabled.
-	AuggieMount   bool `protobuf:"varint,10,opt,name=auggie_mount,json=auggieMount,proto3" json:"auggie_mount,omitempty"`
+	AuggieMount bool `protobuf:"varint,10,opt,name=auggie_mount,json=auggieMount,proto3" json:"auggie_mount,omitempty"`
+	// layout_presets is the fleet's list of saved session-layout templates
+	// (issue #150). repeated message: an empty list (the default) means no
+	// presets. Names are unique within the list (the server normalizes and
+	// rejects duplicates, like custom_mounts).
+	LayoutPresets []*LayoutPreset `protobuf:"bytes,11,rep,name=layout_presets,json=layoutPresets,proto3" json:"layout_presets,omitempty"`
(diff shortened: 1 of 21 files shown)


BenjaminBenetti:
/qa-manager Please re-run the QA pass against the current tip commit `5e41360`. Your last run validated `068ed77`; the only change since is a one-line defensive fix in buildPresetSessionScript (adding a `--` flag terminator before the literal `tmux send-keys` command) so that a pane startup command beginning with `-` is sent literally instead of being parsed as a send-keys flag. Worth a quick re-verify of the templated-session creation + startup-command execution path.

PR #174, opened 2026-06-15
feat/fleet-copy-bidirectional — generic scp-style copy in both directions and between instances (#164)

### LINKED ISSUE #164: Fleet Copy  should go both ways!
fleet copy should follow `scp` style setup. It should be possible to copy in both directions. Currently only items can be copied out of an instance but not into an instance.

BenjaminBenetti:
## Summary

Closes #164. `fleet copy` (alias `fc`) was one-directional — it could only pull a file *out* of an instance. It's now a generic, scp-style copy that works in **both directions and between instances**, built as a single engine rather than per-direction patches.

Each of the two arguments is an endpoint, parsed uniformly:

- `[fleet/]instance:path` — a file inside an instance (reachable even on a remote fleet)
- a plain path — a file on the orchestrating machine
- `:path` — a file in the **current** instance (in-instance form)

Direction is inferred from which side is local vs an instance. One engine (`fleetclient.Copy`) dispatches all four combinations:

| src → dst | mechanism |
|---|---|
| local → instance | **new `CopyInto` client-streaming RPC** (`tar -xf -` into the container) |
| instance → local | `CopyFile` (unchanged) |
| instance → instance | relay one `CopyFile` stream into one `CopyInto` stream |
| local → local | a plain file copy |

```
fleet copy alpha:bin/tool ./tool          # download
fleet copy ./tool alpha:/usr/local/bin/   # upload
fleet copy alpha:a beta:b                 # between instances
fc :build/out ~/Downloads/                # this instance → your machine
fc report.csv :/tmp/                      # your machine → this instance
fc :out other/inst:/tmp/                  # this instance → another
```

## Notable decisions / behavior changes

- **Atomic upload.** The server extracts into a temp name and `mv`s it into place, so a truncated/oversized stream fails the RPC without ever clobbering an existing destination — the same atomicity the download direction already had. The declared size is a hard cap; the destination directory is probed up front for clean errors; the basename is validated.
- **In-instance `fc` now always delegates to the host TUI.** The container can't reach the host fleetd over gRPC (only the control socket is bind-mounted), so `fc` sends `{src,dst}` over the control socket and the connected TUI runs the same engine against the **host** fleet with its own disk as "local". This is a deliberate change from the previous behavior where the in-instance colon form resolved against a nested fleet — instance names now resolve against the user's own fleet, which is what makes `fc foo other-instance:/tmp/` mean what you'd expect. As with the old out-direction, it needs a host TUI attached.
- **Backend caveat (pre-existing, shared with the out-direction):** the codespaces backend allocates a PTY for exec, whose line discipline can mangle binary data on stdin; the strict size check turns any resulting truncation into a clean error rather than a silent partial write. devcontainer (default) and coder exec over a clean binary channel.

## Breaking (internal, single-binary)

- proto `FileCopy` fields `path`/`dest` → `src`/`dst`; control `CopyFilePayload` `Path`/`Dest` → `Src`/`Dst`. All consumers updated in lockstep.

## Tests

- New: endpoint parser, engine combos (over an in-process gRPC server), `CopyInto` handler (real `tar`/`test` exec) incl. the atomicity guarantee, TUI/host local-path policies, control round-trip, watch event.
- Integration `025_copy.sh` extended with upload, directory-upload, missing-dir/missing-local failure, and instance→instance.

### DIFF
--- a/fleetgrpc/exec.pb.go
+++ b/fleetgrpc/exec.pb.go
@@ -1014,6 +1014,245 @@ func (x *CopyFileMeta) GetSize() int64 {
 	return 0
 }
 
+// CopyInto streams one file INTO an instance — the reverse of CopyFile, the
+// other half of scp-style `fleet copy`. It is CLIENT-streaming: the FIRST chunk
+// MUST carry `open` (which instance, the destination, and the file's
+// name/mode/size); every following chunk carries `data`. The server resolves the
+// destination inside the container (an existing directory keeps the source
+// basename; otherwise `dest` is the target path) and writes the bytes through
+// the backend with a single `tar -xf -` exec — the mirror of CopyFile's
+// `tar -chf -` read — preserving the file mode. Works unchanged over the remote
+// gateway: pushing a local file into a fully remote deployment.
+type CopyIntoChunk struct {
+	state protoimpl.MessageState `protogen:"open.v1"`
+	// Types that are valid to be assigned to Msg:
+	//
+	//	*CopyIntoChunk_Open
+	//	*CopyIntoChunk_Data
+	Msg           isCopyIntoChunk_Msg `protobuf_oneof:"msg"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
+}
+
+func (x *CopyIntoChunk) Reset() {
+	*x = CopyIntoChunk{}
+	mi := &file_exec_proto_msgTypes[14]
+	ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
+	ms.StoreMessageInfo(mi)
+}
+
+func (x *CopyIntoChunk) String() string {
+	return protoimpl.X.MessageStringOf(x)
+}
+
+func (*CopyIntoChunk) ProtoMessage() {}
+
+func (x *CopyIntoChunk) ProtoReflect() protoreflect.Message {
+	mi := &file_exec_proto_msgTypes[14]
+	if x != nil {
+		ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
+		if ms.LoadMessageInfo() == nil {
+			ms.StoreMessageInfo(mi)
+		}
+		return ms
+	}
+	return mi.MessageOf(x)
+}
+
+// Deprecated: Use CopyIntoChunk.ProtoReflect.Descriptor instead.
+func (*CopyIntoChunk) Descriptor() ([]byte, []int) {
+	return file_exec_proto_rawDescGZIP(), []int{14}
+}
+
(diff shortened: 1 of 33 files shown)


BenjaminBenetti:
/qa-manager Heads up on a behavior change since the commit you last tested (ffa4481) — all 12 cases passed then, thank you.

As of dcc072e, a delegated in-instance `fc` copy that **touches the host machine** (reading a host file into an instance, or writing one out — including the 1-arg download-to-Downloads shorthand) now requires a **confirmation in the fleet TUI** before it runs. The modal offers: **[a]llow once**, **[s] allow for session** (remembers that instance in memory while the TUI stays open), and **[d]eny**.

What this means for the QA cases that go to/from the host (e.g. "in-instance fc instance→host" and "host→instance"): after running `fc ...` inside the instance, the copy will block until you approve it at the TUI — press **a** (or **s**) in the fleet TUI to allow it, then verify the file as before. A denied copy is expected to NOT transfer.

Pure **instance→instance** copies (e.g. `fc :file other:/tmp/`) do **not** prompt and behave exactly as you tested. The host-form `fleet copy` (run outside an instance) also does not prompt.

This gate exists because anything in a container can send a copy request, so the human at the TUI confirms host file access. Let me know if the harness can't drive the TUI keypress and I'll find another way to make these cases testable.

BenjaminBenetti:
/qa-manager Thanks for the thorough run on `77e1745` — cases 1–5 (download, upload, instance→instance, local→local, error handling) all passed. 

On **Case 6 (in-instance fc), marked blocked**: I don't think this should fail the PR, for two reasons:

1. **It's a harness/environment limitation, not a code issue.** As you noted, the headless env can't spin up and attach the host TUI's control socket (`/fleet-mounts/control/fleet.sock`). `fc` delegating to a connected host TUI is a **pre-existing design requirement** of the feature (the old OUT-only `fc` needed it too) — there's no code path that makes the container reach the host fleetd without a TUI, by design (it's the control-socket → Watch → client split that keeps it working for remote fleets).

2. **These exact flows were already verified passing on the prior run.** Your `dcc072e` run (06:22Z) confirmed in-instance instance→host (allow), host→instance (allow), instance→instance (prompt suppressed), and the **deny** case all worked correctly through the confirmation prompt. The only change from `dcc072e` to `77e1745` is an internal temp-filename generator (an atomic counter so concurrent uploads can't collide) — **zero user-visible behavior change**, so the `dcc072e` in-instance verification still holds for HEAD.

Could you mark Case 6 as **blocked/not-applicable for the headless env** (carrying over the `dcc072e` pass for the in-instance flows) and re-mark the PR `qa-passed`? Happy to add anything that would make the in-instance path testable headlessly if you have a suggestion, but I believe it needs an attached TUI by design.

PR #176, opened 2026-06-16
fix/fleet-copy-local-cwd-and-host-prefix — cwd-local paths, host: prefix, and recursive directory copy

BenjaminBenetti:
## Summary

Follow-ups to the bidirectional `fleet copy` / `fc` work (#174). Three things, all in the copy path:

### 1. Plain paths are cwd-local; new `host:` prefix
A plain path now means a file on the machine the command runs on, relative to its cwd — the host for `fleet copy`, and **this instance** for in-instance `fc` (previously an in-instance plain path resolved against the host's disk, which was surprising). `host:path` reaches the host machine from inside an instance.

```
fc ./build/out host:~/Downloads/   # this instance -> your machine
fc host:report.csv /tmp/           # your machine  -> this instance
fc ./cp-me.txt other:/tmp/         # this instance -> another instance
```

### 2. Recursive directory copy
`fleet copy` / `fc` now copies **directories** recursively in all four directions (upload, download, instance-to-instance relay, local-to-local). A new proto `is_dir` flag selects a tar path; the single-file path is unchanged.

- The server stays a plain `tar` in/out; **all** the tar safety is done in Go on the client (`copytree.go`): symlinks/special files and the `./` root entry are skipped, and every entry is written through an **`os.Root`** rooted at the target — so no `..`/absolute name and no on-disk symlink at any path component can escape it.
- **Refused on the codespaces backend** (its exec PTY can silently corrupt binary tar), and **non-atomic** (cp -r-style in-place merge; fails loudly, never silently, on devcontainer/coder).

### 3. Tidy `fc` messages/help
Trimmed the in-instance confirmation to just `Copying SRC -> DST` (no "approve it in your TUI…" monologue) and shortened two over-explained errors.

## Tests

Endpoint parsing, host/TUI resolution, the message tidy, the engine across all four transports over an in-process server, server-side dir extract with real `tar`/`mkdir`, both codespaces gates, and security tests (path traversal, symlink-entry skip, intermediate-symlink escape, root-entry skip). Integration `025_copy.sh` covers upload/download/relay for both files and directories.

> Manually tested; not run through the automated review/QA bots by request.

### DIFF
--- a/fleetgrpc/exec.pb.go
+++ b/fleetgrpc/exec.pb.go
@@ -801,11 +801,11 @@ func (x *ForwardOpen) GetRemotePort() int32 {
 	return 0
 }
 
-// CopyFile streams one file out of an instance to the client (the `fleet copy` /
-// in-instance `fc` feature). The server reads the file through the backend (a
-// tar pipe over container exec, so metadata + bytes arrive in one exec) and
-// streams it back: the FIRST chunk carries `meta`, every following chunk carries
-// `data`. Only regular files are supported — a directory is an InvalidArgument.
+// CopyFile streams a file OR directory out of an instance to the client (the
+// `fleet copy` / in-instance `fc` feature). The server reads through the backend
+// (a tar pipe over container exec) and streams it back: the FIRST chunk carries
+// `meta`, every following chunk carries `data`. For a file the data is the raw
+// bytes; for a directory (meta.is_dir) it is a tar the client extracts in Go.
 // Works unchanged over the remote gateway, which is the headline use case:
 // pulling a built binary out of a fully remote deployment onto the local machine.
 type CopyFileRequest struct {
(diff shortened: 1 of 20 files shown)


BenjaminBenetti:
/qa-manager This PR **changes the in-instance path semantics** from the fleet-copy plan you ran on #174, so a heads-up to update the expectations:

**What changed:** inside an instance, a **plain path is now THIS instance's file (cwd-relative)**, not the host's. To reach the host machine you now use a **`host:` prefix**. So the syntax from the old plan flips:

| Intent | Old (#174) | New (this PR) |
|---|---|---|
| this instance's file → your machine | `fc :build/out host... `/`fc :f ~/d/` | `fc ./build/out host:~/Downloads/` |
| your machine's file → this instance | `fc report.csv :/tmp/` | `fc host:report.csv /tmp/` |
| this instance → another instance | `fc :out other:/tmp/` | `fc ./out other:/tmp/` |

**Key things to verify (need an attached TUI, like your dcc072e run on #174):**
1. In-instance `fc ./somefile host:/tmp/x` copies the **instance's** ./somefile to the host (prompts for host access — approve with `a`).
2. In-instance `fc ./somefile other-instance:/tmp/` copies between instances and does **NOT** prompt (no host touch).
3. `fc host:/host/file ./dest` reads a **host** file into the instance (prompts).
4. A bare `fc ./instancefile` (1-arg) downloads to your downloads folder (prompts — it writes your machine).

**Host form is unchanged** (`fleet copy ./local instance:/path` etc. all still pass), plus `host:` now works there too as a synonym for a plain path (covered by the headless itest). If the headless env can't attach the TUI for the in-instance cases again, the host-form cases (incl. `host:`) still exercise the parser end-to-end. Thanks!

PR #187, opened 2026-06-20
feat/auto-tag-pr-status — show live GitHub PR status as an instance auto tag

BenjaminBenetti:
## Why

Instances support a user-set **tag** (`t` then type), but most of the time you
just want to know *what state the work is in* — is the PR up, did review pass,
are checks green? This adds an **auto tag**: when an instance has no user-set
tag, the tag slot instead shows a live, glanceable summary of its GitHub PR
state. A user-set tag always takes precedence, so nothing changes for tagged
instances.

## What

For an expanded instance with no user tag, the tag line renders:

```
[PR | PRxN]   [Changes Requested | Approved]   [Checks x/x]
```

Each element is dynamic and colour-coded; absent elements are omitted.

- **PR** — shown when a PR is open on the current branch of the workspace repo
  **or any nested subrepo**. `PRxN` when N>1 (counts summed across repos).
  Green when every PR is mergeable per GitHub (`mergeStateStatus == CLEAN`),
  red on any check failure or changes-requested, yellow otherwise (open, in
  progress).
- **Changes Requested / Approved** — red/green from the aggregate review
  decision; hidden until a decision lands.
- **Checks x/x** — passed/total summed across all open PRs. Green when all
  passed, red on any failure, yellow while any are pending.

### How it works

A new server-side poller (`internal/server/pr_status.go`) runs `gh pr list
--json …` **inside the instance** (so it uses the instance's own gh auth) on a
slow cadence (~45s, plus immediately when a TUI first subscribes), aggregates
the result, and pushes a new `PrStatus` message over the existing runtime
sidecar (`InstanceRuntime.pr_status`). The thin TUI just formats it
(`internal/tui/auto_tag.go`), mirroring how agent activity / stats already flow.

### Graceful degradation

If `gh` is missing or not logged in inside the instance, the probe emits a
sentinel and the auto tag is simply not shown — today's behaviour is preserved.
Transient probe failures keep the last known status rather than clearing it.

## Notes / scope

- The auto tag occupies the **same slot** as the manual tag (shown under an
  expanded instance), per "alternate display mode for tags". Always-visible
  (collapsed) display would be a larger UX change and is left as a follow-up.
- PRs are matched by each repo's **current branch** (`--head`), so the tag
  reflects *this instance's* work rather than every open PR in the repo.

### DIFF
--- a/fleetgrpc/runtime.pb.go
+++ b/fleetgrpc/runtime.pb.go
@@ -201,6 +201,112 @@ func (AgentActivity) EnumDescriptor() ([]byte, []int) {
 	return file_runtime_proto_rawDescGZIP(), []int{2}
 }
 
+// PrSignal is the three-state colour shared by the auto-tag indicators:
+// green = good, yellow = in-progress/neutral, red = needs attention.
+type PrSignal int32
+
+const (
+	PrSignal_PR_SIGNAL_UNSPECIFIED PrSignal = 0
+	PrSignal_PR_SIGNAL_GREEN       PrSignal = 1
+	PrSignal_PR_SIGNAL_YELLOW      PrSignal = 2
+	PrSignal_PR_SIGNAL_RED         PrSignal = 3
+)
+
+// Enum value maps for PrSignal.
+var (
+	PrSignal_name = map[int32]string{
+		0: "PR_SIGNAL_UNSPECIFIED",
+		1: "PR_SIGNAL_GREEN",
+		2: "PR_SIGNAL_YELLOW",
+		3: "PR_SIGNAL_RED",
+	}
+	PrSignal_value = map[string]int32{
+		"PR_SIGNAL_UNSPECIFIED": 0,
+		"PR_SIGNAL_GREEN":       1,
+		"PR_SIGNAL_YELLOW":      2,
+		"PR_SIGNAL_RED":         3,
+	}
+)
+
+func (x PrSignal) Enum() *PrSignal {
+	p := new(PrSignal)
+	*p = x
+	return p
+}
+
+func (x PrSignal) String() string {
+	return protoimpl.X.EnumStringOf(x.Descriptor(), protoreflect.EnumNumber(x))
+}
+
+func (PrSignal) Descriptor() protoreflect.EnumDescriptor {
+	return file_runtime_proto_enumTypes[3].Descriptor()
+}
+
+func (PrSignal) Type() protoreflect.EnumType {
+	return &file_runtime_proto_enumTypes[3]
+}
+
+func (x PrSignal) Number() protoreflect.EnumNumber {
+	return protoreflect.EnumNumber(x)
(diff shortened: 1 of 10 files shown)


BenjaminBenetti:
Thanks for the review — addressed in `8431d1a` (sentinel) and `71a8f6e` (timeout):

- **HIGH — `Wait` blocking on orphaned children:** fixed in `71a8f6e`. The probe now runs in its own process group and a timeout signals the whole group (`syscall.Kill(-pid, SIGKILL)`), with `WaitDelay` as a backstop, so a hung `gh`/child can't hold the stdout pipe open and wedge `Wait`. Added direct tests for the capture and timeout paths.
- **MEDIUM — sentinel matched with `Contains`:** fixed in `8431d1a`. Now a prefix match on the trimmed output. The sentinels are only ever emitted as the sole leading output (the script exits right after), while real output is a JSON array starting with `[`, so a sentinel string appearing *inside* PR JSON can no longer disable the tag. Added a regression test.
- **LOW — gate-tick instead of `h.runtimeEdge` for the subscribe edge:** keeping the gate-tick. It detects the false→true edge within `prStatusGatePoll` (≤5s), and — as you noted — wiring `h.runtimeEdge` in would require reworking that single-consumer channel into a broadcast. A few seconds' latency on a status line that refreshes every 45s isn't worth that. Clarified the rationale in the code comment.

BenjaminBenetti:
/qa-manager Thanks for the thorough run — Steps 1–3 passing covers the core behavior. On Step 4 (multiple subrepos) being blocked: that path is a test-harness limitation rather than a defect, and the mock setup is simpler than it looks because **the mock `gh` does not need to differ per repo**.

The probe discovers repos by `find`ing nested `.git` dirs under the workspace and runs `gh pr list` in each, then the server sums PRs/checks across them. So reusing the *same* Step-2 mock `gh` (returns one open PR for any `gh pr list`) plus one nested git repo is enough to get `PRx2`:

```sh
# in the instance workspace (which is repo #1), with the Step-2 mock gh on PATH:
mkdir -p services/api
git -C services/api init -q
git -C services/api commit --allow-empty -m init -q
git -C services/api checkout -q -b feature
```

The probe then finds both the workspace repo and `services/api`, runs the mock in each, and the auto-tag aggregates to `PRx2` with checks summed. I validated this end-to-end locally: the script emits two PR arrays and the aggregation yields `PRx2`.

The cross-repo aggregation is also covered by unit tests independent of the container: `TestAggregatePRStatus` (the "three PRs summed → PRx3, checks summed, red on any failure" case) and `TestParsePRProbeOutput_ConcatenatedArrays` (parsing concatenated per-repo `gh` arrays).

Could you re-run Step 4 with the recipe above (I'll re-apply `ready-for-qa`)? If it's still impractical in the harness, I'd argue the unit coverage above is sufficient to consider Step 4 verified.

PR #191, opened 2026-06-20
feat/auto-tag-pr-navigate — select and open the PR status auto tag

BenjaminBenetti:
## Why

Make the v1.2.0 PR-status auto tag interactive, on top of a display change that moves it to a "second status line".

## What

**Display (merged from `bb-tmp-agent-transfer-06-2026`):** the PR status no longer has its own row — it rides the **first child row** of an expanded instance (first session, or "+ new session") at the status column, aligned under the instance's status word (`row.prStatusInline` + `renderChildRowLine`). `rowInstanceTag` is now user-tags only.

**Navigation:** with the cursor on that child row (where it already rests):
- **`→`/`l`** selects the PR status — drawn as one pink chunk wrapped in `[ ]`; **`←`/`h`** deselects.
- **`enter`/`space`** (while selected) opens the PR in the system browser; a **click** on the inline status opens it too.
- **Multiple PRs** → a chooser dialog (number, title, URL).
- "View logs" moved from `l` to **`L`**.

**Review states (new):** the review element now shows one of three colour-coded states for an open PR:
- **Changes Requested** (red) — any PR needs changes
- **Accepted** (green) — every open PR is approved
- **Under Review** (yellow) — otherwise (review pending)

### How it works

- `PrStatus` gains `repeated PrRef prs` (number/url/title) and a new `PrReviewState.UNDER_REVIEW`; the server probe collects `url,title` and aggregates the review state (Accepted only when *all* PRs are approved).
- URLs open via a small per-OS opener (`xdg-open`/`open`/`wslview`/`rundll32`).
- A new `viewChoosePR` dialog handles the multi-PR case; `tagSelected` selection is gated on the cursor's inline-PR row and cleared on vertical moves / when the PR closes.

## Notes

- Reconciled on top of the transfer branch's inline model (merge commit); navigation was rebuilt to hang off `prStatusInline` child rows.
- An adversarial multi-agent review pass surfaced 4 fixes (space/tab sub-mode, narrow-terminal click guard, a stale "press l" toast, README) — all addressed.


BenjaminBenetti:
Addressed the openurl.go review in `6758477`:
- **HIGH (PowerShell injection):** the WSL fallback no longer interpolates the URL into the `-Command` string — it passes it via `$env:FLEET_OPEN_URL`, which PowerShell reads as a plain value (never re-parsed as code). Added an http(s)-only scheme check as defense-in-depth.
- **LOW (zombie launcher):** switched `Start()` → `Run()` so the short-lived opener is reaped.

Tests added for the scheme validation; all other openers already pass the URL as a non-shell-interpreted arg.

BenjaminBenetti:
On the `app.go` synthesizedKey finding: this one is a false positive. `synthesizedKey` is initialised to `tea.KeyMsg{Type: tea.KeyEnter}` at the top of the click handler (app.go:659), **not** the zero value. The inline-PR case intentionally leaves that default, so the click synthesizes **Enter** into the `tagSelected` path → `openSelectedPR` → the PR opens. This mirrors how the rowInstance click synthesizes Space and the armada click synthesizes `A`.

Added a regression test (`TestMouseClickOnInlinePROpensPR`, commit `191987a`) that drives a real left-click through `model.Update` and asserts the PR opens (message "Opening PR #99…"). It passes — so the click does open the PR, not merely select it. No code change needed; keeping the default-Enter pattern for consistency with the other click handlers.

Issue #188, opened 2026-06-20
Fleet - Automation TUI mode.


BenjaminBenetti:
### Overview / Workflows.

Introduce a "automation" mode to the fleet TUI. The user can switch a fleet between "instance" view (current view today) and "automation" mode via a toggle in the TUI. Automation mode has 2 core components

#### UX

Instance view (current state)
```
> fleet [automations]   <---- new button toggle
   > instance 1
   > instance 2
.....
```

Automation View
```
> fleet [instances]
   > triggers (x)
      trigger 1 
      trigger 2
   > agents (x)
      agent 1 <- x                 <---- indicates the number of triggers.
      agent 2 
````

#### Triggers 

Triggers let a user define triggers that cause an agent to spin up. A trigger is defined by the following core concepts / fields 

1. Trigger Type - The trigger type defines the broad type of the trigger. To start we will implement two types Schedule, and Webhook
2. Agents - The agent or agents that this trigger activates.
3. Trigger Prompt - the prompt to provide to the agent when invoked by this trigger. 

Depending on the type additional fields will be available (struct composition)

**Schedule Trigger **
1. Cron pattern - pattern on which to fire the trigger using a standard cron expression.

**Webhook**
1. name - this name will appear on the end of the fleet gateway webhook url (implementing this is out of scope)
2. filter type - two options, regex or json path 
3. regex - If filter type is 'regex' then the user may specify a regex expression that will filter incoming events, if the regex matches the event is fed to the agent.
4. json path - If filter type is 'json path' this will be a json path selector to select the value to compare 
5. json value - If filter type is 'json path' this will be the value which must be matched. The value at the specified json path will be compared against this value to figure out if the agent should be triggered. 

#### Agents 

An agent is the core worker how does the task in response to a trigger. An agent is defined by the following fields.

1. Agent Command - this will be a command string that will be invoked to start the agent up. This string will have a special placeholder value ${PROMPT} and ${SYS_PROMPT} that the system will use to inject its prompt. This field will REMEMBER the last value entered by a user, so that they don't need to type it out all the time. However, the default initial value will be `claude --system-prompt '${SYS_PROMPT}'  `. The user does NOT have to use the replacement variables if they don't want to.
[...]

PR #194, opened 2026-06-20
feat/automation-tui-mode — automation mode: triggers that spin up agents (#188)

BenjaminBenetti:
## Summary

Implements the **automation TUI mode** from #188. Each fleet can now be toggled (`m`, or the `[automations]`/`[instances]` button on the fleet header) between its normal **instance view** and a new **automation view** showing two collapsible groups — **triggers** and **agents** — that the user authors in place.

**Model + persistence.** New per-fleet `Agent` and `Trigger` types on `FleetSettings`, persisted inline in `state.json` and synced over gRPC exactly like layout presets (new proto `Agent`/`Trigger` messages, matching legacy↔proto converters on both the server and TUI sides, authoritative `NormalizeAgents`/`NormalizeTriggers` validation in `SetFleetSettings`).
- **Agent**: command (with `${PROMPT}`/`${SYS_PROMPT}` placeholders), tmux mode (default ON), system prompt, env backend.
- **Trigger**: type (Schedule | Webhook), the agent(s) it activates, a prompt, and type-dependent fields — a cron pattern for Schedule, or a webhook name + regex / json-path filter for Webhook.
- Cron matching is a small dependency-free 5-field implementation (`*`, ranges, lists, steps, the Vixie day-of-month/day-of-week rule).

**TUI.** A `buildRows` branch renders the two groups (reusing the existing row/cursor/collapse machinery); an agent shows how many triggers reference it (the issue's "← x"). Two modal editors handle add/edit, persisting through the existing optimistic `SetFleetSettings` path. Renaming an agent rewrites its trigger references; a referenced agent can't be deleted (so a trigger is never orphaned).

**Scheduler (fleetd).** An unconditional background loop fires due Schedule triggers — spawning a normal instance per agent via the same `CreateInstance` job path the RPC uses — then, once the instance is running, launches the agent command in a detached tmux session (so it's openable in the TUI) with the prompts substituted in. Tmux-mode agents that stay inactive past 2 minutes are torn down, detected with the same tmux screen-diff the TUI shows (so it works with no TUI attached). The live operations are package-var seams, so the lifecycle is unit-tested deterministically without containers.

## Scope notes
- **Webhook delivery is out of scope** per the ticket — webhook triggers are fully modeled, validated, persisted, and editable, but the fleet-gateway receiver endpoint is left to a follow-up.
- The agent command's "remember the last value" is implemented by defaulting a new agent's command to the most recently defined agent's command in that fleet (falling back to the built-in `claude --system-prompt '${SYS_PROMPT}'` default).

## Tests
- Unit: cron parse/match, agent/trigger normalization, a legacy→proto→legacy wire round-trip, TUI mode toggle + dialogs (add/edit/delete, validation, rename ref-update, render), and the scheduler (cron due-selection + per-minute dedup, launch-on-running, idle reap, non-tmux never reaped).
- Integration: `390_tui_automation_mode.sh` drives the real binary — toggle into automation mode, add an agent end-to-end, confirm it persists, toggle back.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -5,9 +5,28 @@ on:
     branches: [main]
 
 jobs:
-  integration:
+  # The suite is sharded by test-number range across parallel runners so the
+  # SLOWEST shard — not the whole 80+-test suite — sets the wall-clock. Each
+  # shard runs integration/run.sh with FLEET_ITEST_RANGE; run.sh filters to the
+  # tests whose numeric filename prefix falls in [START, END). The open-ended
+  # "600-" shard catches any future high-numbered tests so none are dropped.
+  integration-shard:
+    name: integration ${{ matrix.range }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    strategy:
+      # Run every shard to completion even if one fails — we want the full
+      # pass/fail picture across the suite, not a first-failure abort.
+      fail-fast: false
+      matrix:
+        range:
+          - "0-100"
+          - "100-200"
+          - "200-300"
+          - "300-400"
+          - "400-500"
+          - "500-600"
+          - "600-"
     steps:
       - uses: actions/checkout@v5
 
@@ -39,4 +58,23 @@ jobs:
       - name: Run integration tests
         env:
           FLEET_BIN: /usr/local/bin/fleet
+          FLEET_ITEST_RANGE: ${{ matrix.range }}
         run: ./integration/run.sh
+
+  # Single aggregate check: green only when every shard passed. Preserves one
+  # stable "integration" status name for branch protection / the review loop,
+  # independent of how many shards the matrix expands to.
+  integration:
+    name: integration
+    runs-on: ubuntu-latest
+    needs: [integration-shard]
+    if: ${{ always() }}
+    steps:
+      - name: Verify all shards passed
+        run: |
+          result="${{ needs.integration-shard.result }}"
+          echo "integration shard matrix result: ${result}"
+          if [ "${result}" != "success" ]; then
+            echo "::error::one or more integration shards failed (result=${result})"
+            exit 1
+          fi
--- a/fleetgrpc/domain.pb.go
+++ b/fleetgrpc/domain.pb.go
@@ -617,7 +617,6 @@ type Agent struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	Name          string                 `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
 	Command       string                 `protobuf:"bytes,2,opt,name=command,proto3" json:"command,omitempty"`
-	TmuxMode      bool                   `protobuf:"varint,3,opt,name=tmux_mode,json=tmuxMode,proto3" json:"tmux_mode,omitempty"`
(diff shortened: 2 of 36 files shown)


BenjaminBenetti:
Thanks for the review — addressed all three:

- **`schedule.go` (watch-set leak, high):** non-tmux fire-and-forget agents are now dropped from the watch set right after launch, with a defensive guard so a tracked non-tmux agent is never idle-reaped.
- **`automation.go` (prompt quoting, high):** `${PROMPT}`/`${SYS_PROMPT}` values are now single-quote-escaped (`'` → `'\''`) so a prompt containing a quote stays a literal argument inside the single-quoted placeholders the default command uses (documented the single-quote expectation).
- **`cron.go` (dow `7` ranges, medium):** fixed — the `7`→`0` fold now happens after range expansion, so `5-7` (Fri–Sun) parses correctly. (Independently caught in a self-review just before your comment landed.)

Added unit tests for the quote-escaping and the non-tmux drop-after-launch.

BenjaminBenetti:
Round 2:

- **`schedule.go` (`lastFired` leak, low):** fixed — `dueSchedules` now prunes `lastFired` entries whose schedule trigger no longer exists (deleted / renamed / type-changed), so the map can't grow without bound. Test added.
- **`automation.go` (`${PROMPT}` not re-expanded inside the system prompt):** this is intentional and already documented on `SubstituteAgentCommand`. Both placeholders expand in a single left-to-right pass precisely so a `${PROMPT}` that ends up *inside* the substituted system-prompt text is not re-expanded — that avoids double-expansion and a system-prompt→prompt injection path, and there's no recursion to worry about. Keeping the single-pass behavior by design.

BenjaminBenetti:
Round 3:

- **`automation.go` (prompt quoting, high):** strengthened — `${PROMPT}`/`${SYS_PROMPT}` are now substituted as **fully shell-quoted** values (wrapped in single quotes with `'` escaped), so they're safe in *every* context including a bare `echo ${PROMPT}`, not just inside single quotes. The default command's placeholder is now bare (`--system-prompt ${SYS_PROMPT}`) since the value self-quotes; documented that callers must not add their own quotes. Inject-guard test added.
- **`schedule.go` (sequential blocking probes, medium):** the per-agent activity probe now runs concurrently (bounded to 8) so a large watch set can't stall the tick. Map mutations and reap decisions stay on the single scheduler goroutine; each probe only touches its own detector/instance and the mutex-guarded `backendFor`, so it's race-free (verified with `-race`). Multi-agent concurrent-reap test added.
- **`automation_view.go` (synchronous gRPC in Update, medium):** this is the established pattern — the edit-fleet and layout-preset dialogs all call `setFleetSettingsRemote` synchronously in `Update`, to the **local** fleetd over a unix socket (fast). Making only the automation dialogs async would be inconsistent; a codebase-wide move to async settings writes is out of scope for this PR. Leaving as-is for consistency.

BenjaminBenetti:
Re the re-flagged `setFleetSettingsRemote` synchronous-save note (medium): keeping as-is by design, as discussed earlier in this review. Every settings mutation in the TUI — the edit-fleet dialog and the layout-preset flow included — calls `setFleetSettingsRemote` synchronously in the update loop; it's a gRPC call to the **local** fleetd over a unix socket (fast). Making only the automation dialogs async would make the codebase inconsistent; a blanket move to async settings writes is a separate, codebase-wide change out of scope for this PR.

Note for reviewers: QA is being run manually for this PR (automated QA budget exhausted), so the `ready-for-qa` cycle won't post a bot verdict here.

PR #195, opened 2026-06-21
feat(automation): enable/disable triggers

BenjaminBenetti:
## What

Adds a way to **disable / enable an automation trigger** without deleting it. Previously, once a trigger was created it always fired.

## How

- **Model**: new `Trigger.Disabled bool` (`disabled,omitempty`) mirrored by proto `bool disabled = 11`. Default `false`, so freshly-created triggers — and every trigger persisted before this field existed — keep firing. The flag is carried through all four wire converters (server + TUI, both directions).
- **Scheduler**: `dueSchedules` skips disabled triggers — no fire, and no `lastFired` bookkeeping entry, so re-enabling a trigger fires it on the next matching minute rather than suppressing it.
- **TUI**:
  - **Edit dialog**: an `Enabled: [ yes ]/[ no ]` selector row (toggle with `enter`/`space`/`h`/`l`), alongside Type/Filter. Edits preserve the flag.
  - **List**: press **`s`** on a highlighted trigger to flip enabled/disabled instantly (like start/stop on instances). Disabled triggers render **dimmed** with a `· disabled` marker, and the row hint advertises `s: enable/disable`.

Webhook delivery is still out of scope (as in #188), but the same `Disabled` gate will apply there once wired.

## Tests

- `TestFleetSettingsAutomationRoundTrip` now carries a disabled trigger (wire round-trip).
- `TestDueSchedulesSkipsDisabled` — a disabled schedule trigger never fires and records no `lastFired`.
- `TestToggleTriggerDisabled` — the list quick-toggle flips and persists.
- `TestEditTriggerPreservesDisabled` — editing keeps the flag; the dialog's Enabled toggle flips it back.

`go build ./...`, `go vet`, `gofmt`, `make proto-check`, and the affected package tests all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,18 @@ URL** appears under the Public MCP URL (computed by the gateway from its
 `--public-grpc-url` flag) — that value is what you feed another `fleet` as
 `FLEET_GATEWAY`. The toggles are independent: expose MCP, remote control, or both.
 
+A third, independent toggle — **Enable Webhook** — exposes this daemon's
+**automation webhook** endpoint through the same gateway. With it on, a read-only
+**Public Webhook URL** base appears (e.g. `https://gateway.example.com/webhook/<id>`),
+served on the gateway's public HTTP listener — no extra gateway flag needed. A
+remote system (CI, a SaaS webhook, `curl`) POSTs an event to
+`<public-webhook-url>/<name>`, where `<name>` is a [webhook trigger](#automation)
+you defined; the gateway relays it down the tunnel and fleetd fires that trigger's
+agents if the event passes the trigger's filter (a regex over the body, or a JSON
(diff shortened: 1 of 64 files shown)


BenjaminBenetti:
Merged latest `main` (resolved conflicts in `page_fleet_keys.go` and `automation_test.go` against #198's UX tweaks — kept both my `s: enable/disable` hint and #198's `a: add <noun>`/`TestTriggerWebhookURL`).

Also extended the feature to cover **webhook** triggers: #197 landed webhook delivery while this PR was open, so a `Disabled` gate is now also applied in `collectWebhookFires` (`webhook.go`) — a disabled webhook trigger still resolves its name (200, not 404) but never fires. Covered by `TestServeWebhook_DisabledTriggerDoesNotFire`.

PR #200, opened 2026-06-22
feat(automation): hand the triggering event to spawned agents

BenjaminBenetti:
## What

Automation-spawned agents now receive context about the event that triggered them. The static trigger prompt alone doesn't tell an agent *what just happened* — a webhook fired with *which* payload, or *when* a schedule ran. This wires that through.

At launch, each run's event payload is written into its instance at `/tmp/fleet-event-<timestamp>`, and a note is appended **after** the user prompt pointing the agent at that file:

- **webhook** triggers → the raw request body
- **schedule** triggers → the fire time (RFC3339)

Example appended note:

> This session was started automatically by the `"deploy"` webhook trigger. Its payload has been written to `/tmp/fleet-event-20260622T090000Z` in this instance — read that file for the event details.

## How

- **Thread the webhook body through.** It was read for filter matching and then discarded; `webhookFire` now carries it from the receiver to the scheduler, and `fireTriggerAgents` hangs a `triggerEvent` (kind, trigger name, fire time, webhook name, body) off each `watchedAgent`.
- **Write the payload into the instance** by reusing the fleet-copy exec seam (`copyIntoExecCommand`), streaming the payload as **base64 over stdin** and decoding in-container. This dodges the `MAX_ARG_STRLEN` (~128 KiB) limit a shell-embedded write would hit on a ~1 MiB webhook body, keeps arbitrary bytes intact, and stays PTY-safe over the codespaces backend's exec (which mangles raw binary). The write happens in the launch goroutine *before* the agent starts, so the file exists the moment it's referenced. Best-effort: a write failure still launches the agent (it just finds no file) rather than dropping the run.
- **Augment the prompt before substitution** so the note rides `${PROMPT}` exactly like the user prompt does — no keystroke injection.

## Tests

New `automation_event_test.go`:
- `payload()` for both trigger kinds
- `appendEventPrompt` (webhook/schedule/blank-prompt cases)
- `automationEventPath` format
- `writeAutomationEventFile` round-trips arbitrary bytes (binary/json/empty) through real `base64 -d`
- `fireTriggerAgents` attaches the right event on both paths
- end-to-end: the request body survives `serveWebhook` → scheduler hand-off

`go build`, `go vet ./...`, and `go test ./internal/server/ ./internal/fleet/` all pass.

### DIFF
--- a/internal/server/automation_event_test.go
+++ b/internal/server/automation_event_test.go
@@ -1,0 +1,172 @@
+package server
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
+	"github.com/BenjaminBenetti/fleet-man/internal/state"
+)
+
+func TestTriggerEventPayload(t *testing.T) {
+	now := time.Date(2026, 6, 22, 9, 30, 15, 0, time.UTC)
+
+	// Webhook: the raw request body is the payload, verbatim.
+	body := []byte(`{"action":"opened","number":7}`)
+	web := &triggerEvent{kind: fleet.TriggerWebhook, body: body, firedAt: now}
+	if got := web.payload(); !bytes.Equal(got, body) {
+		t.Fatalf("webhook payload = %q, want %q", got, body)
+	}
+
+	// Schedule: no external payload, so the fire time stands in.
+	sch := &triggerEvent{kind: fleet.TriggerSchedule, firedAt: now}
+	if got, want := string(sch.payload()), "2026-06-22T09:30:15Z\n"; got != want {
+		t.Fatalf("schedule payload = %q, want %q", got, want)
+	}
+}
+
+func TestAppendEventPrompt(t *testing.T) {
+	path := "/tmp/fleet-event-20260622T093015Z"
+
+	web := &triggerEvent{kind: fleet.TriggerWebhook, triggerName: "ci", webhookName: "ci"}
+	got := appendEventPrompt("fix the build", web, path)
+	for _, want := range []string{"fix the build", "\n\n---\n", `"ci" webhook trigger`, path, "read that file"} {
+		if !strings.Contains(got, want) {
+			t.Fatalf("webhook prompt missing %q:\n%s", want, got)
+		}
+	}
+
+	sch := &triggerEvent{kind: fleet.TriggerSchedule, triggerName: "nightly"}
+	got = appendEventPrompt("run the report", sch, path)
+	for _, want := range []string{"run the report", `"nightly" schedule trigger`, path} {
+		if !strings.Contains(got, want) {
+			t.Fatalf("schedule prompt missing %q:\n%s", want, got)
+		}
+	}
+
+	// A blank prompt becomes just the note (no leading separator/whitespace).
+	got = appendEventPrompt("   ", web, path)
+	if strings.HasPrefix(got, " ") || strings.Contains(got, "---") {
+		t.Fatalf("blank prompt should yield just the note, got:\n%s", got)
+	}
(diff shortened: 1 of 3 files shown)


BenjaminBenetti:
## 🧪 Manual QA — PR #200 (automation event payload)

**Verdict: ✅ PASS.** Drove the real daemon end-to-end for both trigger kinds. The triggering event is written into the spawned agent's instance and the prompt is correctly augmented with a pointer to it, exactly as the ticket describes. No defects found.

### Environment
Built `./bin/fleet` from the PR branch (`go build` clean). Ran the actual daemon + scheduler against a `file://` fixture repo whose devcontainer installs a fake `claude` (from `integration/fixture-automation`) that records the args it's launched with. Set `FLEET_AUTOMATION_TICK_INTERVAL=5s`. Agents launch into real Docker devcontainers — I inspected the spawned instances with `fleet exec`. *(This is behaviour-level QA, not the unit/integration suites.)*

### 1) Schedule trigger
Seeded a fleet with a `writer` agent + a `schedule` trigger (cron = current minute, prompt `QASCHEDULEMARKER77`). Started the daemon; its startup tick fired the trigger and spawned instance `writer-033000-1`.

- **Event file** `/tmp/fleet-event-20260622T033000Z` contained exactly the fire time:
  ```
  2026-06-22T03:30:00Z   (RFC3339, UTC, + trailing \n)
  ```
- **Filename** matches `firedAt` (`automationEventPath`).
- **Prompt the agent received** (captured from the launched command):
  > `... QASCHEDULEMARKER77 --- This session was started automatically by the "fire" schedule trigger. The time it fired has been written to /tmp/fleet-event-20260622T033000Z in this instance.`
  - Original prompt preserved, `---` separator, note **after** the prompt, path in the note = the real file path. ✅
  - System prompt (`be-terse`) still delivered — no regression. ✅

### 2) Webhook trigger (full path, via a real gateway)
Ran a local `fleet gateway` (cert-less h2c), enabled Remote MCP + Webhook in the daemon config, let the daemon register and receive its Public Webhook URL, then `POST`ed to `…/webhook/<id>/deploy` with body:
```json
{"event":"deploy","service":"checkout","ref":"v2.3.1","marker":"QAWEBHOOKBODY99"}
```
Response `received: 1 trigger(s) fired` (HTTP 200) → spawned instance `writer-033445-1`.

- **Event file** held the **raw request body, byte-for-byte** (81 bytes, no added newline) — confirms `payload()` returns the body for webhooks and the body threads receiver → scheduler → launch intact. ✅
- **Prompt note** used the webhook wording, pointing at the matching path:
  > `... QAWEBHOOKMARKER88 --- This session was started automatically by the "hooktrigger" webhook trigger. Its payload has been written to /tmp/fleet-event-20260622T033445Z in this instance — read that file for the event details.` ✅

### 3) Large / hostile payload (validates the base64-over-stdin design)
`POST`ed a **405,404-byte** body (≈3× the ~128 KiB `MAX_ARG_STRLEN` limit) containing newlines, unicode, quotes, backslashes, and shell metacharacters (`$(touch /tmp/PWNED)`, backticks, `; rm -rf /`).

- Event file inside the instance: **SHA-256 matched the source exactly** (`fa528bcc…`, 405404 bytes) → arbitrary bytes survive and the size far exceeds the argv limit a shell-embedded write would hit. ✅
- `/tmp/PWNED` was **not** created → the payload's shell metacharacters never reached a shell (decoded from stdin via `base64 -d`); no command injection. The body is not embedded in the prompt either. ✅

### Other observations
- Daemon log clean across all 3 runs — no `write event file failed` / `launch agent failed` / `create instance failed` warnings.
- Idle reaping still works (idle instances torn down after the timeout), so the new write step doesn't disrupt the existing lifecycle.
- Blank-prompt case (`appendEventPrompt` → note only) not exercised live but is covered by the new unit tests.

**Cleanup:** all QA containers/images, the daemon, the gateway, and scratch state were removed.

PR #201, opened 2026-06-22
feat(automation): edit-dialog field reorder, one-line prompt, instant-save

BenjaminBenetti:
Three UX tweaks to the automation agent/trigger dialogs.

## 1. Edit-trigger field reorder
Reordered the trigger dialog to a single canonical order, applied to both navigation (`visibleTriggerRows`) and rendering:

> Name → Type → Enabled → Prompt → [Webhook → Filter → Regex | JSON path+value] → [Cron] → Agents → [URL] → Save

Prompt moves up under Enabled, Agents drops to just above the (former) Save, and the derived webhook URL sits last as a copy-only footer.

## 2. One-line prompt preview
The editor-backed prompt / sys-prompt preview could wrap to a second line inside the 50-wide dialog box. `promptFieldPreview` now budgets the **whole** preview (first line + `(N lines)` suffix) to a width that fits the narrowest dialog row, truncating instead of wrapping.

## 3. Instant-save when editing
Editing an existing agent or trigger now saves immediately, matching every other settings page in fleet:

- Every change persists right away via new `autosaveTrigger` / `autosaveAgent` helpers (text-field commit, `$EDITOR` return, Type/Enabled/Filter/Backend selectors, agent multi-select).
- The **Save row is gone** in edit mode (navigation + render); `q`/`esc` just closes ("Close", not "Cancel").
- A validation failure (empty name, bad cron, incomplete webhook, …) surfaces inline and does **not** persist — the last-good state is kept, reusing the existing optimistic `persistAutomationSettings` revert.
- **Creating** a new agent/trigger is unchanged — the explicit Save still owns creation; `autosave*` is a no-op there.

Shared `triggerCandidate` / `agentCandidate` builders keep the autosave and explicit-save paths from drifting.

## Tests
- `go build ./...` + `go vet ./internal/tui/` clean.
- Added 6 tests: Save-row present-when-new / absent-when-editing (trigger + agent), instant-save on toggle and on field-commit (trigger + agent), and validation-failure-keeps-last-good.
- Full `internal/tui` and `internal/fleet` suites pass.

### DIFF
--- a/internal/tui/dialog_automation_trigger.go
+++ b/internal/tui/dialog_automation_trigger.go
@@ -344,7 +344,9 @@ func (fleetPage *fleetPage) cycleTriggerType(m *model) {
 	}
 	// The applicable rows changed; park the cursor back on the type selector.
 	st.row = trigRowType
-	fleetPage.autosaveTrigger(m)
+	// Quiet: a Schedule⇄Webhook switch makes the other type's required fields
+	// appear empty, so don't flash a validation error for that transient state.
+	fleetPage.autosaveTriggerQuiet(m)
 }
 
 func (fleetPage *fleetPage) cycleFilterType(m *model) {
@@ -355,7 +357,8 @@ func (fleetPage *fleetPage) cycleFilterType(m *model) {
 		st.filterType = fleet.WebhookFilterRegex
 	}
 	st.row = trigRowFilterType
-	fleetPage.autosaveTrigger(m)
+	// Quiet: a Regex⇄json-path switch swaps in different required fields.
+	fleetPage.autosaveTriggerQuiet(m)
 }
 
 func (fleetPage *fleetPage) toggleTriggerAgent(m *model) {
@@ -432,10 +435,14 @@ func (fleetPage *fleetPage) cancelAutomationTrigger(m *model) tea.Cmd {
 	st.input.Blur()
 	fleetPage.mode = viewNormal
 	// Editing is instant-save, so closing is just "done" — only a new (unsaved)
-	// trigger is actually discarded.
-	if st.editIdx < 0 {
+	// trigger is actually discarded. If an edit was left in an unsavable state,
+	// say so rather than closing silently (the last good state stayed on disk).
+	switch {
+	case st.editIdx < 0:
 		m.message = "Cancelled"
-	} else {
+	case fleetPage.triggerFormError(m) != nil:
+		m.message = "Closed; last change not saved: " + fleetPage.triggerFormError(m).Error()
+	default:
 		m.message = ""
 	}
 	return nil
@@ -471,7 +478,17 @@ func (fleetPage *fleetPage) triggerCandidate(f *fleet.Fleet) fleet.Trigger {
 // explicit Save button owns creation. A validation/RPC failure surfaces inline
 // and leaves the last-good persisted state intact (the in-memory revert lives in
 // persistAutomationSettings); a success clears the error.
-func (fleetPage *fleetPage) autosaveTrigger(m *model) {
+func (fleetPage *fleetPage) autosaveTrigger(m *model) { fleetPage.autosaveTriggerImpl(m, false) }
+
+// autosaveTriggerQuiet is autosave for a structural selector switch (Type or
+// Filter), which legitimately leaves the now-required fields empty mid-edit. It
+// persists once the trigger validates but never flashes a validation error for
+// the incomplete intermediate state — it clears any stale error instead, since a
+// switch changes which fields even apply. The close-time check (triggerFormError)
+// is the backstop that still tells the user if they leave it unsavable.
+func (fleetPage *fleetPage) autosaveTriggerQuiet(m *model) { fleetPage.autosaveTriggerImpl(m, true) }
+
+func (fleetPage *fleetPage) autosaveTriggerImpl(m *model, quiet bool) {
 	st := &fleetPage.triggerDlg
 	if st.editIdx < 0 {
 		return
@@ -483,7 +500,11 @@ func (fleetPage *fleetPage) autosaveTrigger(m *model) {
 	oldName := f.Settings.Triggers[st.editIdx].Name
 	newSettings, err := fleet.UpdateTrigger(f.Settings, oldName, fleetPage.triggerCandidate(f))
 	if err != nil {
-		st.errMsg = err.Error()
+		if quiet {
+			st.errMsg = ""
+		} else {
+			st.errMsg = err.Error()
+		}
 		return
 	}
 	if err := fleetPage.persistAutomationSettings(m, st.fleetName, newSettings); err != nil {
@@ -491,6 +512,20 @@ func (fleetPage *fleetPage) autosaveTrigger(m *model) {
 		return
 	}
 	st.errMsg = ""
+}
+
+// triggerFormError returns the validation error the current form would hit if
+// saved now, without persisting — used on close to tell the user when an
+// instant-save edit was left in an unsavable state. nil when it would save.
+func (fleetPage *fleetPage) triggerFormError(m *model) error {
+	st := &fleetPage.triggerDlg
+	f, ok := m.st.Fleets[st.fleetName]
+	if !ok || st.editIdx < 0 || st.editIdx >= len(f.Settings.Triggers) {
+		return nil
+	}
+	oldName := f.Settings.Triggers[st.editIdx].Name
+	_, err := fleet.UpdateTrigger(f.Settings, oldName, fleetPage.triggerCandidate(f))
+	return err
 }
 
 func (fleetPage *fleetPage) saveAutomationTrigger(m *model) tea.Cmd {
--- a/internal/tui/dialog_automation_trigger.go
+++ b/internal/tui/dialog_automation_trigger.go
@@ -117,15 +117,14 @@ func (fleetPage *fleetPage) openEditTriggerDialog(m *model, fleetName string, id
 }
 
 // visibleTriggerRows is the ordered list of currently-applicable field ids.
+// Order: Name, Type, Enabled, Prompt, then the type-specific rows (webhook:
+// name, filter, regex/json — schedule: cron), then the Agents list, then the
+// webhook URL (last, after Agents), then Save.
 func (fleetPage *fleetPage) visibleTriggerRows(m *model) []int {
 	st := &fleetPage.triggerDlg
-	rows := []int{trigRowName, trigRowType, trigRowEnabled}
-	for i := range fleetAgents(m, st.fleetName) {
-		rows = append(rows, trigRowAgentBase+i)
-	}
-	rows = append(rows, trigRowPrompt)
+	rows := []int{trigRowName, trigRowType, trigRowEnabled, trigRowPrompt}
 	if st.triggerType == fleet.TriggerWebhook {
-		rows = append(rows, trigRowWebhookName, trigRowWebhookURL, trigRowFilterType)
+		rows = append(rows, trigRowWebhookName, trigRowFilterType)
 		if st.filterType == fleet.WebhookFilterJSONPath {
 			rows = append(rows, trigRowJSONPath, trigRowJSONValue)
 		} else {
@@ -134,7 +133,19 @@ func (fleetPage *fleetPage) visibleTriggerRows(m *model) []int {
 	} else {
 		rows = append(rows, trigRowCron)
 	}
-	return append(rows, trigRowSave)
+	for i := range fleetAgents(m, st.fleetName) {
+		rows = append(rows, trigRowAgentBase+i)
+	}
+	if st.triggerType == fleet.TriggerWebhook {
+		rows = append(rows, trigRowWebhookURL)
+	}
+	// Editing an existing trigger is instant-save (like the settings page): every
+	// change persists immediately, so there is no Save row. A new trigger still
+	// needs an explicit Save to create it.
+	if st.editIdx < 0 {
+		rows = append(rows, trigRowSave)
+	}
+	return rows
 }
 
 func (fleetPage *fleetPage) moveTriggerRow(m *model, delta int) {
@@ -162,6 +173,7 @@ func (fleetPage *fleetPage) updateAutomationTrigger(m *model, msg tea.Msg) tea.C
 		switch key.String() {
 		case "enter":
 			fleetPage.commitTriggerField()
+			fleetPage.autosaveTrigger(m)
 			return nil
 		case "esc":
 			st.fieldActive = false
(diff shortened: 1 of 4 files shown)


BenjaminBenetti (commented):
🤖 **Automated review — PR #201**

Reviewed the full diff plus surrounding code (`dialog_automation_trigger.go`, `dialog_automation_agent.go`, `editor.go`, `automation_mutate.go`/`NormalizeTrigger`, and `persistAutomationSettings`). `go build ./...`, `go vet ./internal/tui/`, and the `internal/tui` + `internal/fleet` test suites all pass locally.

Overall this is clean, well-tested, and consistent with the existing instant-save convention (`persistFleetSettings` in `dialog_edit_fleet.go` does the same optimistic-apply-then-revert). A few things I specifically verified are correct:

- `agentCandidate` / `triggerCandidate` nicely de-dupe the autosave vs. explicit-save paths so they can't drift.
- `UpdateAgent` / `UpdateTrigger` replace the item **in place** (`newAgents[idx] = norm`), so `editIdx` stays valid across an instant-save rename — no index drift, and a subsequent autosave reads the new name as `oldName`. Good.
- Autosave only ever runs against committed field values: the `fieldActive` `enter` path calls `commit*Field()` (which sets `fieldActive=false`) before `autosave*`, and the toggle/cycle paths are unreachable while a field is active. No partial/in-progress text leaks into a save.
- The `promptFieldPreview` width math is right: at `promptPreviewWidth = 32` the agent `Sys prompt:` row is `marker(2) + "Sys prompt:"(11) + " "(1) + 32 = 46`, exactly the usable width inside the `Width(50)`+`Padding(1,2)` box, so it no longer wraps. The suffix is budgeted out of the same 32, so the total preview never exceeds it.

No correctness bugs found. A few **minor, non-blocking** observations:

**1. Instant-save fires a synchronous gRPC per change.** Each `autosave*` calls `persistAutomationSettings` → `setFleetSettingsRemote` → `mutate`, which is a blocking round-trip (up to `mutationTimeout` = 5s) on the Bubble Tea update loop. This matches the existing settings-page pattern, but instant-save raises the call frequency a lot: toggling several agents in the trigger multi-select, or repeatedly cycling Type/Filter/Backend, now issues one blocking RPC *per keystroke*. On a healthy local gateway it's invisible; on a laggy link the dialog can stall per keypress. Worth being aware of — debouncing/async would be a larger change and is probably out of scope here.

**2. Type/Filter switch surfaces a validation error immediately.** When editing, switching a Schedule trigger to Webhook (or Regex→json-path) instant-saves *before* the now-required fields exist, so `NormalizeTrigger` rejects it and the dialog flashes `webhook name is empty` / `webhook json path is empty` the moment you flip the selector — i.e. on a perfectly normal multi-step edit. It clears once the dependent fields are filled, but the transient error reads as alarming. Optional: suppress the inline error specifically on a type/filter switch (only show it on field-commit), or treat an incomplete type-switch as "not yet savable" rather than an error.

**3. After a rejected autosave, the form diverges from the persisted state, and closing is silent.** E.g. deselecting the last agent shows 0 checkboxes but disk keeps the previous set (correct "keep last good"), yet `q`/`esc` then sets `m.message = ""` and discards `st.errMsg` — so a user who made an invalid edit and closed gets no signal that their last change didn't persist. Behaviorally safe, but a user could believe a rejected edit was saved. Optional: surface a brief "discarded unsaved change" / keep the last error visible on close when `st.errMsg != ""`.

Nice touch extracting the shared candidate builders and reusing the existing optimistic-revert for the validation-failure path. None of the above blocks merging.

internal/tui/dialog_automation_trigger.go
```diff
@@ -331,16 +344,18 @@ func (fleetPage *fleetPage) cycleTriggerType() {
 	}
 	// The applicable rows changed; park the cursor back on the type selector.
 	st.row = trigRowType
+	fleetPage.autosaveTrigger(m)
```
BenjaminBenetti: In edit mode this autosaves the instant the user flips Type from Schedule→Webhook (and `cycleFilterType` does the same on Regex↔json-path). At that point the now-required webhook fields are still empty, so `NormalizeTrigger` rejects it and the dialog immediately shows `webhook name is empty` / `webhook json path is empty` during what is a normal multi-step edit. It self-heals once the fields are filled, but the flash of a validation error on a mere selector toggle is jarring. Consider only autosaving type/filter switches when the resulting trigger validates (or surfacing the error on field-commit rather than on the toggle).

BenjaminBenetti:
## 🧪 Manual QA — PR #201 (automation edit-dialog tweaks)

**Verdict: ✅ PASS.** I built `cmd/fleet`, ran the real TUI in tmux, seeded a `qa` fleet (2 agents `builder`/`reviewer`, 2 triggers `nightly` schedule + `on-pr` webhook) into `~/.fleet/state.json`, and drove the automation edit dialogs by hand (`m` → automations, `e` → edit). I verified persistence by inspecting `state.json` after each action. All three features work as described; no defects found.

### 1. Edit-dialog field reorder ✅
- **Schedule trigger** (`nightly`) renders in order: **Name → Type → Enabled → Prompt → Cron → Agents** (no URL row). Prompt now sits directly under Enabled, Agents dropped to the bottom.
- **Webhook trigger** (`on-pr`) renders: **Name → Type → Enabled → Prompt → Webhook → Filter → JSON path → JSON value → Agents → URL** — the derived `URL:` row is last, after Agents, as a copy-only footer. The regex variant (`Filter → Regex`) renders correctly too.
- **Agent** dialog renders: **Name → Command → Sys prompt → Backend**.
- In **edit** mode the **Save row is gone** and the footer close key reads **`[q/esc] Close`**. In **new** mode (`a` → "New agent"/"New trigger") the **`[ Save ]` row is present** and the close key reads **`[q/esc] Cancel`** — creation unchanged.

### 2. One-line prompt preview ✅
Seeded multi-line prompts with a long first line:
- Trigger prompt preview → `Prompt:  This is a deliberate…  (3 lines)` — single line, truncated, suffix fits.
- Agent **Sys prompt** (the narrowest field) → `Sys prompt: This is a deliberate…  (4 lines)` — single line, no wrap to a second row.

The whole preview (first line + `(N lines)`) stays within the box for both the wide and narrow dialog rows.

### 3. Instant-save when editing ✅
Every edit persisted to `state.json` immediately — confirmed by re-reading the file after each action:
- **Enabled toggle** (space) → `disabled:true` written, toggling back removed it.
- **Cron text-field commit** (`enter`) → new cron value persisted.
- **Filter cycle** json-path → regex → re-renders fields; after filling a valid regex, persisted with the stale `jsonPath`/`jsonValue` cleared.
- **Agent multi-select** (space) → `agentNames` updated, kept in the fleet's stored agent order.
- **Backend cycle** (`h/l`) → `devcontainer` → `coder` persisted.
- **Agent rename** `builder` → `builder2` → agent renamed **and both referencing triggers rewritten** (`nightly` and `on-pr`), then renamed back cleanly.

### Validation-failure-keeps-last-good ✅
- Invalid cron (`totally bad cron`) → inline error `trigger "nightly": cron expression must have 5 fields (got 3)`; `state.json` kept the last-good cron (not persisted).
- Empty regex after switching filter type → inline error `webhook regex is empty`; not persisted.
- Even with an unrelated valid change attempted while a field was invalid, nothing persisted and the last-good record stayed intact. Closing the dialog (`q`) discarded the in-progress invalid form and reopening showed the original valid trigger — **persisted state was never corrupted**.
- New-agent dialog autosave is correctly a **no-op**: cycling backend/fields added nothing to `state.json`; `q` showed "Cancelled".

### Sanity
Toggling back to the instance view (`m`) and general navigation work; nothing else appears broken.

### Minor notes (not blockers, not regressions)
- The agent **`Command:`** field still wraps to two lines inside the 50-wide box. This is the plain text field (`field()`), not the editor-backed prompt preview, so it's outside this PR's one-line fix — pre-existing behavior, flagging only for awareness.
- Instant-save requires the **entire** candidate to validate, so toggling a valid field (e.g. Enabled) while a *different* field is invalid updates the on-screen control but does **not** persist (last-good kept, error shown). This is the intended/correct behavior; just noting a user could briefly be surprised that a visible toggle didn't save while an error banner is up.
- I did not interactively exercise the **`$EDITOR`-return** autosave path (`applyEditorResult` → `autosaveAgent`/`autosaveTrigger`) since driving `vi` in tmux is fiddly; the same `persistAutomationSettings` mechanism is wired there and was exercised via the seven other autosave paths above.

*Tested manually against `55a25eb` by building and running the TUI; no automated tests were run.*

internal/tui/dialog_automation_trigger.go
BenjaminBenetti: Fixed in eb5655a. `cycleTriggerType`/`cycleFilterType` now call a quiet autosave variant: it persists once the trigger validates but no longer flashes a validation error for the transient state where the now-required fields are still empty. The close-time `triggerFormError` check is the backstop — if you leave the switch incomplete and close, you get "Closed; last change not saved: …" rather than a silent discard. Added `TestEditTriggerTypeSwitchDoesNotFlashError` to lock it in.

BenjaminBenetti:
Thanks for the thorough review. Addressed the two UX points in eb5655a:

**#2 — error flash on Type/Filter switch:** Structural switches (`cycleTriggerType`/`cycleFilterType`) now autosave *quietly* — they persist once the trigger validates again but never surface a validation error for the transient "required fields not filled yet" state. Field commits still give immediate inline feedback, as you suggested.

**#3 — silent discard on close:** Closing an edit that's been left unsavable now reports `Closed; last change not saved: <reason>` (a dry-run validator, `triggerFormError` / `agentFormError`, computed without an extra RPC). A clean edit still closes silently; the last good state always stays on disk. Applied to both the trigger and agent dialogs.

New tests: `TestEditTriggerTypeSwitchDoesNotFlashError`, `TestEditTriggerCloseSignalsUnsavedChange`, `TestEditTriggerCloseQuietWhenClean`, `TestEditAgentCloseSignalsUnsavedChange`.

**#1 — synchronous gRPC per change:** Leaving as-is intentionally — it matches the existing settings-page instant-save convention (`persistFleetSettings`), and debouncing/async would be a broader change better done across all the instant-save dialogs at once rather than just here.

internal/tui/dialog_automation_trigger.go
```diff
@@ -450,6 +471,75 @@ func (fleetPage *fleetPage) saveAutomationTrigger(m *model) tea.Cmd {
...
+	}
+	f, ok := m.st.Fleets[st.fleetName]
+	if !ok || st.editIdx >= len(f.Settings.Triggers) {
+		return
+	}
+	oldName := f.Settings.Triggers[st.editIdx].Name
+	newSettings, err := fleet.UpdateTrigger(f.Settings, oldName, fleetPage.triggerCandidate(f))
+	if err != nil {
+		if quiet {
+			st.errMsg = ""
```
BenjaminBenetti, line 504: `autosaveTriggerQuiet` clears `errMsg` unconditionally on *any* validation failure here, not only for fields the Type/Filter switch emptied. Sequence: clear Name (inline shows *"name required"*) → cycle Type → this branch wipes the error even though the form is still unsavable for an unrelated reason. `triggerFormError` on close re-surfaces it, so the persisted state is safe, but the inline error vanishing mid-edit is a little misleading. Consider only clearing when the residual error is about a field the switch removed (or re-running a non-quiet validate after the switch).

internal/tui/dialog_automation_trigger.go
```diff
@@ -411,33 +430,35 @@ func (fleetPage *fleetPage) commitTriggerField() {
...
 	fleetPage.mode = viewNormal
-	m.message = "Cancelled"
+	// Editing is instant-save, so closing is just "done" — only a new (unsaved)
+	// trigger is actually discarded. If an edit was left in an unsavable state,
+	// say so rather than closing silently (the last good state stayed on disk).
+	switch {
+	case st.editIdx < 0:
+		m.message = "Cancelled"
+	case fleetPage.triggerFormError(m) != nil:
+		m.message = "Closed; last change not saved: " + fleetPage.triggerFormError(m).Error()
```
BenjaminBenetti, line 444: Minor: `triggerFormError(m)` is evaluated twice — once in the `case` guard and again for `.Error()` — each rebuilding the candidate and running `NormalizeTrigger`. Bind it once:
```go
if err := fleetPage.triggerFormError(m); err != nil {
    m.message = "Closed; last change not saved: " + err.Error()
}
```
Same pattern in `cancelAutomationAgent` (lines 237–238).

BenjaminBenetti (commented):
## Automated review — PR #201

Solid, well-tested change. `go build ./...`, `go vet ./internal/tui/`, and the full `internal/tui` + `internal/fleet` suites all pass on the head commit (`eb5655a`). The render order and `visibleTriggerRows` navigation order stay in lockstep across every type/filter permutation, the shared `triggerCandidate`/`agentCandidate` builders genuinely keep the save and autosave paths from drifting, and rename-rewrite correctly reuses `fleet.UpdateAgent`/`UpdateTrigger` (both replace in place, so the `editIdx` → `oldName` recompute stays valid across successive autosaves). Test coverage is thorough: save-row presence, instant-save on toggle/commit, validation-keeps-last-good, type-switch-no-flash, and close-signals-unsaved.

A few non-blocking observations (two anchored inline):

1. **Quiet autosave can hide an *unrelated* inline error.** `autosaveTriggerQuiet` clears `errMsg` on any validation failure, so a pre-existing error not caused by the Type/Filter switch (e.g. an empty Name) silently disappears while the form is still unsavable. The close-time `triggerFormError` backstop recovers it, so nothing is lost on disk — it's only a transient UI wrinkle. (inline)

2. **Double `*FormError` evaluation** in both cancel paths (trigger + agent) — cheap, but easy to bind once. (inline)

3. **Instant-save fires a synchronous `setFleetSettingsRemote` gRPC on the bubbletea Update goroutine** for every field commit, selector cycle, agent toggle, Enabled toggle, and `$EDITOR` return. This matches the existing settings flows (`toggleTriggerDisabled`, `dialog_edit_fleet`), but the per-change frequency is much higher now — on a slow or unreachable gateway each of those keypresses blocks the TUI and can surface an RPC error inline. Worth confirming that's acceptable; not a blocker.

4. **`promptFieldPreview` still doesn't bound the placeholder.** The value is truncated to `promptPreviewWidth` (now 32), but the empty-state `placeholder` is returned untruncated. The agent `Sys prompt:` placeholder `(optional, injected into ${SYS_PROMPT})` (~39 cols) still exceeds the ~32-col budget and can wrap — the same wrap the PR sets out to remove, just in the empty state. Pre-existing; minor.

Nice work overall — the change is coherent and the intent is well-documented in the comments.

PR #204, opened 2026-06-23
test(webhook): end-to-end coverage for the json-path filter

BenjaminBenetti:
## Why

You weren't sure the json-path filter for webhook triggers was actually working. It is — but the **end-to-end** path had a real coverage gap.

The matcher itself is well covered: `TestMatchesWebhook_JSONPath` (23 cases) and the in-process handler test `TestServeWebhook_JSONPathMatchFires`. But the **full delivery stack** — real HTTPS client → gateway `/webhook/<name>` → `TagWebhook` tunnel → fleetd receiver → matcher → scheduler — was only ever exercised with a **regex** filter (`TestWebhookDeliveryEndToEnd`).

json-path is the branch that matters most here: unlike regex (which matches raw bytes), it has to **parse the delivered body as JSON**. If anything mangled the body in transport, the regex-only e2e test would never catch it.

## What

- Parameterize the e2e stack helper by trigger: `webhookStack` → thin wrapper over new `webhookStackWith(t, triggers...)`. No behavior change for the existing tests.
- Add `TestWebhookJSONPathDeliveryEndToEnd`, asserting **both directions** against one live stack:
  - **fires** when `$.ref == "refs/heads/main"`.
  - **does not fire** when `$.ref` differs but an *unrelated* field carries `refs/heads/main` (a substring/regex match would mis-fire) — proving the filter routes on JSON **structure**, not raw bytes, end to end.

## Test

```
go test ./internal/server/ -run 'TestWebhook|TestServeWebhook' -v
```

All webhook tests pass, including the new full-stack json-path case (~0.5s) and the existing regex e2e (now via the shared helper).

### DIFF
--- a/internal/server/webhook_remote_e2e_test.go
+++ b/internal/server/webhook_remote_e2e_test.go
@@ -27,11 +27,24 @@ import (
 // proves the whole path — HTTPS → gateway /webhook route → TagWebhook tunnel
 // stream → fleetd demux → webhook receiver → trigger match → scheduler enqueue.
 
-// webhookStack stands up the full real stack with one webhook trigger ("ci",
-// regex `"action":"opened"`) and returns the gateway-minted public webhook base
-// URL, a cert pool trusting the gateway, and the service (whose webhookFires
+// webhookStack stands up the full real stack with the default webhook trigger
+// ("ci", regex `"action":"opened"`) and returns the gateway-minted public webhook
+// base URL, a cert pool trusting the gateway, and the service (whose webhookFires
 // channel the test drains to confirm a fire).
 func webhookStack(t *testing.T) (publicWebhookBase string, pool *x509.CertPool, svc *service) {
+	t.Helper()
+	return webhookStackWith(t, fleet.Trigger{
+		Name: "ci", Type: fleet.TriggerWebhook, AgentNames: []string{"a"},
+		WebhookName: "ci", FilterType: fleet.WebhookFilterRegex, Regex: `"action":"opened"`,
+	})
+}
+
+// webhookStackWith is webhookStack parameterized by the fleet "alpha" webhook
+// triggers, so a test can exercise the delivery stack with any filter (regex or
+// json-path). The triggers are the only thing that varies between webhook e2e
+// tests; everything below — MCP loopback, TLS gateway, and the fleetd TagWebhook
+// tunnel — is identical real machinery.
+func webhookStackWith(t *testing.T, triggers ...fleet.Trigger) (publicWebhookBase string, pool *x509.CertPool, svc *service) {
 	t.Helper()
 	isolateFleetDir(t)
 	if err := os.MkdirAll(fleetpaths.Dir(), 0o700); err != nil {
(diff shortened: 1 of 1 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary** — Test-only PR, and a worthwhile one. It closes a real gap: the full delivery stack (HTTPS → gateway → `TagWebhook` tunnel → fleetd receiver → matcher → scheduler) was only ever e2e-tested with a *regex* filter, yet json-path is the branch that actually has to **parse the transported body as JSON** — exactly the thing a transport bug would break. The refactor and new test are clean, and the reasoning in the comments holds up against the implementation.

**Verified**
- `webhookStack` → `webhookStackWith(t, …)` is a pure, backward-compatible extraction; the two existing callers (`TestWebhookDeliveryEndToEnd`, `TestWebhookUnknownName404`) are unchanged in behavior.
- The negative case is genuinely discriminating: with `$.ref`/`refs/heads/main` configured, body `{"ref":"refs/heads/dev","note":"refs/heads/main"}` resolves `$.ref` → `"refs/heads/dev"` (no match), while a raw-byte regex for `refs/heads/main` *would* mis-fire on the unrelated `note` field. This matches `MatchesWebhook` / `jsonPathLookup` semantics in `internal/fleet/automation_match.go`.
- The timing note checks out. `serveWebhook` enqueues **before** `WriteHeader(200)` (`webhook.go:88‑95`), and the non-match path returns 200 **without enqueuing anything** (`webhook.go:74‑81`), so by the time the POST returns the fire/no-fire decision is already made — the 500ms drain window is safe (belt-and-suspenders, even stronger than the comment implies). `webhookFires` is buffered at 64 (`schedule.go:142`), so the matching send can't deadlock against the test's later channel drain.
- The first (matching) fire is fully drained in its own `select` before the second POST is sent, so no stale batch can leak into the non-match assertion.
- `go test ./internal/server -run 'TestWebhook|TestServeWebhook' -count=1` passes in **0.53s**; `go vet ./internal/server` is clean.

**Optional (non-blocking)**
- The PR's own motivation is "if anything mangled the body in transport." The positive case already proves the body arrives parseable, but a third sub-case posting a **non-JSON / malformed body** (expect 200 + no fire) would exercise the parse-failure branch end-to-end — arguably the most transport-sensitive path. Unit tests already cover invalid JSON, so this is purely about completing the e2e story.
- Minor consistency nit: this test uses `resp.Body.Close()` directly while the sibling `TestWebhookDeliveryEndToEnd` uses `defer resp.Body.Close()`. The direct close is actually the *better* choice here (two sequential requests in one function), so no change needed — just flagging the divergence in case uniformity is preferred.

Solid, well-commented addition. No correctness or coverage concerns.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed PR #204 — a test-only change adding end-to-end coverage for the json-path webhook filter. I checked it out, ran the suite (`go test ./internal/server/ -run 'TestWebhook|TestServeWebhook'`) and `go vet`: all green, including the new `TestWebhookJSONPathDeliveryEndToEnd` (~1s).

**Assessment: solid, no blocking concerns.** This is a genuine coverage gap well filled — the prior e2e path only ever ran a *regex* filter, which matches raw bytes, so it could never catch the json-path branch's `json.Decode` step being broken by something mangling the body in transport.

What I verified specifically:

- **All three assertions match production behavior** (`Trigger.MatchesWebhook`, `automation_match.go`): the exact-`$.ref` body fires; the body with `refs/heads/main` in an *unrelated* field does not (proving structure-vs-substring routing, the whole point); and the non-JSON body hits the `dec.Decode` error → `false` branch, returning 200 because the name is wired. That decode-failure path is the most transport-sensitive case and regex never reaches it — good incremental value.
- **The negative-case timing is sound, not flaky.** `serveWebhook` enqueues synchronously *before* `WriteHeader(200)`, and `webhookFires` is buffered (64), so the send never blocks and any fire is already queued by the time the client sees 200. The 500ms drain window for the "must not fire" assertions is therefore generous and deterministic — and a better choice here than the non-blocking `drainBatch` helper, which would be racier across the tunnel.
- **The refactor is clean and backward-compatible.** `webhookStack` → thin wrapper over `webhookStackWith(t, triggers...)`; the existing regex e2e is unchanged in behavior, and the doc-comment style matches the file's conventions.

**Minor (non-blocking) nit:**

- The PR description's "What" section says the new test asserts *"both directions"*, but the second commit (`d253e26`) added a third case — the unparseable-body branch. The commit message covers it; just a small drift between the PR body and the final code. Worth a one-line update to the description so reviewers know to expect the malformed-body assertion.

Nice work — the reasoning in the test comments (why 200-not-404, why the short drain window is safe) is exactly the kind of context that makes an async e2e test maintainable.

Issue #1, opened 2026-04-06
Multiple tmux sessions per instance.


BenjaminBenetti:
The user should be able to hit "space" to expand an instance line in the TUI. 

When expanded the user will see a tabbed in list of tmux sessions. The user then can select individual ones and hit enter to open the session. This works just like the regular workflow, if a session is already open then this one replaces it just like selecting a different instance.

- There should be a item that is just "+" the user can hit enter on it to add new sessions.
- When creating a session the user should be able to specify a name, though there should be an auto generated name.
- The user should be able to rename the session with "r" -> opens a dialog to change name

TUI layout 
```
> fleet-name 
         > instance name 
                 <icon> session name 
                 <icon> session 2 name 
```

Tmux Hotkey setup

The user should also be able to switch between sessions with ctrl+pageup and ctlr+pagedown This cycling should be reflected in the visuals of fleet (active session highlighted). Also include ctrl+shfit+t for new session 

Keys:
- ctrl+pageup - switch session up
- ctrl+pagedown - switch session down
- ctrl+shfit+t - create new session, and switch to it

PR #205, opened 2026-06-23
feat(backup): hourly fleetd state snapshots with 1-month retention

BenjaminBenetti:
## What

Adds an automatic state-backup loop to **fleetd**. Every hour the daemon snapshots the files that make up its own durable state into a single compressed archive, so a corrupted `state.json`, a botched write, or an errant destroy can be recovered from a recent point in time.

Backed-up files (flat, by base name):
- `~/.fleet/config.json`
- `~/.fleet/gateway_session.json`
- `~/.fleet/mcp.env`
- `~/.fleet/mcp.port`
- `~/.fleet/mcp.token`
- `~/.fleet/state.json`

Archives land at **`~/.fleet/backup/<date>/<hour>.tar.xz`** (e.g. `~/.fleet/backup/2026-06-23/14.tar.xz`).

## How

- New unconditional background loop `runBackupLoop`, launched next to the automation scheduler in `Serve` and stopped on shutdown via the same `hubCtx`. It snapshots once immediately, then every `backupInterval`.
- Compression is **xz** (best ratio of the common formats) via the **pure-Go `github.com/ulikunitz/xz`** writer — so the daemon needs no `xz` binary on the host, yet the result unpacks anywhere with the standard tools:
  ```
  tar -xJf ~/.fleet/backup/2026-06-23/14.tar.xz
  ```
- **Idempotent**: paths are keyed by `<date>/<hour>`, so a daemon restart within the hour just rewrites that hour's archive.
- **Atomic**: writes to a temp file + `rename`, so a reader never sees a partial archive. The backup tree is `0700` and archives are `0600` (they embed `mcp.token`).
- **Retention**: after each snapshot, archives older than **30 days** are pruned. Age is derived from the `<date>/<hour>` path (not mtime) for an exact boundary; emptied day-directories are removed; anything that isn't a `<date>/<hour>.tar.xz` is left untouched.
- **Robust**: missing sources are skipped; a pristine `~/.fleet` writes nothing rather than an empty archive.
- Interval/retention are overridable via `FLEET_BACKUP_INTERVAL` / `FLEET_BACKUP_RETENTION` (Go durations), mirroring the automation scheduler's test seams.

> **Restore is intentionally out of scope** — this PR only captures and prunes; a restore mechanic is a planned follow-up.

## Tests

`internal/server/backup_test.go`:
- round-trip (write → unpack, content + `0600` perms verified)
- missing-source skipping
- pristine `~/.fleet` writes nothing
- same-hour overwrite (one archive per hour, latest content wins)
- retention pruning incl. the boundary day (per-hour granularity) and that foreign entries survive

Also manually verified the produced archive passes system `xz -t` and `tar -xJf`.


internal/server/backup.go
```diff
@@ -0,0 +1,270 @@
...
+		return false, nil
+	}
+	data, err := os.ReadFile(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	hdr, err := tar.FileInfoHeader(info, "")
```
BenjaminBenetti, line 229: 🔍 **CodeReview Agent**

`hdr.Size` here is taken from `info` (the `os.Stat` at line 184), but the bytes written at line 209 come from the *separate* `os.ReadFile` at line 194. The backup sources are rewritten live by this same daemon — `state.Save`/`SaveConfig` do an in-place, non-atomic `os.WriteFile` on `state.json`/`config.json` — so if one of those writes lands between the stat and the read, `hdr.Size` won't match `len(data)` and `tar.Writer` fails the snapshot (`ErrWriteTooLong`, or `"missed writing N bytes"` at `Close`). That drops the whole hourly backup with only a `Warn`.

Make the header consistent with the bytes you actually archive:

```go
data, err := os.ReadFile(path)
// ... handle err (incl. os.IsNotExist) ...
hdr, err := tar.FileInfoHeader(info, "")
// ...
hdr.Name = filepath.Base(path)
hdr.Size = int64(len(data)) // size must match the bytes written below
```

Then a concurrent rewrite at worst captures the old-or-new file in full, instead of erroring out the snapshot.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the hourly fleetd state-backup loop (#205). Overall this is clean and well-documented: the temp-file + `rename` atomicity, hour-keyed idempotency, `0700`/`0600` perms on the secret-bearing archives, path-derived (not mtime) retention, and the env-var test seams that mirror the scheduler are all the right calls. Test coverage on round-trip, missing-source, pristine, same-hour-overwrite, and the retention boundary is solid. A few things worth a look, ordered by substance:

**1. Tar header size comes from a different read than the data (concurrent-write race) — see inline.** `addFileToTar` builds the header from `os.Stat` (so `hdr.Size` = the stat'd size) but writes bytes from a *separate* `os.ReadFile`. These sources are written live by this same daemon: `state.Save` and `SaveConfig` rewrite `state.json`/`config.json` **in place** via non-atomic `os.WriteFile(...)`, not temp+rename. If a snapshot's stat and read straddle one of those writes, `hdr.Size != len(data)` and `archive/tar` fails the whole snapshot (`ErrWriteTooLong`, or `"missed writing N bytes"` on the next `WriteHeader`/`Close`) — a silently-dropped hourly backup (only a `Warn` line). Fix is cheap: read first, then `hdr.Size = int64(len(data))` so the header is self-consistent with the bytes archived.

**2. Snapshots can capture a torn `state.json`/`config.json`.** Same root cause as #1: a concurrent in-place `os.WriteFile` is observable by the backup's `os.ReadFile` as a truncated/partial file. Since the explicit goal of this feature is recovering from a corrupted/botched `state.json`, a backup that itself captures a half-written `state.json` is a weak recovery point. The robust fix is to make `state.Save`/`SaveConfig` atomic (temp+rename) — which also resolves #1 for free. That's arguably outside this PR's scope, but this feature is precisely what makes the existing non-atomicity matter, so flagging it.

**3. Orphaned temp files are never reclaimed.** A crash (SIGKILL/power loss) mid-`writeBackup` leaves a `.backup-*.tar.xz.tmp` in the day dir (the deferred `os.Remove` won't run). `pruneBackups` intentionally skips anything not matching `<hour>.tar.xz`, so these are never cleaned — and a lingering temp file also keeps an otherwise-empty day dir from being removed (`len(rest) == 0` never holds, so the `os.Remove(dayPath)` is skipped forever). Consider sweeping stale `*.tmp` (on startup, or as part of prune).

**4. Test mutates a package global without restoring it.** `TestPruneBackupsRetention` assigns `backupRetention = 30*24*time.Hour` directly with no `defer` to restore. Harmless today (same as the default), but it couples test ordering to shared mutable state; `defer func(prev time.Duration){ backupRetention = prev }(backupRetention)` keeps it isolated.

**5. Minor coverage gap.** The "foreign entry survives" assertion only covers a non-date dir at the backup root (`notes/`). There's no test that a *day dir containing a non-archive file* is preserved (and not emptied/rmdir'd), even though prune both deletes archives and removes day dirs. One extra case would lock that behavior down.

Nothing here blocks the feature's intent — #1 is the one I'd resolve before merge. Nice work on the docstrings and the test seams.

BenjaminBenetti:
Thanks for the thorough review — addressed in e6822c2:

- **#1 (tar size race) — fixed.** `addFileToTar` now sets `hdr.Size = int64(len(data))` from the bytes actually read, so the header is self-consistent regardless of a concurrent in-place rewrite. A snapshot can no longer be dropped by `ErrWriteTooLong`/`"missed writing N bytes"`; at worst it captures the old-or-new file in full.
- **#3 (orphaned temp files) — fixed.** `pruneBackups` now sweeps `.backup-*.tmp` orphans in each day dir (safe: snapshots are serial and the tick's `writeBackup` has already finished by the time prune runs). This also unblocks empty-day-dir removal, which a lingering temp file previously prevented. Not counted as a pruned archive.
- **#4 (test global) — fixed.** `TestPruneBackupsRetention` restores `backupRetention` via `defer`; dropped the no-op `t.Setenv` (it ran after package init, so it never took effect).
- **#5 (coverage) — added.** A new case covers an expired day dir that retains a foreign file (archive pruned, dir + file survive), plus a dedicated test for orphan-temp sweeping.
- **#2 (torn `state.json`/`config.json`) — deferred, as you suggested.** The real fix is making `state.Save`/`SaveConfig` atomic (temp+rename), which also subsumes #1. That touches core state persistence and is broader than this feature, so I've left it as a follow-up rather than widen this PR's blast radius. With #1 in, a torn read at least no longer fails the snapshot.

CI was green before these commits; will confirm it stays green.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed after `e6822c2` (synchronize). This commit responds directly to the prior pass, so I verified each fix against the new code rather than trusting the changelog:

- **#1 — tar size race: resolved.** `addFileToTar` now reads the file first and sets `hdr.Size = int64(len(data))`, overriding the stat-derived size from `tar.FileInfoHeader`, so the header is self-consistent with the bytes actually written. A concurrent in-place `os.WriteFile` of `state.json`/`config.json` can no longer produce `ErrWriteTooLong` / `"missed writing N bytes"` and silently drop the hourly snapshot. ✅
- **#3 — orphan temp sweep: resolved.** `pruneBackups` deletes `.backup-*…\.tmp` orphans in each day dir before the archive parse, and since `writeBackup` always creates its temp inside the dated day dir (`os.CreateTemp(dir, …)`), every orphan is reachable by the sweep. It also unblocks empty-day-dir removal and is correctly *not* counted toward `removed`. Single-goroutine serialization (the tick's `writeBackup` completes before `prune` runs) makes the unconditional delete safe. ✅
- **#4 — test global isolation: resolved.** `TestPruneBackupsRetention` restores `backupRetention` via `defer`, and dropping the post-init `t.Setenv` is correct — it ran after package init, so it never took effect. ✅
- **#5 — coverage: resolved.** New assertions cover an expired day dir that retains a foreign `README` (archive pruned, dir + file survive), plus a dedicated `TestPruneBackupsSweepsOrphanTempFiles`. ✅
- **#2 — torn `state.json`/`config.json`: deferred, agreed.** Making `state.Save`/`SaveConfig` atomic (temp+rename) is the right home for this and is broader than this PR. One wording nit: the new docstring/response say "at worst we capture the old-or-new file in full," but because the source write is a non-atomic truncate-in-place, a concurrent `os.ReadFile` can still observe a **truncated/partial** file. The snapshot stays *valid* (Size matches the bytes), but the captured `state.json` may be torn — which is exactly what the deferred atomic-save fix closes. Worth keeping #2 on the follow-up list with that framing.

One minor edge still open (pre-existing, low severity): if every source passes the first `os.Stat` in `writeBackup` but all of them vanish before `addFileToTar` reads them, `count` ends at 0 yet an essentially-empty `.tar.xz` is still renamed into place for the hour and returned with `n==0` (so `backupTick` logs nothing). The "writes nothing when there's nothing to capture" guarantee only holds at the first stat pass. Not worth blocking on — just flagging.

Net: the follow-up cleanly resolves the actionable items from the last pass, the fixes look correct, and the #2 deferral is reasonable. Nice work.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — built and ran `fleetd` end-to-end; the hourly state-backup loop behaves exactly as the ticket describes.

### Environment
- Isolated Docker container (`golang:1.25-bookworm`), workspace mounted, fresh `HOME` so `~/.fleet` started empty.
- Built `./cmd/fleet` (`go build`, clean). Ran the real daemon: `HOME=… FLEET_BACKUP_INTERVAL=3s fleet server` inside tmux, driving the loop in seconds instead of an hour.
- Seeded `~/.fleet` with the six source files (recognizable content) plus a pre-existing backup tree to exercise retention.

### What I tested & observed

**1. Snapshot is written to the documented path with all sources** ✅
- Log: `backup: wrote snapshot path=~/.fleet/backup/2026-06-23/01.tar.xz files=6` — fires immediately on startup, then every interval.
- Archive lands at `~/.fleet/backup/<date>/<hour>.tar.xz` exactly as specified; all 6 files captured (config.json, gateway_session.json, mcp.env, mcp.port, mcp.token, state.json).

**2. Archive unpacks with standard tools, content matches** ✅
- The container had **no `xz` binary** when the daemon wrote the archive (pure-Go writer claim holds).
- `xz -t` → integrity OK. `tar -xJf 01.tar.xz` → extracts cleanly.
- Extracted as a **flat bag of base names** (no dir tree); byte-for-byte `diff` against all 6 live sources → MATCH on every file.

**3. Permissions / atomicity** ✅
- Fresh run: backup root `drwx------` (0700), day dir `drwx------` (0700), archive `-rw-------` (0600). `mcp.token` keeps 0600 inside the tar.
- No orphan `.backup-*.tmp` files left behind after repeated ticks (temp+rename is clean).

**4. Idempotent same-hour overwrite** ✅
- ~30 ticks over the test all rewrote the **same** `01.tar.xz`; the hour dir always held exactly **1** archive — no proliferation.

**5. Retention pruning** ✅ (`FLEET_BACKUP_RETENTION` default 30d, now=2026-06-23)
- Seeded `2026-05-01/03.tar.xz` + `2026-05-01/14.tar.xz` (>30d old) → both **pruned** (`pruned expired archives removed=2`).
- `2026-06-23/00.tar.xz` (recent) → **kept**.
- Foreign `2026-05-01/notes.txt` → **survived**, and its day-dir was correctly **not** removed (only emptied day-dirs are dropped).

**6. Env test-seams & lifecycle** ✅
- `FLEET_BACKUP_INTERVAL=3s` took effect (ticks 3s apart) — override works.
- SIGINT (Ctrl-C): daemon logged `fleet server stopping (signal)` and exited cleanly; **no further snapshots** after shutdown — the loop stops via the shutdown context.

### Notes (non-blocking)
- The "pristine `~/.fleet` writes nothing" path can't be reproduced against the live daemon because `fleetd` provisions its own state files on startup (so there's always something to capture); it's covered by the unit tests and the source-presence filter, which I read.
- I did not directly observe an *empty* day-dir being auto-removed (my aged dir retained the foreign `notes.txt`), but the non-empty-dir-preserved path was confirmed.

No defects or regressions found. Marking **qa-passed**.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Re-test on `e6822c2`** (`feat/fleetd-auto-backup`) — re-ran a full manual pass after the review fixes. **Verdict: ✅ PASS.**

I built `fleet` and drove the **real `fleetd` daemon** (`fleet server`) inside a throwaway `golang:1.25` Docker container with the documented test seams (`FLEET_BACKUP_INTERVAL`, `FLEET_BACKUP_RETENTION`), and inspected the on-disk results as an operator would (`xz -t`, `tar -xJf`). No code/unit tests — behavior only.

### What I tested & observed

**1. Hourly snapshot (happy path) — PASS**
- Seeded `~/.fleet` with all six sources, ran the daemon.
- Archive appeared at `~/.fleet/backup/2026-06-23/01.tar.xz` — correct `<date>/<hour>.tar.xz` layout.
- `xz -t` ✅ valid; `tar -tJf` lists exactly the 6 **flat base names** (`config.json`, `gateway_session.json`, `mcp.env`, `mcp.port`, `mcp.token`, `state.json`).
- Unpacked content **byte-matches** the live sources (all 6).
- Perms: archive `0600`, `backup/` and the day dir `0700` (token-bearing → owner-only). ✅

**2. Idempotent same-hour overwrite — PASS**
- 10+ ticks in the same hour produced **exactly one** `01.tar.xz` (no per-tick proliferation).
- Changed `state.json` live → next tick's archive reflected the new content (latest wins).

**3. Retention pruning + robustness — PASS** (covers fixes #3/#5)
- Seeded expired day dirs (`2026-04-0x`, retention `720h`), a within-retention day (`2026-06-20`), a foreign file, and an orphan `.backup-*.tmp`.
- After a tick: expired archives removed (`removed=4`), empty day dirs deleted, **`2026-06-20` and today kept**.
- Foreign `notes.txt` **and its day dir survived**. ✅
- Orphan `.backup-abc.tar.xz.tmp` **swept**, and notably **not counted** in `removed` (4, not 5). ✅

**4. Missing-source skipping / no empty archive — PASS**
- Fresh `~/.fleet`: daemon created only `mcp.{env,port,token}`; the snapshot captured exactly those **3** files (`files=3`) — absent sources skipped silently, no error, no empty archive.

**5. Integration / regression — PASS**
- Daemon starts cleanly with the new `runBackupLoop` (no panics, no `level=ERROR`); loop runs unconditionally with no TUI connected.
- Across dozens of ticks while I rewrote `state.json` concurrently, **every** snapshot succeeded (`files=6`, no `ErrWriteTooLong`) — consistent with the #1 self-consistent-tar-size fix.
- `SIGTERM` → daemon logged `fleet server stopping (signal)` and the backup loop **halted** (snapshot count froze), confirming it stops via `hubCtx`.

### Notes / scope
- **Restore** is explicitly out of scope for this PR — not tested.
- I could not catch the exact *zero-sources-present* instant against the live daemon (it creates the `mcp.*` files immediately on boot), but the equivalent guarantee — only present sources are archived and no empty archive is emitted — is verified by the partial-subset case (#4).

**Environment:** `golang:1.25-bookworm` container, `xz-utils` 5.4.1, daemon driven via tmux; `FLEET_BACKUP_INTERVAL=2s`, `FLEET_BACKUP_RETENTION=720h`. Container torn down after the run.

BenjaminBenetti:
Added the **`fleet_restore_backup` MCP tool** (commit b7c489e) — the restore-side counterpart to this feature, as requested.

It's **documentation-only** (mutates nothing): fleetd rewrites `~/.fleet` live, so the only safe restore is manual. The tool hands the calling agent everything it needs:
- `backup_dir` / `restore_into` — absolute `~/.fleet/backup` and `~/.fleet`
- `available` (archives that exist, newest first) + `latest`
- `contents` — the six files, each described (incl. the `mcp.token` secret caveat)
- `procedure` — ordered manual steps: stop every `fleet`/`fleetd` process → pick an archive → optional pre-restore copy → `tar -xJf <archive> -C ~/.fleet` → restart + verify
- `warning` — stop fleetd **before** unpacking

Named `fleet_restore_backup` to fit the `fleet_*` tool namespace; the admiral `SKILL.md` gets a "Recover daemon state" note + a reference line so it's discoverable. New unit tests cover the newest-first listing and the handler output (incl. the no-backups-yet case). Local `go build`, `go vet`, and the server + admiralskill suites are green.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Thorough, well-documented PR — atomic archive write+rename, `0700`/`0600` perms for the secret-bearing tree, path-derived (not mtime) retention boundary, hour-keyed idempotent writes, `envDuration*` test seams mirroring the scheduler, a deliberately doc-only restore tool that sidesteps a self-race, and genuinely good test coverage. The notes below are the few things worth a second look; none are blocking.

---

### 1. Snapshot can capture a *torn* `state.json` / `config.json` (the one thing the feature exists to protect against)

`addFileToTar` reads the sources with a raw `os.ReadFile`, i.e. **outside `state.mu`**. But `state.Save`/`SaveConfig` (`saveLocked` → `os.WriteFile(StatePath(), …)` and `SaveConfig` → `os.WriteFile(ConfigPath(), …)`) write **non-atomically** — `os.WriteFile` opens with `O_TRUNC` and then writes, so there is a window where a concurrent reader observes a truncated/empty/partial file. (Note `state.SaveArmada` already does the atomic temp-file + `rename` dance — the two hot files are the exception.)

So a snapshot that lands mid-write can archive a corrupt `state.json` — which directly undercuts the PR's stated goal ("a corrupted `state.json` … can be recovered"). The `addFileToTar` comment claims *"at worst we capture the old-or-new file in full"*; that's only true for the tar **header** (size is derived from the bytes read, so the archive stays well-formed) — it does **not** hold for the **contents**, which can be torn.

Probability is low (one hourly read vs. occasional mutation writes — though the immediate-on-start snapshot overlaps daemon-startup writes), but impact is high for a backup. Cheapest durable fix is to make the two writers atomic like `SaveArmada` already is (temp + `rename`) — that makes **every** reader, not just backup, see a complete file. Alternatively, read `state.json` through `state.Load()` (re-serialized) so the snapshot rides the same mutex. Worth at minimum softening the comment so it isn't read as a guarantee.

### 2. Rare empty-archive write contradicts the "writes nothing rather than an empty archive" guarantee

`writeBackup` decides to proceed from the `stat`-scan into `present` (line ~114), but the actual bytes are read later in `addFileToTar`. If every present source disappears between the scan and the read, `count == 0` yet the code still `MkdirAll`s the day dir and `rename`s a **valid-but-empty** `tar.xz` into place. Very unlikely for these files, but it's the exact case the early `len(present) == 0` guard is trying to prevent. Consider skipping the `rename` (and removing the just-made temp/day dir) when `n == 0`.

### 3. Minor

- `present := sources[:0:0]` (line 114): correct (cap 0 forces a fresh backing array on first `append`, so no aliasing of `sources`), but it's a subtle idiom whose zero cap also defeats any preallocation. `present := make([]string, 0, len(sources))` reads clearer and avoids regrowth.
- Shutdown isn't awaited: `Serve` launches `go svc.runBackupLoop(hubCtx)` with no `WaitGroup`, so an in-flight xz compression can be abandoned on shutdown, leaving a `.backup-*.tmp`. Self-healing (the next run's `pruneBackups` sweeps it) and consistent with `runScheduler`, so just noting — not asking for a change.

Nice work overall.

internal/server/backup.go
```diff
@@ -0,0 +1,414 @@
...
+// is a flat bag of ~/.fleet files, not a directory tree). A file that vanished
+// between the stat above and here is skipped rather than failing the whole
+// snapshot.
+//
+// The header Size is taken from the bytes actually read, NOT from the stat: the
+// sources are rewritten live by this same daemon (state.Save/SaveConfig do an
+// in-place, non-atomic os.WriteFile of state.json/config.json), so a rewrite
+// landing between stat and read would leave hdr.Size != len(data) and make
+// tar.Writer fail the whole snapshot. Deriving Size from the read keeps the
+// header self-consistent — at worst we capture the old-or-new file in full.
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This "at worst we capture the old-or-new file in full" claim holds for the tar **header** (size is taken from the bytes read, so the archive stays well-formed) but **not** for the **contents**: this `os.ReadFile` runs outside `state.mu`, and `state.Save`/`SaveConfig` write `state.json`/`config.json` non-atomically (`os.WriteFile` = `O_TRUNC` then write). A read that interleaves a write can capture a truncated/partial file — i.e. the snapshot can preserve a corrupt `state.json`, the very thing this feature is meant to recover from.

`state.SaveArmada` already writes atomically (temp + `rename`); making `saveLocked`/`SaveConfig` do the same fixes this for every reader, not just backup. (Alternatively, snapshot `state.json` via `state.Load()` so it rides the same mutex.) At minimum, reword this comment so it isn't read as a guarantee.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — re-ran after the `synchronize` push (head `b7c489e`, which adds the `fleet_restore_backup` MCP tool). All acceptance criteria behave as described.

### Environment
Built and ran the real `fleetd` inside a throwaway `golang:1.25` Docker container (workspace + Go module cache mounted). Notably the image has **no `xz` binary**, which let me confirm the daemon produces xz archives with no host `xz` dependency. Drove the loop fast via `FLEET_BACKUP_INTERVAL=3s`. Build: `go build ./cmd/fleet` succeeds (only needed `-buildvcs=false`, an artifact of the bind-mounted git dir — not a code issue).

### What I tested (all ✅)
| # | Behavior | Result |
|---|----------|--------|
| 1 | Backup loop starts unconditionally with the daemon; snapshots immediately then every interval | ✅ `backup: wrote snapshot` on startup, then every 3s (54+ ticks, all OK) |
| 2 | Archive path `~/.fleet/backup/<date>/<hour>.tar.xz` | ✅ `~/.fleet/backup/2026-06-23/02.tar.xz` |
| 3 | Permissions: tree `0700`, archive `0600` (embeds `mcp.token`) | ✅ `drwx------` dir, `-rw-------` archive |
| 4 | Valid xz, unpacks with standard tools | ✅ `xz -t` OK and `tar -tvJf`/`-xJf` work with the **system xz 5.8.1** (daemon used pure-Go writer) |
| 5 | Flat tar of base names, no directory tree | ✅ entries are bare `mcp.env`, `mcp.port`, `mcp.token`, … |
| 6 | Content round-trips | ✅ unpacked files match originals byte-for-byte |
| 7 | All 6 sources captured when present; missing ones skipped | ✅ `files=1` → `files=3` → `files=6` as `mcp.*`, then `config.json`/`state.json`/`gateway_session.json` appeared |
| 8 | Idempotent same-hour overwrite (one archive/hour) | ✅ 60 ticks all rewrote the single `02.tar.xz` |
| 9 | Retention prune (live daemon, default 30d) | ✅ see below |
| 10 | `fleet_restore_backup` MCP tool — read-only docs | ✅ see below |
| 11 | Clean shutdown stops the loop | ✅ SIGINT → "fleet server stopping (signal)", no ticks afterward |
| 12 | No warnings/errors/panics | ✅ log clean across all ticks |

### Retention prune (live daemon)
Seeded fixtures, then let the running daemon (default 30-day retention; cutoff ≈ `2026-05-24`) tick once:
- `2026-05-01/10.tar.xz` (≈53d old) → **pruned** (`backup: pruned expired archives removed=1`)
- `2026-05-01/keepme.txt` (foreign) → **survived**, and its day-dir was **kept** because it's non-empty
- `2026-06-20/05.tar.xz` (3d old) → **kept**
- `notadate/` (non-date dir) → **untouched**

### `fleet_restore_backup` MCP tool
Called it over the live local MCP HTTP endpoint (initialize → `tools/call`). It is read-only and mutated nothing. `available` listed exactly the **2 real archives, newest first** (`2026-06-23/02`, `2026-06-20/05`) — correctly excluding the pruned archive and the foreign `keepme.txt`/`notadate` entries — plus `latest`, the 6 documented file descriptions, the 5-step manual procedure (stop fleetd → pick archive → backup → `tar -xJf … -C ~/.fleet` → restart), and the "stop fleetd first / archive holds a secret" warning. The paths/commands it prints match the actual on-disk layout.

### Defects
None found.

---
_Manual behavior-level QA only; no code modified. QA container torn down after the run._

BenjaminBenetti:
Addressed the review in b34ebef:

- **#1 (torn `state.json`/`config.json`) — fixed at the root.** As you noted, `SaveArmada` already wrote atomically while the two hot writers didn't. I factored the temp+rename dance into a single `state.atomicWriteFile` and routed `saveLocked`, `SaveConfig`, **and** `SaveArmada` through it. Every unlocked reader (backup loop, state poller) now sees the old or new file in full — never a torn one — so a snapshot can no longer preserve a corrupt `state.json`. The `addFileToTar` comment is reworded accordingly (the contents guarantee now actually holds, not just the header).
- **#2 (rare empty-archive write) — fixed.** `writeBackup` now skips the rename when `n==0` (every present source vanished between the scan and the read), honoring the "writes nothing rather than an empty archive" contract, and removes the day dir if this call just created it.
- **#3 minors:** swapped `sources[:0:0]` for `make([]string, 0, len(sources))`. Left the shutdown-not-awaited point as-is per your note (self-healing via the next prune's temp sweep, and consistent with `runScheduler`).

New tests cover `atomicWriteFile` (round-trip, overwrite, perms, no leftover temp files) and a Save/Load loop asserting unlocked reads always see a complete `state.json`. Local `go build`/`go vet` and the state + server + admiralskill suites are green.

internal/server/backup.go
```diff
@@ -0,0 +1,423 @@
...
+	if err != nil {
+		return "", 0, err
+	}
+	if n == 0 {
+		// Every source vanished between the presence scan and the read. Honor the
+		// "writes nothing rather than an empty archive" contract: the deferred
+		// cleanup drops the temp file, and we remove the day dir if this call just
+		// created it (Remove no-ops when earlier-hour archives still live there).
+		_ = os.Remove(dir)
+		return "", 0, nil
```
BenjaminBenetti, line 189: 🔍 **CodeReview Agent**

`os.Remove(dir)` here runs while the temp file is still present: the temp cleanup is the function-level `defer os.Remove(tmpName)`, which fires only when `writeBackup` returns — i.e. *after* this inline call. So at this point `dir` still contains the `.backup-*.tmp` file, `Remove` fails with `ENOTEMPTY`, and the empty `<date>/` directory is **not** removed, contrary to the comment above ("we remove the day dir if this call just created it"). Impact is low because the `pruneBackups` call later in the same `backupTick` reaps the empty dir anyway — but consider removing/closing the temp file before attempting the dir removal (or dropping this line and letting prune handle it), and update the comment to match.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the hourly backup loop, the retention/prune logic, the documentation-only `fleet_restore_backup` MCP tool, and the `atomicWriteFile` refactor. This is a careful, well-documented change with strong test coverage (round-trip, missing sources, same-hour overwrite, retention boundary day, orphan-temp sweep, restore docs). The atomic-write centralization is a clean fix for the torn-read concern, and the secret handling (0700 tree / 0600 archives, hour-keyed idempotency) is solid. A few things worth a look — none blocking:

**1. `writeBackup` — the `n == 0` branch leaves an empty day dir behind; its comment is inaccurate (minor)**
In the "every source vanished between the presence scan and the read" branch, `os.Remove(dir)` runs while the temp file is *still* in `dir`. The temp cleanup is the function-level `defer os.Remove(tmpName)`, which runs only when `writeBackup` returns — i.e. *after* this inline `os.Remove(dir)`. So the dir is non-empty at that moment, `Remove` fails with `ENOTEMPTY`, and the empty `<date>/` directory is left behind — the opposite of what the comment claims ("we remove the day dir if this call just created it"). In practice it's self-healed by the `pruneBackups` call later in the same `backupTick`, so impact is low. Consider removing/closing the temp before the dir (or dropping the dir-removal entirely and letting prune reap it), and fixing the comment. Inline note added.

**2. Only 2 of the 6 snapshotted files are actually written atomically (consistency / follow-up)**
The last commit routes `state.json` / `config.json` / armada through `atomicWriteFile`, closing the torn-read window the backup loop exposed — good. But the snapshot also captures `gateway_session.json`, `mcp.env`, `mcp.port`, and `mcp.token`, and those are still written with a plain `os.WriteFile` (O_TRUNC + write) in `internal/server/mcp.go` and `internal/server/remote/session.go`. A snapshot landing during a `gateway_session.json` rewrite (the most frequently re-written of the four) can still capture a torn file — the exact failure mode this PR set out to prevent. Worth either routing those writers through the same atomic helper (would need exporting it from `state`, or a small shared util) or explicitly calling them out as out-of-scope follow-ups in the PR description.

**3. Duplicated tree-walk between `pruneBackups` and `listBackupArchives` (minor / DRY)**
Both re-implement the same walk: iterate day dirs → validate `<date>` → validate `<hour>.tar.xz` (`Atoi`, 0–23). Factoring the per-archive parse into a single helper would remove the duplication and keep the two from drifting. Minor related nit: `pruneBackups` parses the day with `time.ParseInLocation(..., now.Location())` while `listBackupArchives` uses `time.Parse` (UTC). Harmless today (the listing only validates format), but inconsistent.

**Nits (non-blocking):**
- `atomicWriteFile` / `writeBackup` don't `fsync` the file or the parent dir before/after `rename`, so writes are atomic for concurrent *readers* but not crash-*durable* (a power loss right after rename can still lose the new bytes on some filesystems). This matches the pre-existing `SaveArmada` behavior, so fine to keep — just flagging since "durably" / "A crash mid-write leaves only a stale temp file" appears in the doc comment.
- `os.Chmod(tmpName, 0o600)` in `writeBackup` is redundant — `os.CreateTemp` already creates the file 0600 — but harmless.

Confirmed the new `time.NewTicker(backupInterval)` is safe against a malformed/zero `FLEET_BACKUP_INTERVAL`: `envDurationDefault` only accepts `d > 0` and otherwise falls back to the default, so the ticker can't panic. 👍

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — `feat(backup): hourly fleetd state snapshots with 1-month retention` (commit `b34ebef`)

Manual, behavior-level QA. Built `cmd/fleet` and ran the real `fleet server` daemon inside a throwaway `golang:1.25` Docker container (`HOME` controlled, `FLEET_BACKUP_INTERVAL=2s` so the hourly loop fires fast), then drove it as a user/agent would and inspected the on-disk results. No unit/integration tests were run.

### What I tested & observed

**1. Snapshot loop produces the archive at the documented path** ✅
Daemon wrote `~/.fleet/backup/2026-06-23/02.tar.xz` (`<date>/<hour>.tar.xz`) and logged `backup: wrote snapshot … files=6`.

**2. All six state files captured, flat by base name** ✅
`tar -tJf` lists exactly: `config.json gateway_session.json mcp.env mcp.port mcp.token state.json` (no directory tree).

**3. Unpacks with standard tools / pure-Go xz is valid** ✅
`xz -t` passes, and `tar -xJf <archive> -C <dir>` (the exact command in the PR/docs) extracts cleanly — the daemon needs no `xz` binary, yet the output unpacks with system tar+xz.

**4. Byte-perfect round-trip + perms preserved** ✅
Every extracted file `cmp`-matches the live `~/.fleet/<file>`. Per-file modes preserved (`mcp.*` `0600`, `config/state` `0644`). My sentinels (`QA-CONFIG-SENTINEL-205`, `QA-STATE-SENTINEL-205`) survived intact.

**5. Tree/archive permissions** ✅
`~/.fleet/backup` `0700`, day dir `0700`, archive `0600` (secret-bearing, as designed).

**6. Idempotent same-hour overwrite, latest-content-wins** ✅
77 ticks over the run → still exactly **one** archive for the hour. After I modified `config.json` on disk mid-run, the next tick's archive reflected the new content (`QA-CONFIG-SENTINEL-205-MODIFIED`).

**7. Retention pruning (default 30d), exact boundary, foreign-safe** ✅
Seeded `2026-05-01/10.tar.xz` (53d old), `2026-04-15/09.tar.xz` + a foreign `notes.txt`, `2026-06-22/05.tar.xz` (1d old), a top-level `README`, and a `not-a-date-dir/`. Next tick logged `pruned … removed=2`:
- old archives removed; the emptied `2026-05-01` day dir removed;
- `2026-04-15` **kept** because its foreign `notes.txt` remained (archive pruned, foreign file untouched);
- recent `2026-06-22/05` **kept**; current `2026-06-23/02` **kept**;
- top-level `README` and `not-a-date-dir/` left untouched.

**8. `fleet_restore_backup` MCP tool — documentation-only & read-only** ✅
Called it over the live MCP Streamable-HTTP endpoint (bearer auth). It mutated nothing (backup tree unchanged afterward), listed only the valid archives **newest-first** (correctly excluding pruned/foreign entries), set `latest` to the newest, and returned the correct `backup_dir`/`restore_into`, documented contents, and the step-by-step manual restore procedure (`tar -xJf … -C ~/.fleet`).

**9. Loop is unconditional and stops on shutdown** ✅
Snapshots wrote with no TUI connected. After the daemon process exited, the snapshot-log count held steady (no further writes) — the loop ends with the daemon via `hubCtx`.

**10. Atomic state writes** ✅ (behaviorally) — every captured `config.json`/`state.json` round-tripped whole/non-torn across 77 concurrent snapshot+write cycles. No `backup … fail/error/warn` lines appeared in the daemon log.

### Notes (non-blocking)
- The "pristine `~/.fleet` writes nothing" contract isn't reachable against a live daemon (the daemon itself creates `mcp.port/token/env` on startup, so sources always exist); it's covered by the unit suite. Not a defect.
- During my forced full-daemon teardown the container's broad `pkill` wrapper hung briefly; this is the daemon's overall shutdown path in a backend-less container, unrelated to the backup feature — the backup goroutine stopped immediately when the process died.

All acceptance criteria from the PR description were exercised against the running app and behaved as specified. Applying `qa-passed`.

BenjaminBenetti:
Addressed round 3 in c747094:

- **#1 empty-dir bug — fixed.** Right call — `os.Remove(dir)` ran while the temp file was still in `dir` (the deferred temp cleanup only fires on return), so it failed `ENOTEMPTY`. Now the temp is removed first, then the dir. Also dropped the redundant `os.Chmod` (`os.CreateTemp` already makes 0600).
- **#2 only 2 of 6 atomic — fixed for all six.** Moved the temp+rename dance into a leaf package `internal/atomicfile` and routed every writer through `atomicfile.Write`: `state.json`/`config.json`/`armada.json` plus `gateway_session.json` (`remote/session.go`) and `mcp.port`/`mcp.token`/`mcp.env` (`mcp.go`). `gateway_session.json` — the one you flagged as most-rewritten — is no longer torn-readable.
- **#3 DRY — fixed.** Extracted `parseArchiveHour` as the single "what counts as a `<hour>.tar.xz` archive" check, used by both `pruneBackups` and `listBackupArchives`.
- **Nits:** softened the helper's doc to claim reader-atomicity (not fsync durability), matching the prior `SaveArmada` behavior; removed the redundant `Chmod`. Left shutdown-not-awaited as-is per your earlier note.

New tests: `atomicfile` round-trip + missing-parent-dir error; a state Save/Load loop asserting unlocked reads always see a complete `state.json`. `go build`/`go vet` and the state, atomicfile, server, remote, and admiralskill suites are green locally.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

## Review — hourly fleetd state snapshots (#205)

Thorough, well-structured change. The atomicity story is coherent end-to-end: every unlocked-read source file (`config.json`, `state.json`, `gateway_session.json`, `mcp.{port,token,env}`) is converted to `atomicfile.Write`, so the lock-free backup reader can never observe a torn write. The design choices are sound and well-documented — pure-Go `xz` (no host binary, still `tar -xJf`-able), hour-keyed idempotent paths, atomic temp+rename for the archive itself, path-derived (not mtime) retention boundary, `0700`/`0600` perms protecting the embedded `mcp.token`, orphan-temp sweeping in prune, and a deliberately docs-only restore tool that refuses to mutate live state. Test coverage is excellent (round-trip, missing-source skip, pristine no-op, same-hour overwrite, retention boundary + foreign-entry survival, orphan-temp sweep, listing order, restore docs). `mcpRestoreBackup`'s `struct{}` input and the `runBackupLoop` loop shape both match existing conventions (`mcpVersion`, `runScheduler`).

No blocking issues. A few questions / minor notes — all non-blocking:

**1. `armada.json` is atomic-converted but not snapshotted (question).** `SaveArmada` is switched to `atomicfile.Write` in this PR, and `armada.json` lives in `~/.fleet` and embeds remote-gateway bearer tokens — yet it's absent from `backupSources()`. Excluding it is defensible (it's arguably the *client's* remote-connection registry rather than fleetd's own durable state), but since it's durable and secret-bearing, was the omission intentional? See inline note on `backup.go`.

**2. The immediate startup snapshot races the MCP-file writers (minor).** `go svc.runBackupLoop(hubCtx)` is launched in `Serve` before `startMCPServer` writes `mcp.port`/`mcp.token`/`mcp.env`, and `runBackupLoop` snapshots once immediately. So the startup-hour archive can capture a partial set (missing the `mcp.*` files). Low impact — those files are host-local and regenerated on every daemon start, and every subsequent hourly archive includes them — but worth a conscious decision. See inline note on `server.go`.

**3. Rare empty-base-dir leak (very minor).** In `writeBackup`, if sources exist at the presence scan but all vanish before the read (`n == 0` after compression), the code `MkdirAll`s `~/.fleet/backup/<date>` and then removes only the `<date>` dir, leaving an empty `~/.fleet/backup` behind. Harmless (owner-only, reused by the next snapshot, tolerated by prune/list), and the common pristine case is correctly handled by the earlier `present == 0` early-return — just noting the edge doesn't quite honor the "writes nothing" intent.

Nice work overall.

internal/server/backup.go
```diff
@@ -0,0 +1,434 @@
...
+	// retention of 1 month"). Anything older than now-retention is pruned after
+	// each snapshot. Overridable via FLEET_BACKUP_RETENTION for tests.
+	backupRetention = envDurationDefault("FLEET_BACKUP_RETENTION", 30*24*time.Hour)
+)
+
+// backupSources returns the absolute paths of the ~/.fleet files captured in
+// each snapshot. Resolved per call (not cached) because the paths derive from
+// $HOME, which tests rebind. Missing files are skipped at archive time, so a
+// pristine ~/.fleet that has not yet grown all of these is fine.
+func backupSources() []string {
```
BenjaminBenetti, line 52: 🔍 **CodeReview Agent**

Question: `armada.json` is converted to `atomicfile.Write` in this PR (`SaveArmada`) and lives in `~/.fleet` alongside these sources, embedding remote-gateway bearer tokens — but it's not listed here, so it's never snapshotted. Excluding it is reasonable if you consider it the client's remote-connection registry rather than fleetd's own durable state, but given it's durable and secret-bearing, was the omission deliberate? If so, a one-line comment here noting *why* `armada.json` is intentionally out of the backup set would prevent it being "fixed" later by mistake.

internal/server/server.go
```diff
@@ -85,6 +85,11 @@ func Serve(ctx context.Context) error {
 	// agents. Runs unconditionally — cron triggers must fire even when no TUI is
 	// connected — and stops on shutdown via hubCtx.
 	go svc.runScheduler(hubCtx)
+	// State backup loop: snapshots ~/.fleet's durable state (config/state/mcp
+	// files) to ~/.fleet/backup/<date>/<hour>.tar.xz every hour and prunes
+	// archives older than the retention window. Unconditional and stops on
+	// shutdown via hubCtx.
+	go svc.runBackupLoop(hubCtx)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Minor ordering note: this goroutine is launched here, but `startMCPServer` (which writes `mcp.port`/`mcp.token`/`mcp.env`) runs later in `Serve`. Since `runBackupLoop` takes a snapshot immediately on entry, the startup-hour archive can race those writers and capture a partial set (missing the three `mcp.*` files). Impact is low — they're host-local and regenerated on each daemon start, and every later hourly archive includes them — but if you want the very first snapshot to be complete, launching this loop after `startMCPServer` (or having it run its first tick once MCP discovery files exist) would close the gap. Not blocking.

BenjaminBenetti:
Addressed round 4 (all non-blocking) in 5f2285b:

- **#2 first-snapshot race — fixed.** Moved `go svc.runBackupLoop(hubCtx)` to *after* `startMCPServer` in `Serve`, so the immediate first-tick snapshot already sees the freshly-written `mcp.port`/`mcp.token`/`mcp.env` instead of racing them.
- **#1 armada.json — intentional, now documented.** It wasn't in the issue's file list and it's the *client's* remote-connection registry rather than fleetd's own state, so it's deliberately out of the self-backup set. Added a comment in `backupSources()` saying so (and why), so it isn't re-added by mistake now that it's an `atomicfile` writer.
- **#3 empty-base-dir leak — closed.** The `n==0` path now also removes `~/.fleet/backup` if this call just created it, fully honoring "writes nothing rather than an empty archive".

`go build`/`go vet` and the server, state, and atomicfile suites are green locally.

BenjaminBenetti:
Follow-up: **`armada.json` is now included in the backup set** (commit ee46564) — re: the round-4 question about it. It's durable `~/.fleet` state (remote-fleet registry with per-fleet gateway URL + bearer token) and worth recovering, so it's in. Reverted the "intentionally excluded" note, added it to `backupSources()` and to `fleet_restore_backup`'s documented contents (flagged SECRET, like `mcp.token`); the existing 0600-archive / 0700-tree perms already cover it. Round-trip + restore-doc tests bumped to 7 files. Green locally.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — manual QA of the hourly fleetd state-backup loop against commit `c747094`.

All testing ran inside an isolated throwaway Docker container (`golang:1.25-bookworm`, `xz-utils`/`tar` installed) with the PR branch built (`go build ./cmd/fleet`) and the real daemon (`fleet server`) driven there. `FLEET_BACKUP_INTERVAL=3s` was used so hourly behavior could be observed in seconds. The MCP tool was exercised over the live Streamable‑HTTP endpoint with a proper MCP client (bearer auth).

### What I tested & observed

**1. Snapshot creation & path layout** ✅
Daemon wrote `~/.fleet/backup/2026-06-23/03.tar.xz` (matches `<backup_dir>/<date>/<hour>.tar.xz`). Log: `backup: wrote snapshot ... files=3`.

**2. Permissions (embeds the secret `mcp.token`)** ✅
Backup tree `drwx------` (0700), archive `-rw-------` (0600). Extracted files retained `0600`.

**3. Archive is valid with standard host tools** ✅
- `xz -t <archive>` → OK
- `tar -tJvf` lists files as flat **base names** (no dir tree)
- `tar -xJf` extracts cleanly; extracted bytes `cmp`‑identical to the live `~/.fleet/*` files (`mcp.token`, `mcp.port`, `mcp.env`, and a `state.json` marker round‑tripped exactly)

**4. Captures all six sources when present** ✅
With `config.json`, `gateway_session.json`, `state.json` also present, the next tick logged `files=6` and the archive listed all six. With only the mcp.* files present it logged `files=3` (missing sources skipped, not failed).

**5. Idempotent same‑hour overwrite** ✅
Every 3s tick rewrote the *same* `03.tar.xz` — one archive per clock hour, latest content wins, no duplicates.

**6. Retention pruning (default 30 days)** ✅
Planted a mix and observed one tick (`backup: pruned expired archives removed=2`):
- `2026-01-01/12.tar.xz` (only archive in its dir) → pruned **and** the emptied day‑dir removed
- `2026-02-01/08.tar.xz` → pruned, but the foreign `README` beside it **and** its day‑dir were kept
- `2026-06-22/23.tar.xz` (within window) → kept
- foreign `2026-06-23/notanarchive.txt` → kept

**7. Per‑hour boundary granularity** ✅
On the exact cutoff day (`2026-05-24`, cutoff ≈ 03:0x): hours `01` and `03` (≤ cutoff) pruned; hours `04` and `23` (> cutoff) kept — confirming age is derived from the `<date>/<hour>` path, not mtime.

**8. `fleet_restore_backup` MCP tool (documentation‑only)** ✅
Via a live MCP client over `http://127.0.0.1:<port>` with bearer auth: tool is registered, `isError=false`, and returns accurate `backup_dir` / `restore_into` / `path_layout` / `compression`, the 6 documented file descriptions, the 5‑step manual procedure, the secret warning, and an `available` list of the real archives **newest‑first** (`latest` = newest). It correctly reflected the post‑prune tree and mutated nothing (read‑only).

**9. No orphan temp files** ✅ — no `.backup-*.tmp` left in the tree (clean temp+rename atomic writes).

**10. Regression sanity** ✅
`fleet list` / `fleet status` work against the daemon (empty fleet state); the `mcp.*` discovery files are now written via the new `atomicfile.Write` path with no issues; daemon stayed healthy throughout.

### Notes (not defects)
- A transient `automation: load state failed ... cannot unmarshal array into ... map[string]*fleet.Fleet` warning appeared **only** because I deliberately planted a schema‑invalid `state.json` to test byte‑level capture; it cleared as soon as I removed my test file. The backup loop snapshots bytes by design and is independent of state validity.
- `fleet version` printed an empty string, but that's an artifact of building with `-buildvcs=false` / no version ldflags in this QA container — unrelated to this PR.

All acceptance criteria from the PR description were verified by actually running the daemon and inspecting/unpacking the produced archives.

internal/server/backup.go
```diff
@@ -0,0 +1,438 @@
...
+			"2. Pick the archive to restore from `available` (newest first); `latest` is the most recent. Each is an absolute path to a <date>/<hour>.tar.xz.",
+			"3. (Recommended) Snapshot the current settings dir before overwriting, e.g. `cp -a " + fleetDir + " " + fleetDir + ".pre-restore`.",
+			"4. Unpack the archive over the settings dir: `tar -xJf <archive> -C " + fleetDir + "`. The archive holds base names only, so each file lands back at its original ~/.fleet/<file> path, overwriting the current copy.",
+			"5. Restart the daemon — run any `fleet` command (it auto-spawns fleetd) or start `fleet server` directly — then verify with the `fleet_version` and `fleet_status` tools.",
+		},
+		Warning: "Stop fleetd BEFORE unpacking. It rewrites these files continuously and snapshots hourly, so restoring under a live daemon can corrupt state.json or have your restore immediately overwritten. The archives contain mcp.token (a secret) — handle them accordingly.",
```
BenjaminBenetti, line 440: 🔍 **CodeReview Agent**

Non-blocking doc nit, surfaced by the latest push: now that `armada.json` is in `backupSources()`, each archive embeds **two** secrets (its per-fleet gateway bearer tokens *and* `mcp.token`), but this top-level `Warning` still names only `mcp.token`. The `Contents` entry above flags `armada.json` as SECRET, so the information is present — it's just that the one line an agent is most likely to surface verbatim understates the sensitivity.

Consider widening it, e.g. `…The archives contain secrets (mcp.token and armada.json's bearer tokens) — handle them accordingly.` Purely a wording tweak; the `0600`-archive / `0700`-tree perms already protect both.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed the full diff at `ee46564` after the `synchronize` push (which adds `armada.json` to the backup set). This PR has already absorbed four prior review rounds, and the structure is clean and well-tested. No blocking issues.

**Verified the feature's core invariant.** The whole point of the backup loop is to never capture a torn file while reading sources without `state.mu`. I confirmed every production writer of all seven backed-up files now routes through `atomicfile.Write` (temp + rename): `state.json` (`saveLocked`), `config.json` (`SaveConfig`), `armada.json` (`SaveArmada`), `gateway_session.json` (`saveSession`), and `mcp.port`/`mcp.token`/`mcp.env` (`mcp.go`). Grepping the writers of these paths shows no remaining plain-`os.WriteFile` site for any of them. `atomicfile.Write` is correct (same-dir temp, `Chmod` before rename, removes temp on early return), and rename is atomic cross-process on POSIX — so the client-written `armada.json` is safe against the daemon's unlocked read too.

**Other checks that pass:**
- `envDurationDefault` rejects non-positive values (`d > 0`), so a misconfigured `FLEET_BACKUP_INTERVAL=0s`/negative can't feed `time.NewTicker` a panic-inducing duration — it falls back to the default.
- The `n == 0` cleanup ordering (remove temp → day dir → backup root, each a no-op when non-empty) correctly honors "writes nothing rather than an empty archive" without nuking sibling archives.
- `parseArchiveHour` is the single source of truth shared by prune + list, so they can't drift; reverse-lexical sort on the zero-padded `<date>/<hour>` path gives a correct newest-first order.
- `hdr.Size` derives from the bytes read (round-1 fix) so a concurrent size change can't fail the snapshot.
- `armada.json` inclusion is covered by tests (`TestWriteBackupRoundTrip` asserts `n == 7`; `TestMcpRestoreBackupDocuments` asserts 7 documented files). Unit CI is green.

**One minor, non-blocking nit** (left inline on `backup.go`): the `fleet_restore_backup` top-level `Warning` string still names only `mcp.token` as the embedded secret, but the archive now also carries `armada.json`'s bearer tokens. Worth widening the wording so the most-surfaced line reflects both secrets. The perms (`0600` archive / `0700` tree) already cover it, so this is documentation only.

Nice work — the atomic-write refactor is a genuinely good outcome that hardens more than just the backup path.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — re-ran after the latest commit `ee46564` ("include armada.json in the backup set"). All acceptance criteria met; no defects found.

### Environment
Isolated `golang:1.25-bookworm` Docker container (`xz-utils`/`tar`/`tmux` installed), workspace mounted, `fleet` built from `./cmd/fleet`. Ran the real daemon (`fleet server`) under a sandboxed `$HOME`, driving the loop fast via `FLEET_BACKUP_INTERVAL=2s` (default 30-day retention). Container torn down afterward.

### What I tested (behavior-level, driving the running daemon)

| # | Scenario | Result |
|---|----------|--------|
| 1 | Backup loop runs unconditionally; first snapshot fires **immediately** at startup, then every interval | ✅ logged `backup: wrote snapshot … files=6` |
| 2 | Archive lands at `~/.fleet/backup/<date>/<hour>.tar.xz` (`2026-06-23/03.tar.xz`) | ✅ |
| 3 | Archive integrity — passes system `xz -t` and unpacks with `tar -xJf` | ✅ |
| 4 | Flat, base-name layout (no dir tree) | ✅ `config.json, state.json, armada.json, mcp.env, mcp.port, mcp.token` |
| 5 | Content round-trips byte-for-byte vs originals (all files `MATCH`) | ✅ |
| 6 | **`armada.json` included** (this PR's new commit) and round-trips with its secret token | ✅ |
| 7 | Missing source skipped — `gateway_session.json` (omitted) not in archive, no failure (`files=6`, not 7) | ✅ ABSENT |
| 8 | Idempotent same-hour overwrite — ~16 ticks all rewrote the same `03.tar.xz`; exactly one archive for the hour | ✅ |
| 9 | Atomic write (temp + rename) — only final `.tar.xz` ever visible, no partials | ✅ |
| 10 | Permissions on a **pristine** run: backup root `0700`, day dir `0700`, archive `0600` (protects embedded `mcp.token`) | ✅ |
| 11 | Retention prune (default 30d): `2026-05-01/14.tar.xz` removed (`removed=1`); `2026-06-22/10.tar.xz` kept | ✅ |
| 12 | Foreign/non-archive entries left untouched: `notes.txt`, `not-a-date/` survived; emptied day dir handling correct | ✅ |
| 13 | `FLEET_BACKUP_INTERVAL` env seam honored | ✅ |
| 14 | `fleet_restore_backup` MCP tool (auth'd HTTP call): documentation-only, lists real archives newest-first + `latest`, correct `backup_dir`/`restore_into`/`path_layout`/`compression`/`contents`/`procedure`/`warning`, **excludes pruned & foreign entries**, mutates nothing | ✅ |
| 15 | Documented restore procedure actually works — corrupted `state.json` recovered by unpacking `latest` over `~/.fleet`; `mcp.token` restored `0600` | ✅ |
| 16 | Sanity: full `go build ./...` clean; daemon starts normally (gRPC socket, MCP server, `mcp.*` discovery files) — no regression from the atomic-write refactor | ✅ |

### Reproduction highlights
- Snapshot+prune in one run (seeded an old `2026-05-01/14.tar.xz`, a recent `2026-06-22/10.tar.xz`, a foreign `notes.txt`, a `not-a-date/` dir):
  - `~/.fleet/backup/2026-06-23/03.tar.xz` created; `2026-05-01/14.tar.xz` pruned (`removed=1`); recent + foreign entries survived.
- Round-trip: `tar -xJf …/03.tar.xz` → all 6 present files `diff`-identical to originals; `xz -t` OK; archive `0600`, day dir `0700`.
- Restore: stop daemon → `printf 'CORRUPTED…' > state.json` → `tar -xJf <latest> -C ~/.fleet` → `state.json` back to good content; `armada.json` secret intact.

### Notes (non-blocking)
- The "pristine `~/.fleet` writes nothing / empty archive" edge isn't reachable via the live daemon (it always creates `mcp.port/token/env`, so there's always ≥1 source). It's covered by the unit tests; I verified the inverse (missing individual sources are skipped) against the running daemon.
- Backup-root `0700` holds because the daemon's `MkdirAll(...,0700)` creates it; a pre-existing looser `backup/` wouldn't be re-tightened (an earlier 755 I saw was just my own fixture seeding the dir). Not a defect.

No defects, regressions, or build/run failures. Marking **qa-passed**.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary** — reviewed at `5f2285b`. I checked the branch out and verified locally: `go build ./...`, `go vet`, and the new suites all pass, including under `-race` (`atomicfile`, `state`, and the `server` backup/restore tests). This is a clean, unusually well-documented change. I found **no correctness bugs** — only the low-priority notes below.

**What's solid**
- The lockless-read premise actually holds: every one of the 6 backup sources (`state.json`, `config.json`, `gateway_session.json`, `mcp.env`/`.port`/`.token`) now writes through `atomicfile.Write` (temp+rename). I grepped the tree for stray `os.WriteFile` on those paths — none remain, so the backup loop can't observe a torn file. 👍
- The loop mirrors the existing `runScheduler` idiom (tick-at-top, `ctx.Done`/ticker `select`), runs on `hubCtx`, and is correctly placed **after** `startMCPServer` so its immediate first tick already sees the freshly-written `mcp.*` files instead of racing them.
- Secret-aware perms (`0700` tree / `0600` archives), idempotent hour-keying, atomic archive writes, orphan-temp sweeping, skip-missing sources, and the pristine-`~/.fleet` no-op are all handled and covered by tests.

**Observations (all low priority)**

1. **`armada.json` exclusion — confirm the scoping.** It's now deliberately excluded with a clear rationale ("client registry, not daemon state"). Worth a second look, though: `armada.json` is durable, **secret-bearing** state that lives in `~/.fleet`, and losing it (corruption / an errant `destroy` — the exact scenarios this PR targets) means losing every remote-fleet gateway URL + bearer token with no recovery path. Since the daemon already reads/writes it and you just made it an `atomicfile.Write` writer, including it would be nearly free. Totally fine to keep it out if that's the intent — just flagging the asymmetry for a conscious decision.

2. **Local-time hour keying can silently drop an hour (DST / backward clock step).** `writeBackup` buckets on the local wall-clock hour; on a DST fall-back the 01:00–01:59 hour recurs and the second snapshot overwrites the first `01.tar.xz`. Details inline. Low impact, easy to harden by bucketing on UTC.

3. **Retention is effectively hour-rounded.** Age is derived from the hour-boundary start rather than write time — intentional and fine for the 30-day default. Just note the `FLEET_BACKUP_RETENTION` test seam can prune the *just-written* archive if set below ~1h (this tick's archive is already dated at the hour start, i.e. before `now-retention`). Harmless for production values; only a caveat for the seam.

Nice work overall — the comments and test coverage make the intent easy to follow.

internal/server/backup.go
```diff
@@ -0,0 +1,441 @@
...
+		return "", 0, nil
+	}
+
+	dir := filepath.Join(backupBaseDir(), now.Format("2006-01-02"))
+	// 0700: the archive embeds mcp.token (a secret), so the whole backup tree is
+	// owner-only, matching how ~/.fleet itself is created.
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return "", 0, err
+	}
+	finalPath := filepath.Join(dir, now.Format("15")+".tar.xz")
```
BenjaminBenetti: 🔍 **CodeReview Agent**

DST / backward-clock note: the archive is keyed on the **local** wall-clock hour (`now.Format`). During a DST fall-back the 01:00–01:59 hour repeats, so the second snapshot rewrites the first `01.tar.xz` and that hour of backup history is silently lost — a backward NTP step does the same. `pruneBackups` already resolves day dirs with `now.Location()`, so the only remaining gap is this wall-clock ambiguity. Bucketing the path on UTC (`now.UTC().Format(...)` for both the `<date>` and `<hour>` segments, and matching it in `pruneBackups`) would make the hour monotonic and unambiguous. Low priority.

internal/server/backup.go
```diff
@@ -0,0 +1,441 @@
...
+	days, err := os.ReadDir(base)
+	if err != nil {
+		return nil
+	}
+	var out []restoreBackupArchive
+	for _, day := range days {
+		if !day.IsDir() {
+			continue
+		}
+		if _, err := time.Parse("2006-01-02", day.Name()); err != nil {
```
BenjaminBenetti, line 377: 🔍 **CodeReview Agent**

Minor consistency nit: this validates day dirs with `time.Parse` (UTC), while `pruneBackups` uses `time.ParseInLocation(..., now.Location())`. Both only check the *format* here (the parsed value is discarded in the listing), so there's no behavioral bug — but a shared helper (e.g. an `isBackupDay(name)` sibling to `parseArchiveHour`) would keep the two from drifting on what counts as a backup day directory.

BenjaminBenetti:
Thanks — QA PASS noted, and addressed the round-5 CodeReview notes in a4de69b:

- **#2 DST hour-drop — fixed.** `writeBackup` now buckets the `<date>/<hour>` path on `now.UTC()` (and `pruneBackups` parses day names as UTC to match), so a DST fall-back can't make two instants collide on one local-hour file and silently drop a backup. Documented in the file header + `fleet_restore_backup`'s `path_layout`; new test asserts a non-UTC `now` still lands at the UTC date/hour. Existing tests unaffected (they already use UTC `now`).
- **#1 armada.json — already done** in ee46564 (you asked for it to be included; the review was at the earlier 5f2285b).
- **#3 hour-rounded retention — intentional**, no change. It's only observable via the sub-1h `FLEET_BACKUP_RETENTION` test seam (which can prune the just-written archive); fine at the 30-day default.

`go build`/`go vet` and the full server suite are green locally.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Round 6 — `synchronize` on `a4de69b` (UTC-hour bucketing).**

Verified the latest change and the state of the whole PR.

✅ **The round-5 fix is correct and well-covered.** `writeBackup` now buckets on `now.UTC()` for both `<date>` and `<hour>`, and the two readers were brought into line: `pruneBackups` parses day dirs with `time.Parse` (UTC) and reconstructs the instant as `dayStart + hour` (UTC), and `listBackupArchives` already parsed UTC. The retention comparison (`when.Before(cutoff)`) compares absolute instants, so the cross-zone math holds. `TestWriteBackupBucketsUTC` pins the UTC+5 → previous-UTC-day case. Good call moving off local wall-clock.

✅ **All eight inline notes from prior rounds are addressed** at this head: self-consistent `hdr.Size` taken from the bytes read; atomic writes for every backed-up source (incl. the newly-added `armada.json`) via the extracted `atomicfile.Write`; the `n==0` cleanup drops the temp file *before* the dir removals; the backup loop is launched after `startMCPServer`; and prune/list now agree on UTC.

✅ **Local validation:** `go build ./...`, `go vet`, and the `atomicfile` / `state` / `server` backup suites all pass on `a4de69b`.

A few small, **non-blocking** observations — all optional:

1. **Orphan-temp sweep rests on the single-daemon invariant.** `pruneBackups` deletes any `.backup-*.tmp` in a day dir on the premise that snapshots are serial and the current tick's `writeBackup` has already returned. True for one fleetd, but if a second daemon ever wrote into the same `~/.fleet/backup`, the sweep could remove a *live* temp mid-write and fail that snapshot's rename. The code comment states the assumption — just worth keeping in mind if snapshotting ever becomes concurrent.

2. **Immediate first tick runs before the `ctx.Done()` check.** `runBackupLoop` calls `backupTick(time.Now())` and only then enters the `select`. If `hubCtx` is already cancelled at launch (a very fast shutdown), the loop still performs one snapshot. Harmless — arguably a feature (captures the latest state on the way down) — but a `select { case <-ctx.Done(): return; default: }` before the first tick would make cancellation strictly authoritative if you'd prefer.

3. **Hourly Info-level log lines.** `backupTick` logs `backup: wrote snapshot` (and prune results) at Info every hour. Low volume, but it does accumulate in the daemon log over a long-lived run; consider Debug for the success path if it ever gets noisy.

Nothing here blocks merge — the feature is well-structured (clean `atomicfile` extraction, exact path-derived retention, idempotent hour keying, pure-Go xz so no host `xz` dependency) and the test coverage is thorough. Nice work.

BenjaminBenetti:
Thanks for the round-6 pass. The three optional notes — I've considered each and am intentionally leaving them as-is:

1. **Orphan-temp sweep / single-daemon invariant** — correct, and it's load-bearing across the codebase: exactly one fleetd per user is guaranteed by the lifetime lock (`acquireServerLock` in `server.go`). The code comment already states the assumption. If snapshotting ever goes concurrent this would need revisiting, but that's a much larger change than this PR.
2. **Immediate first tick before `ctx.Done()`** — deliberately kept, for two reasons: it mirrors `runScheduler`'s tick-at-top shape (consistency), and a snapshot on the way down is desirable (captures the latest state). The write is idempotent (hour-keyed), so a cancelled-at-launch daemon just rewrites the current hour harmlessly.
3. **Hourly Info logs** — ~24 lines/day is low volume and matches how the automation scheduler logs its events at Info; the success line is useful for confirming backups are actually happening. `flog` has no Debug level anyway, so I'd rather keep it visible.

Happy to revisit any of these if you'd prefer a different call. CI green, merge `CLEAN`.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — `qa-passed`

Manual, behavior-level QA of the hourly fleetd backup loop. All testing was done inside a throwaway Docker container (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`, the repo's devcontainer image) with the PR branch (`a4de69b`) mounted — the daemon was driven via tmux *inside* the container, no automated tests were run.

### Setup
- Built `fleet` from the branch (`go build ./cmd/fleet`, needed `-buildvcs=false` only due to the mounted `.git` having a different owner inside the container — not a code issue).
- Seeded a fresh `~/.fleet` with marker `config.json` / `state.json` / `armada.json` / `gateway_session.json`, then ran `fleet server` with `FLEET_BACKUP_INTERVAL=3s` so the loop could be observed in seconds.

### What I tested & observed

| # | Scenario | Result |
|---|----------|--------|
| 1 | Loop writes an archive | ✅ `~/.fleet/backup/2026-06-23/03.tar.xz` appears on the immediate first tick and every interval; log: `backup: wrote snapshot … files=7` |
| 2 | All declared sources captured, **flat by base name** | ✅ archive contains `config.json, gateway_session.json, mcp.env, mcp.port, mcp.token, state.json, armada.json` — no directory tree |
| 3 | Path layout = `<UTC-date>/<UTC-hour>.tar.xz` | ✅ bucketed on UTC clock hour |
| 4 | Permissions | ✅ archive file `0600`; `backup/` and day dirs `0700` |
| 5 | Interoperable compression | ✅ system `xz -t` passes integrity; `tar -xJf` (standard tools) unpacks cleanly; marker contents round-trip byte-for-byte |
| 6 | Idempotent same-hour overwrite | ✅ exactly **one** archive per hour; after mutating `state.json`, the next tick refreshed the same `03.tar.xz` to the new content |
| 7 | Retention prune (>30d), age from path | ✅ a 50-day-old archive pruned; `removed=2` logged |
| 8 | **Per-hour** boundary granularity | ✅ on the cutoff day `2026-05-24`, hour `02` (before cutoff) pruned while hour `04` (after cutoff) kept |
| 9 | Foreign entries untouched | ✅ `notes.txt` in a pruned day survived and kept its day dir alive |
| 10 | Emptied day-dir cleanup | ✅ a day holding only expired archives had its now-empty dir removed |
| 11 | Graceful shutdown stops the loop | ✅ `SIGTERM` → log `fleet server stopping (signal)`, process exits, snapshot count freezes (no further ticks) |
| 12 | Sanity: serving coexists with the loop | ✅ a gRPC client call was served while the loop ran (the only error returned was my seeded `state.json` being an array vs the real map schema — a test-fixture artifact, not a backup defect) |

`FLEET_BACKUP_INTERVAL` and `FLEET_BACKUP_RETENTION` overrides were both honored (used to drive the test).

### Reproduce (key step)
```
docker run -dit --name qa -v "$PWD":/work -w /work mcr.microsoft.com/devcontainers/go:2-1.25-bookworm bash
docker exec qa bash -lc 'cd /work && go build -buildvcs=false -o /tmp/fleet ./cmd/fleet'
docker exec qa bash -lc 'HOME=/root FLEET_BACKUP_INTERVAL=3s /tmp/fleet server &'
docker exec qa bash -lc 'sleep 5; tar -tJvf /root/.fleet/backup/*/*.tar.xz; xz -t /root/.fleet/backup/*/*.tar.xz'
```

### Note (not a defect)
The "pristine `~/.fleet` writes nothing" contract can't be exercised through the live daemon, because `fleet server` writes `mcp.port`/`mcp.token`/`mcp.env` at startup *before* the first backup tick, so there is always ≥1 source present. It's covered by the package unit tests and is consistent with the launch-after-MCP ordering in `server.go`.

Everything behaved as the ticket describes. Marking **qa-passed**.

PR #206, opened 2026-06-23
feat(automation): record per-trigger event logs, viewable via TUI/CLI/MCP

BenjaminBenetti:
## What

Makes the automation trigger system debuggable by keeping a durable, host-side log of every trigger firing's **event payload** — and exposing it through the TUI, CLI, and MCP.

Each time a trigger fires (schedule or webhook — both funnel through `fireTriggerAgents`), the daemon writes the firing's payload to:

```
~/.fleet/logs/<fleet>/trigger/<trigger>/event-<UTC timestamp>.log
```

keeping the most recent **100** firings per trigger (oldest pruned as new ones arrive). This is the *same* payload that's already copied into each spawned agent's instance at `/tmp/fleet-event-<ts>` — now captured durably on the host so a firing can be inspected after its instance has been reaped.

- **webhook** firing → log holds the raw request body
- **schedule** firing → log holds the fire-time

Each log carries a short header (`# trigger:` / `# type:` / `# webhook:` / `# fired:`) followed by the payload.

## Reading it back

All three read paths go through a new unary `TriggerLogs` RPC, so they work unchanged when the daemon is **remote**:

- **TUI** — `L` on a selected trigger row pages the concatenated logs in `less` (scroll/search). Help hint + the `?` keybindings overlay updated.
- **CLI** — `fleet trigger logs <fleet> <name>` prints the concatenated logs (pipe to a pager yourself).
- **MCP** — `fleet_trigger_logs {fleet, trigger}` returns `{logs, count}` so the admiral can debug what fired (or failed to fire) an automation.

## Safety

Writer and reader share `state.TriggerLogsDir`, which sanitizes fleet/trigger names to safe single path segments, so a crafted name can't escape the logs tree. Logging is strictly best-effort — a failure is warned and never blocks a firing.

## Tests

- `internal/server/trigger_logs_test.go` — write/read round-trip (webhook + schedule), 100-firing rotation, same-second no-overwrite, empty history, path-escape guard, and the RPC handler end to end.
- `internal/cli/trigger_test.go` — `trigger logs` command (content + empty history).
- `internal/server/mcp_automation_test.go` — `fleet_trigger_logs` tool (empty, populated, missing-arg error); `mcp_test.go` tool-count bumped.
- `internal/tui/trigger_logs_test.go` — `L` on a trigger row (empty/error/launch) and ignored off-trigger.

`go build ./...`, `go test ./...`, and `make lint` (depguard boundary) all pass; proto regenerated via `make proto`.

### DIFF
--- a/internal/server/trigger_logs.go
+++ b/internal/server/trigger_logs.go
@@ -46,12 +46,15 @@ const triggerEventLogKeep = 100
 // not part of the automation's critical path.
 var logTriggerEvent = func(fleetName string, ev *triggerEvent) {
 	dir := state.TriggerLogsDir(fleetName, ev.triggerName)
-	if err := os.MkdirAll(dir, 0o755); err != nil {
+	// 0700 dir / 0600 file: a webhook body can carry signatures or secrets, and
+	// these logs persist durably on the host (unlike the payload's previous home
+	// inside an ephemeral instance), so keep them readable only by the owner.
+	if err := os.MkdirAll(dir, 0o700); err != nil {
 		flog.Warn("automation: trigger log mkdir failed", "fleet", fleetName, "trigger", ev.triggerName, "err", err)
 		return
 	}
 	path := uniqueTriggerLogPath(dir, ev.firedAt)
-	if err := os.WriteFile(path, triggerEventLogContent(ev), 0o644); err != nil {
+	if err := os.WriteFile(path, triggerEventLogContent(ev), 0o600); err != nil {
 		flog.Warn("automation: trigger log write failed", "fleet", fleetName, "trigger", ev.triggerName, "err", err)
 		return
 	}
@@ -78,17 +81,25 @@ func triggerEventLogContent(ev *triggerEvent) []byte {
 	return []byte(b.String())
 }
 
-// uniqueTriggerLogPath returns event-<UTC timestamp>.log in dir, suffixing -N
-// when a log with that second-granular name already exists (two webhook events
-// can fire within the same second), so no firing overwrites another.
+// uniqueTriggerLogPath returns event-<UTC timestamp>.log in dir, suffixing a
+// counter when a log with that second-granular name already exists (two webhook
+// events can fire within the same second), so no firing overwrites another.
+//
+// The counter is zero-padded and uses '_' as the separator on purpose: '_'
+// (0x5F) sorts AFTER '.' (0x2E), and zero-padding makes the counter sort
+// numerically, so lexical filename order stays chronological within a second
+// (event-<ts>.log < event-<ts>_0002.log < event-<ts>_0003.log < ...). That
+// matters because eventLogNames / pruneTriggerLogs / readTriggerLogs all order
+// by name — a naive "-N" suffix would sort the 2nd firing before the 1st and
+// "-10" before "-2", reversing the pager and mis-pruning at the 100-file edge.
 func uniqueTriggerLogPath(dir string, firedAt time.Time) string {
 	base := "event-" + firedAt.UTC().Format("20060102T150405Z")
 	path := filepath.Join(dir, base+".log")
-	for i := 2; i < 100000; i++ {
+	for i := 2; i < 10000; i++ {
 		if _, err := os.Stat(path); errors.Is(err, fs.ErrNotExist) {
 			return path
 		}
-		path = filepath.Join(dir, fmt.Sprintf("%s-%d.log", base, i))
+		path = filepath.Join(dir, fmt.Sprintf("%s_%04d.log", base, i))
 	}
 	return path
 }
--- a/internal/server/trigger_logs.go
+++ b/internal/server/trigger_logs.go
@@ -1,0 +1,162 @@
+package server
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/BenjaminBenetti/fleet-man/fleetgrpc"
+	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
+	"github.com/BenjaminBenetti/fleet-man/internal/flog"
+	"github.com/BenjaminBenetti/fleet-man/internal/state"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// trigger_logs.go records and serves a per-trigger history of automation
+// firings, for debuggability. Each time a trigger fires (the schedule and
+// webhook paths both funnel through fireTriggerAgents), the daemon writes the
+// firing's event payload — the webhook request body, or the schedule fire-time — to
+//
+//	~/.fleet/logs/<fleet>/trigger/<trigger>/event-<UTC timestamp>.log
+//
+// keeping the most recent triggerEventLogKeep files. This is the SAME payload
+// that gets copied into each spawned agent's instance (see launchAutomation-
+// Command / writeAutomationEventFile), captured durably on the host so a firing
+// can be inspected after the instance is reaped — which is the whole point: it
+// makes the trigger system debuggable. The TUI's 'L' pager, the `fleet trigger
+// logs` CLI, and the fleet_trigger_logs MCP tool all read it back through the
+// TriggerLogs RPC, so it works the same whether the daemon is local or remote.
+
+// triggerEventLogKeep bounds how many event logs are retained per trigger. Older
+// ones are pruned as new firings arrive, so a busy trigger can't grow the logs
+// tree without bound.
+const triggerEventLogKeep = 100
+
+// logTriggerEvent records one trigger firing to the trigger's on-host event log
+// directory, then prunes to the newest triggerEventLogKeep files. It is a
+// package-var seam (overridden in tests) and strictly best-effort: a logging
+// failure is warned and never blocks the firing — the log is a debugging aid,
+// not part of the automation's critical path.
+var logTriggerEvent = func(fleetName string, ev *triggerEvent) {
+	dir := state.TriggerLogsDir(fleetName, ev.triggerName)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		flog.Warn("automation: trigger log mkdir failed", "fleet", fleetName, "trigger", ev.triggerName, "err", err)
+		return
+	}
+	path := uniqueTriggerLogPath(dir, ev.firedAt)
+	if err := os.WriteFile(path, triggerEventLogContent(ev), 0o644); err != nil {
+		flog.Warn("automation: trigger log write failed", "fleet", fleetName, "trigger", ev.triggerName, "err", err)
+		return
+	}
+	pruneTriggerLogs(dir, triggerEventLogKeep)
+}
+
+// triggerEventLogContent renders one event log: a short header naming the firing
+// followed by its payload, so a concatenated dump is self-describing. The
+// payload mirrors exactly what the spawned agent received (triggerEvent.payload).
+func triggerEventLogContent(ev *triggerEvent) []byte {
+	var b strings.Builder
+	fmt.Fprintf(&b, "# trigger: %s\n", ev.triggerName)
+	fmt.Fprintf(&b, "# type:    %s\n", ev.kind)
+	if ev.kind == fleet.TriggerWebhook && ev.webhookName != "" {
+		fmt.Fprintf(&b, "# webhook: %s\n", ev.webhookName)
+	}
+	fmt.Fprintf(&b, "# fired:   %s\n", ev.firedAt.UTC().Format(time.RFC3339))
+	b.WriteByte('\n')
+	payload := ev.payload()
+	b.Write(payload)
+	if len(payload) == 0 || payload[len(payload)-1] != '\n' {
+		b.WriteByte('\n')
+	}
+	return []byte(b.String())
+}
+
+// uniqueTriggerLogPath returns event-<UTC timestamp>.log in dir, suffixing -N
+// when a log with that second-granular name already exists (two webhook events
+// can fire within the same second), so no firing overwrites another.
+func uniqueTriggerLogPath(dir string, firedAt time.Time) string {
+	base := "event-" + firedAt.UTC().Format("20060102T150405Z")
+	path := filepath.Join(dir, base+".log")
+	for i := 2; i < 100000; i++ {
+		if _, err := os.Stat(path); errors.Is(err, fs.ErrNotExist) {
+			return path
+		}
+		path = filepath.Join(dir, fmt.Sprintf("%s-%d.log", base, i))
+	}
+	return path
+}
+
+// pruneTriggerLogs keeps only the newest keep event logs in dir, deleting the
+// rest. Event filenames embed a sortable UTC timestamp, so lexical order is
+// (near enough) chronological order.
+func pruneTriggerLogs(dir string, keep int) {
+	logs := eventLogNames(dir)
+	if len(logs) <= keep {
+		return
+	}
+	for _, name := range logs[:len(logs)-keep] {
+		_ = os.Remove(filepath.Join(dir, name))
+	}
+}
+
+// eventLogNames returns the event-*.log filenames in dir, sorted ascending
+// (oldest first). A missing or unreadable directory yields nil.
+func eventLogNames(dir string) []string {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil
+	}
+	var names []string
+	for _, e := range entries {
+		if !e.IsDir() && strings.HasPrefix(e.Name(), "event-") && strings.HasSuffix(e.Name(), ".log") {
+			names = append(names, e.Name())
+		}
+	}
+	sort.Strings(names)
+	return names
+}
+
+// readTriggerLogs concatenates a trigger's event logs (oldest first) for
+// viewing, each preceded by a separator naming its file. It returns the text and
+// the number of logs included. A trigger that has never fired (no directory)
+// yields "" and 0 — not an error — so callers can report "no events" cleanly.
+func readTriggerLogs(fleetName, triggerName string) (string, int, error) {
+	dir := state.TriggerLogsDir(fleetName, triggerName)
+	names := eventLogNames(dir)
+	var b strings.Builder
+	included := 0
+	for _, name := range names {
+		data, err := os.ReadFile(filepath.Join(dir, name))
+		if err != nil {
+			continue // a log removed mid-read just drops out of the dump
+		}
+		fmt.Fprintf(&b, "===== %s =====\n", name)
+		b.Write(data)
+		if len(data) == 0 || data[len(data)-1] != '\n' {
+			b.WriteByte('\n')
+		}
+		b.WriteByte('\n')
+		included++
+	}
+	return b.String(), included, nil
+}
+
+// TriggerLogs serves an automation trigger's recorded event logs. It reads the
+// daemon-owned log files, so it works identically for a local or remote daemon.
+func (s *service) TriggerLogs(_ context.Context, req *fleetgrpc.TriggerLogsRequest) (*fleetgrpc.TriggerLogsReply, error) {
+	if req.GetFleet() == "" || req.GetTrigger() == "" {
+		return nil, status.Error(codes.InvalidArgument, "fleet and trigger are required")
+	}
+	logs, count, err := readTriggerLogs(req.GetFleet(), req.GetTrigger())
+	if err != nil {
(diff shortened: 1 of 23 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the full diff for #206. This is a clean, well-tested feature: `safeLogSegment` path sanitization is solid (and covered by `TestTriggerLogsDirNoEscape`), logging is correctly best-effort (a failure is warned and never blocks a firing), and folding the per-agent `triggerEvent` into one shared, log-once value in `fireTriggerAgents` is a nice simplification. The write path only ever runs on the scheduler goroutine, so the stat-then-write in `uniqueTriggerLogPath` is race-free. A few things worth a look — none blocking:

**1. Same-second collision suffix breaks chronological ordering (correctness, low–moderate).**
`uniqueTriggerLogPath` names collisions `event-<ts>-N.log`. But `-` (0x2D) sorts before `.` (0x2E), and the numeric suffix sorts lexically — so the ordering that `eventLogNames` / `pruneTriggerLogs` / `readTriggerLogs` all rely on is wrong within a same-second burst:
- `event-<ts>-2.log` (2nd firing) sorts **before** `event-<ts>.log` (1st firing) → intra-second order is reversed.
- `event-<ts>-10.log` sorts before `event-<ts>-2.log` → numeric suffixes don't order numerically.

Cross-second ordering is fine, so this is confined to firings within the same second — realistic for a bursty webhook trigger. Effects: out-of-order display in the pager, and `pruneTriggerLogs` can drop a not-actually-oldest file when same-second logs straddle the 100-file boundary. Zero-padding the suffix (and suffixing the base too, e.g. always `-0001`), or adding sub-second precision / a monotonic counter, would make lexical order match firing order. Inline note added.

**2. Trigger delete/rename orphans the log dir (housekeeping, low).**
Neither `mcpTriggerDelete` nor the TUI `deleteTrigger` removes `~/.fleet/logs/<fleet>/trigger/<trigger>/`, and a rename (`new_name`) starts a fresh dir while orphaning the old one — so logs persist forever for triggers that no longer exist. Probably fine as a follow-up, but a `RemoveAll` on delete would close the loop.

**3. Doc nits (low).**
- The separator documented in `exec.proto` / `TriggerLogsReply.logs` (and the MCP `TriggerLogsOutput.Logs` doc) says `"=== event-<timestamp>.log ==="` (3 `=`), but `readTriggerLogs` actually emits `"===== %s ====="` (5 `=`).
- `state.TriggerLogsDir`'s doc comment says "keeping the most recent **few**" — it's actually `triggerEventLogKeep` = 100.

**Informational (not a finding):** webhook bodies — which can carry signatures/secrets — are now persisted durably on the host at `0644` inside a `0755` tree, whereas before the payload only lived inside the ephemeral instance. This matches the existing `~/.fleet` convention (`state.json`, warn files are all `0644`), so it's not a deviation; just flagging that the exposure window moved from ephemeral to durable. Tighter perms (`0600`/`0700`) would be reasonable defense-in-depth if multi-user hosts are ever in scope.

Test coverage across server / CLI / MCP / TUI is thorough. Nice work. 👍

internal/server/trigger_logs.go
```diff
@@ -0,0 +1,162 @@
...
+// when a log with that second-granular name already exists (two webhook events
+// can fire within the same second), so no firing overwrites another.
+func uniqueTriggerLogPath(dir string, firedAt time.Time) string {
+	base := "event-" + firedAt.UTC().Format("20060102T150405Z")
+	path := filepath.Join(dir, base+".log")
+	for i := 2; i < 100000; i++ {
+		if _, err := os.Stat(path); errors.Is(err, fs.ErrNotExist) {
+			return path
+		}
+		path = filepath.Join(dir, fmt.Sprintf("%s-%d.log", base, i))
```
BenjaminBenetti: 🔍 **CodeReview Agent**

The `-N` collision suffix breaks the lexical-order assumption that `eventLogNames` / `pruneTriggerLogs` / `readTriggerLogs` depend on. `-` (0x2D) sorts before `.` (0x2E), so `event-<ts>-2.log` sorts **before** `event-<ts>.log` — the 2nd same-second firing ends up ahead of the 1st. Numeric suffixes also sort lexically (`-10` before `-2`). Cross-second ordering is unaffected, so the impact is limited to same-second bursts (plausible for a busy webhook trigger): reversed order in the pager, and `pruneTriggerLogs` possibly dropping a not-actually-oldest file at the 100-file boundary. Zero-padding the suffix — and suffixing the base too (e.g. `event-<ts>-0001.log`) — or adding sub-second precision would keep lexical order == chronological order.

BenjaminBenetti: Fixed in 48f57a3. Switched the collision suffix from `-N` to a zero-padded `_NNNN`: `_` (0x5F) sorts *after* `.` (0x2E), so the base `event-<ts>.log` (1st firing) stays first, and zero-padding makes the counter order numerically — so lexical filename order now equals firing order within a same-second burst (cross-second ordering was already correct since the differing digit precedes `Z`). `TestUniqueTriggerLogPathNoOverwrite` now fires three same-second webhook events with distinct bodies and asserts they read back in firing order.

BenjaminBenetti:
Thanks for the thorough review! Addressed in 48f57a3:

**1. Same-second ordering (fixed).** Collision suffix is now zero-padded `_NNNN` instead of `-N`. `_` (0x5F) sorts after `.` (0x2E) so the base `event-<ts>.log` stays first, and zero-padding makes the counter sort numerically — lexical filename order == firing order within a second, which is what `eventLogNames`/`pruneTriggerLogs`/`readTriggerLogs` rely on. The ordering test now proves it (three same-second firings read back in firing order).

**3. Doc nits (fixed).** Separator in `exec.proto` corrected to `=====` (5, matching the `"===== %s ====="` the code emits), and `state.TriggerLogsDir`'s comment now says "the most recent 100".

**4. Perms (hardened).** Good call — since webhook bodies can carry signatures/secrets and these now persist durably on the host (vs. the previously-ephemeral instance file), trigger log dirs are now `0700` and files `0600` (was `0755`/`0644`). New `TestTriggerLogPerms` covers it.

**2. Orphaned log dir on delete/rename (deferring, as you suggested).** Leaving this as a follow-up: the only server-side chokepoint is `SetFleetSettings` (deletes/renames arrive there as a wholesale settings replace), so doing it right means diffing old-vs-new trigger names there — and a rename is indistinguishable from delete+add at that level, so "carry logs forward on rename" vs "drop on delete" needs a deliberate design rather than a `RemoveAll` bolted onto the write path. Orphaned dirs are bounded (≤100 small files per vanished trigger name) and harmless in the meantime. Happy to open a tracking issue.

CI is green (unit + all integration shards); `go build`, `go test ./...`, `make lint`, and `make proto` all clean.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — every user-facing flow in this PR behaves as described in the ticket.

## Environment
- Isolated throwaway Docker container (`golang:1.25`), workspace mounted, PR branch `feat/trigger-event-logs` @ `9ca387c`.
- Built `fleet` inside the container (`go build ./cmd/fleet`, clean).
- All app execution + interactive TUI driven **inside** the container (tmux + `docker exec`).
- No docker-in-docker: instance provisioning intentionally fails (`failed` status) — which is the realistic worst case, and confirms the event log is written **before** provisioning, so a firing is captured regardless.

## What I tested (drove it like a user)

### 1. Write path — a firing records a durable host-side log ✅
Seeded a fleet `qafleet` with agent `writer` + a schedule trigger `fire` (`cron: * * * * *`) and a never-firing trigger `never` (`0 0 1 1 *`), started the daemon (`fleet ls`, `FLEET_AUTOMATION_TICK_INTERVAL=2s`). On each tick the trigger fired and wrote one file per firing under `~/.fleet/logs/qafleet/trigger/fire/event-<UTC>.log`, e.g.:
```
# trigger: fire
# type:    schedule
# fired:   2026-06-23T02:09:46Z

2026-06-23T02:09:46Z
```
- Header (`# trigger:` / `# type:` / `# fired:`) + schedule fire-time payload exactly as specified.
- One file **per firing** (`event-...T020946Z.log`, `...T021000Z.log`, ...); a new file every minute.
- The never-firing trigger created **no** directory/log. ✔

### 2. CLI — `fleet trigger logs <fleet> <name>` ✅
- **Populated:** prints all firings concatenated oldest-first, each under a `===== <file> =====` separator.
- **Empty history:** `No events recorded for trigger "never" in fleet "qafleet"` (exit 0).
- Unknown trigger / unknown fleet: same clean "No events recorded" message (no crash).

### 3. MCP — `fleet_trigger_logs` tool ✅
Drove the daemon's Streamable-HTTP MCP endpoint (token from `~/.fleet/mcp.token`).
- Tool is advertised in `tools/list`.
- `{fleet:qafleet, trigger:fire}` → `{"count": 2, "logs": "...===== event-...log ====="}` (matches CLI output).
- `{..., trigger:never}` → `{"count": 0, "logs": ""}`.
- Missing `trigger` arg → tool error (`required: missing properties: ["trigger"]`).

### 4. TUI — `L` on a trigger row ✅
Ran the TUI in tmux, entered automation mode (`m`), selected the `fire` trigger row:
- Footer hint now shows **`L: logs`** on a trigger row.
- Pressing **`L`** opened the concatenated event logs in `less` (pager `(END)`), content matching the CLI/MCP output; temp file `/tmp/fleet-trigger-log-*.log` used.
- Pressing **`L`** on the never-fired `never` trigger showed the status message `No events recorded for trigger "never"` instead of an empty pager.
- Settings → **Keybindings** overlay lists the new entry **`L  View trigger event logs`** (and `L  View logs` for the instance view).

### 5. Safety / regression sanity ✅
- **Path-escape guard:** `fleet trigger logs qafleet "../../../etc/passwd"` sanitized to a single safe segment and returned "No events recorded" — no traversal; logs tree contained only safe dirs.
- Adjacent automation CLI unaffected: `trigger list`, `agent list` (TRIGGERS count), `trigger create`/`delete` all work; a freshly-created trigger reports "No events recorded" until it fires.

## Notes
- The **webhook** firing path was not driven end-to-end (it requires the gateway webhook endpoint wired up). It shares the exact same `fireTriggerAgents` → `logTriggerEvent` code as the schedule path verified live; only the per-kind header line (`# webhook:`) and the payload (raw request body) differ, which are covered by the PR's unit tests.

**Result:** All acceptance criteria met across TUI / CLI / MCP. Applying `qa-passed`.

BenjaminBenetti:
Thanks for the thorough QA pass — appreciate the live coverage of all four surfaces plus the path-escape and regression checks.

One note: QA ran against `9ca387c`; there's since a follow-up commit `48f57a3` addressing the CodeReview bot. None of it changes the flows you verified — it only (1) makes same-second collision filenames sort chronologically (`_NNNN` suffix; matters only for >1 firing within the same second), (2) tightens the log perms to `0700`/`0600`, and (3) fixes doc comments. The header/payload format, CLI/MCP/TUI behavior, and the "No events recorded" / path-escape handling you saw are all unchanged.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the per-trigger event-log feature (#206). This is a solid, carefully-built change — the path-escape sanitizer, `0600`/`0700` perms on potentially-secret webhook bodies, best-effort logging that never blocks a firing, the same-second collision-suffix that keeps lexical order chronological, and the thorough test coverage across server/CLI/MCP/TUI are all well done. A few things worth weighing before merge:

**1. `TriggerLogs` is unary and buffers the whole concatenation — a large reply exceeds gRPC's 4 MB default and fails (moderate).**
`readTriggerLogs` concatenates up to 100 firings' payloads into a single `TriggerLogsReply.logs` string. The fleet client dials with only `insecure.NewCredentials()` (no `grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(...))` anywhere in `fleetclient`), so the client's **default 4 MB receive cap** applies. Webhook bodies are arbitrary external input (GitHub allows payloads up to ~25 MB), so:
- a *single* large webhook firing can make even its own log unreadable, and
- ~100 moderate (~40 KB) payloads cross 4 MB and make the entire history unviewable from CLI/TUI/MCP — the call fails with `ResourceExhausted` rather than degrading gracefully.

The pre-existing `Logs` RPC sidesteps this by streaming. Options: stream `TriggerLogs` too, tail to a byte budget (most-recent-N-bytes), or raise `MaxCallRecvMsgSize`. A per-file write cap would also bound disk growth. (See inline note.)

**2. `safeLogSegment` can collapse two distinct names to the same directory (low–moderate).**
Trigger names are validated only as non-empty + unique-within-fleet (`NormalizeTrigger`); there is no charset restriction. `safeLogSegment` maps every rune outside `[A-Za-z0-9-._]` to `_`, so two legitimately-distinct triggers like `build/test` and `build_test` resolve to the same `…/trigger/build_test/` dir. Their firings then interleave, share the 100-file cap, and one trigger's `logs` view surfaces the other's webhook payloads (which may carry secrets). The same applies to fleet names. Consider disambiguating collisions (e.g. append a short hash of the raw name) or restricting trigger names to a safe charset at creation. (See inline note.)

**3. Log directories are orphaned on delete/rename (minor).**
Deleting a trigger doesn't remove `~/.fleet/logs/<fleet>/trigger/<trigger>/`, and `fleet_trigger_update` with `new_name` starts logging under the new sanitized name while the old dir's logs become unreachable. Not harmful, but the on-disk history leaks and silently "disappears" after a rename.

**4. "debug what fired (or failed to fire)" is only half captured (minor / docs).**
`logTriggerEvent` runs inside `fireTriggerAgents`, which a webhook only reaches *after* filter-match and only when the trigger has ≥1 live agent (`fireWebhookBatch` `continue`s otherwise with just a `flog.Warn`). So filtered-out or agent-less deliveries — the common "why didn't it fire?" cases — aren't recorded here. Worth softening the wording or logging those misses too.

Nits: the schedule log repeats the fire-time in both the `# fired:` header and the payload (cosmetic); `readTriggerLogs` always returns a `nil` error, so the handler's `codes.Internal` branch is currently dead.

Nothing here is a hard blocker — (1) and (2) are the ones I'd most want addressed. Nice work overall.

internal/server/trigger_logs.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	sort.Strings(names)
+	return names
+}
+
+// readTriggerLogs concatenates a trigger's event logs (oldest first) for
+// viewing, each preceded by a separator naming its file. It returns the text and
+// the number of logs included. A trigger that has never fired (no directory)
+// yields "" and 0 — not an error — so callers can report "no events" cleanly.
+func readTriggerLogs(fleetName, triggerName string) (string, int, error) {
```
BenjaminBenetti, line 141: 🔍 **CodeReview Agent**

`readTriggerLogs` buffers all (≤100) event logs into one string that becomes a single unary `TriggerLogsReply`. The fleet client dials with only `insecure.NewCredentials()` — no `MaxCallRecvMsgSize` override anywhere in `fleetclient` — so the gRPC **default 4 MB receive cap** applies. Webhook bodies are arbitrary external input (GitHub permits up to ~25 MB), so one large firing, or ~100 × ~40 KB firings, pushes the reply past 4 MB and the RPC fails with `ResourceExhausted` for CLI/TUI/MCP — meaning *no* logs are viewable, not just the oldest pruned away. Consider streaming (like the existing `Logs` RPC), tailing to a byte budget, or capping per-file write size.

internal/state/paths.go
```diff
@@ -31,6 +32,37 @@ func WarnPath(fleetName, instanceName string) string {
...
+// the writer and the reader both go through here, so they always agree on the
+// same directory.
+func TriggerLogsDir(fleetName, triggerName string) string {
+	return filepath.Join(FleetDir(), "logs", safeLogSegment(fleetName), "trigger", safeLogSegment(triggerName))
+}
+
+// safeLogSegment reduces name to one filesystem-safe path segment: it keeps
+// letters, digits, and -._, maps everything else to '_', and collapses the
+// empty / "." / ".." forms (which would escape or alias a directory) to "_".
+func safeLogSegment(name string) string {
```
BenjaminBenetti, line 49: 🔍 **CodeReview Agent**

`safeLogSegment` maps every rune outside `[A-Za-z0-9-._]` to `_`, but trigger/fleet names are only validated as non-empty + unique (`NormalizeTrigger`), not charset-restricted. So two distinct, individually-valid triggers such as `build/test` and `build_test` collapse to the same `…/trigger/build_test/` directory — their firings interleave, share the 100-file cap, and one trigger's log view leaks the other's webhook payloads (possible secrets). Sanitizing for path-safety is the right instinct; consider also disambiguating collisions (e.g. suffix a short hash of the raw name) or restricting names to a safe charset at creation so two names can't alias one directory.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

**Verdict: ✅ PASS** — I manually exercised the new per-trigger event-log feature end to end against a real `fleetd` daemon. The write path and all three read surfaces (CLI, MCP, TUI) behave as described in the ticket. No defects found.

## Environment
- Isolated Docker container (`golang:1.25-trixie`), PR branch `feat/trigger-event-logs` @ `48f57a3`. Built `go build ./cmd/fleet` (full-tree `go build ./...` also clean).
- Drove a real daemon by seeding `~/.fleet/state.json` with a fleet + `writer` agent + two **schedule** triggers: `fire` (cron `* * * * *`) and `quiet` (cron `0 0 1 1 *`, never matches now), with `FLEET_AUTOMATION_TICK_INTERVAL=2s` so the scheduler's startup tick fires `fire` immediately.

## What I tested

### ✅ Write path — a firing records an event log
On daemon start the scheduler fired `fire` and the daemon wrote exactly the documented path:
```
~/.fleet/logs/qafleet/trigger/fire/event-20260623T022403Z.log
```
- Directory perms `700`, file perms `600` (matches the stated "webhook bodies may carry secrets" safety intent).
- Header + payload exactly as designed — and **no** `# webhook:` line for a schedule firing:
  ```
  # trigger: fire
  # type:    schedule
  # fired:   2026-06-23T02:24:03Z

  2026-06-23T02:24:03Z
  ```
- `quiet` (cron never matches) produced **no** directory — correct.
- Later minutes accumulated additional logs; filenames sort chronologically.

### ✅ CLI — `fleet trigger logs`
- `fleet trigger logs qafleet fire` → every firing concatenated, each under a `===== event-<ts>.log =====` separator, **oldest first**.
- `fleet trigger logs qafleet quiet` → `No events recorded for trigger "quiet" in fleet "qafleet"`.
- `fleet trigger logs --help` documents the command and the 100-firing retention.

### ✅ MCP — `fleet_trigger_logs`
Driven over the daemon's StreamableHTTP MCP endpoint (bearer token from `~/.fleet/mcp.token`):
- `{fleet:qafleet, trigger:fire}` → `{"count":2,"logs":"..."}` with the same concatenated, oldest-first output.
- `{fleet:qafleet, trigger:quiet}` → `{"count":0,"logs":""}`.
- Missing `trigger` arg → tool error (`required: missing properties: ["trigger"]`).
- Tool is present in `tools/list`.

### ✅ TUI — `L` on a trigger row
- Entered the automation view (`m`) and selected the `fire` trigger.
- The contextual help bar now shows **`L: logs`**.
- `L` paged the concatenated logs in `less` (all firings, oldest-first; temp file `/tmp/fleet-trigger-log-*.log`).
- Quitting `less` returned cleanly to the TUI and **removed the temp file**.
- `L` on the empty `quiet` trigger showed the inline status `No events recorded for trigger "quiet"` instead of launching an empty pager.
- `?` → Keybindings overlay lists the new **Automation (trigger / agent view)** group, including `L  →  View trigger event logs`.

### ✅ Regression sanity
- `fleet trigger list` / `fleet agent list` still render correctly.
- `go build ./...` is clean.

## Not exercised live (verified by code inspection; covered by the PR's unit tests)
- **Webhook firing** (body-as-payload + the `# webhook:` header line): the webhook receiver is only served over the gateway tunnel (a `ChanListener`), not a local HTTP port, so a live POST would require standing up a full gateway tunnel. It funnels through the same `fireTriggerAgents → logTriggerEvent → triggerEventLogContent` machinery I exercised above via the schedule path, differing only in `kind`/`webhookName`/payload-is-body.
- **100-firing rotation** and the **same-second no-overwrite counter suffix** — impractical to drive in real time (1 firing/minute); logic reviewed and covered by `internal/server/trigger_logs_test.go`.

## Observations (non-blocking, by design)
- `fleet trigger logs` / `fleet_trigger_logs` on an **unknown fleet or trigger** returns an empty result ("No events recorded") rather than a not-found error. This is consistent with the documented "missing dir ⇒ empty, not an error" design, though it differs slightly from `trigger list/edit/delete`, which error on an unknown fleet. Harmless — noting only for awareness.

PR #208, opened 2026-06-23
fix/webhook-jsonpath-reject-non-json — reject non-JSON body for json-path webhook filters with 400 (#207)

### LINKED ISSUE #207: Webhook trigger with JSONPath filter never fires despite matching payload
## Summary

A webhook automation trigger that uses a `jsonpath` filter never fires, even though the configured `json_path` selects a value from the delivered payload that is exactly equal to the configured `json_value`. An equivalent `regex` filter on the same gateway endpoint and same event type fires correctly, so delivery/routing to the gateway is working — the problem appears to be in JSONPath filter evaluation.

## Trigger config

```
name:         hello-pr-benjamin-only
type:         webhook
webhook_name: github-test-2          # GitHub webhook configured with the full gateway URL for this name
filter_type:  jsonpath
json_path:    $.pull_request.user.login
json_value:   BenjaminBenetti
agents:       [HelloPR]
```

## Steps to reproduce

1. Create a webhook trigger with the `jsonpath` filter above.
2. Configure a GitHub webhook whose Payload URL is the full gateway URL for `github-test-2`, subscribed to **Pull requests** events.
3. Push a commit to a PR authored by `BenjaminBenetti` (produces a `pull_request` / `synchronize` delivery).

## Expected

The trigger fires — `json_path` selects `"BenjaminBenetti"`, which equals `json_value` — and the `HelloPR` agent instance is spun up.

## Actual

The trigger does not fire. No agent instance is created. No comment is posted.

## Delivered payload (relevant excerpt)

```json
{
  "action": "synchronize",
  "number": 205,
  "pull_request": {
    "user": { "login": "BenjaminBenetti" }
  },
  "sender": { "login": "BenjaminBenetti" }
}
```

`pull_request.user.login` = `"BenjaminBenetti"`, which matches `json_value` exactly.

## Notes / hypotheses

- A `regex` filter on the **same** webhook endpoint and the **same** event type fires correctly, which rules out gateway routing / delivery as the cause and isolates this to JSONPath filter evaluation.
- Possible JSONPath dialect mismatch: the evaluator may not accept the `$.` root prefix. Some Go path libraries (e.g. gjson-style dotted paths) expect `pull_request.user.login` and treat a leading `$` as a literal key, so the selector resolves to nothing and the equality check silently fails. If that's the case, either the `$.` prefix should be supported or the expected syntax should be documented/validated.

## Suggested investigation

- Confirm which JSONPath library/dialect the gateway uses and whether `$.`-prefixed paths are supported.
- Add logging when a `jsonpath` filter selects nothing or fails to match, so silent non-matches are diagnosable.
- Consider validating `json_path` at trigger-create time and rejecting/​warning on unsupported syntax.

BenjaminBenetti:
## Summary

Fixes #207 — a webhook trigger using a **json-path** filter never fired even though its `json_path` selected a value equal to `json_value`, while an equivalent **regex** filter on the same endpoint fired fine.

The json-path syntax (`$.pull_request.user.login`) was never the problem — that resolves correctly. The real cause is the **webhook content type**: GitHub's webhook config defaults to `application/x-www-form-urlencoded`, which delivers the event as a `payload=<url-encoded-json>` form field instead of raw JSON. So:

- the **regex** filter matches raw bytes → `BenjaminBenetti` is still literally present in the form body → it fires;
- the **json-path** filter does `json.Decode(body)` → the `payload=` prefix isn't valid JSON → decode fails → it silently never matches.

That's the exact asymmetry in the report.

Rather than guess at sender-specific form encodings (their shape isn't reliably recoverable — only GitHub uses `payload=`), this **rejects the delivery with `400 Bad Request`** when an active json-path trigger carries the webhook name but the body isn't JSON. fleetd's 400 reverse-proxies back through the gateway, so the misconfiguration is **visible in the sender's delivery dashboard** (e.g. GitHub's "Recent Deliveries") instead of vanishing. The error tells the user what to do: set the webhook content type to `application/json`.

Regex filters match raw bytes and accept any body — they are unaffected.

### Behavior

- json-path trigger + JSON body → unchanged (matches/fires as before).
- json-path trigger + non-JSON body (form-encoded, plain text, empty) → **400**, nothing fires.
- regex trigger + any body → unchanged (fires on raw-byte match).
- A name carrying **both** a regex and a json-path trigger + non-JSON body → the 400 wins (nothing fires), so the misconfiguration stays visible and a sender retry can't double-fire the regex trigger.
- A **disabled** json-path trigger doesn't arm the check.

### Tests

- New `serveWebhook` cases: form-encoded/non-JSON body → 400 + no fire; regex still fires on a form-encoded body; mixed regex+json-path → 400 wins; disabled json-path doesn't reject.
- Updated the end-to-end tunnel test to assert the new 400 contract for a non-JSON json-path delivery (it previously encoded the buggy 200 behavior).

### DIFF
--- a/internal/server/webhook.go
+++ b/internal/server/webhook.go
@@ -1,7 +1,9 @@
 package server
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -63,11 +65,32 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	fires, nameFound := collectWebhookFires(st, name, body)
+	fires, nameFound, jsonPathActive := collectWebhookFires(st, name, body)
 	if !nameFound {
 		// No trigger anywhere carries this name. Same 404 the gateway gives an
 		// unknown id, so a probe can't enumerate configured webhook names.
 		http.Error(w, "no webhook trigger with that name", http.StatusNotFound)
+		return
+	}
+
+	// A json-path filter can only evaluate a JSON body. The classic trap is
+	// GitHub's webhook default content type, application/x-www-form-urlencoded,
+	// which delivers the event as a `payload=<url-encoded-json>` form field
+	// rather than raw JSON — so json.Decode fails and the filter silently never
+	// matches (issue #207). We don't try to unwrap form encodings (their shape is
+	// sender-specific and not reliably recoverable); instead, when an active
+	// json-path trigger carries this name but the body isn't JSON, reject the
+	// delivery with 400. fleetd's response is reverse-proxied straight back to the
+	// sender, so the failure shows up in the sender's delivery dashboard (e.g.
+	// GitHub's "Recent Deliveries") instead of vanishing. Regex triggers match raw
+	// bytes and accept any body, so they are unaffected by this check.
+	//
+	// If a name carries BOTH a regex and a json-path trigger and a non-JSON body
+	// arrives, the 400 wins (nothing fires): returning 200 would hide the
+	// misconfiguration, and firing the regex trigger then returning 400 would
+	// double-fire it on the sender's retry.
+	if jsonPathActive && !bodyIsJSON(body) {
+		http.Error(w, "json-path webhook filter requires a JSON body; set the webhook content type to application/json", http.StatusBadRequest)
 		return
 	}
 
@@ -97,12 +120,14 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 }
 
 // collectWebhookFires scans every fleet for webhook triggers named name and
-// returns the subset whose filter matches body, plus whether ANY trigger with
-// that name exists at all (to distinguish 404 from "matched nothing"). Firing all
+// returns the subset whose filter matches body, whether ANY trigger with that
+// name exists at all (to distinguish 404 from "matched nothing"), and whether any
+// ACTIVE (non-disabled) trigger with that name uses the json-path filter (so the
+// caller can reject a non-JSON body with 400 — see serveWebhook). Firing all
 // matches across all fleets is intentional: webhook names are per-fleet in the
 // model, so the same name can legitimately drive triggers in several fleets, and
 // the common single-match case degenerates to exactly one fire.
-func collectWebhookFires(st *state.State, name string, body []byte) (fires []webhookFire, nameFound bool) {
+func collectWebhookFires(st *state.State, name string, body []byte) (fires []webhookFire, nameFound, jsonPathActive bool) {
 	for fleetName, f := range st.Fleets {
 		if f == nil {
 			continue
@@ -115,12 +140,26 @@ func collectWebhookFires(st *state.State, name string, body []byte) (fires []web
 			if t.Disabled {
 				continue // a disabled trigger still exists (200, not 404) but never fires
 			}
+			if t.FilterType == fleet.WebhookFilterJSONPath {
+				jsonPathActive = true
+			}
 			if t.MatchesWebhook(body) {
 				fires = append(fires, webhookFire{fleet: fleetName, trigger: t, body: body})
 			}
 		}
 	}
-	return fires, nameFound
+	return fires, nameFound, jsonPathActive
+}
+
+// bodyIsJSON reports whether body decodes as a single JSON value — the same parse
+// MatchesWebhook's json-path filter performs (json.Decoder, so trailing
+// whitespace is fine and trailing junk is ignored, matching the matcher exactly).
+// It is used to reject a json-path delivery whose body isn't JSON, rather than let
+// the filter silently never match. The body is already size-bounded by the
+// MaxBytesReader in serveWebhook.
+func bodyIsJSON(body []byte) bool {
+	var v any
+	return json.NewDecoder(bytes.NewReader(body)).Decode(&v) == nil
 }
 
 // enqueueWebhookFires hands a request's matched events to the scheduler goroutine
--- a/internal/server/webhook.go
+++ b/internal/server/webhook.go
@@ -88,7 +88,14 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	// If a name carries BOTH a regex and a json-path trigger and a non-JSON body
 	// arrives, the 400 wins (nothing fires): returning 200 would hide the
 	// misconfiguration, and firing the regex trigger then returning 400 would
-	// double-fire it on the sender's retry.
+	// double-fire it on the sender's retry. Note this is keyed on the webhook NAME
+	// globally: jsonPathActive is OR'd across EVERY fleet (collectWebhookFires
+	// scans them all, since one name can drive triggers in several fleets), so a
+	// json-path trigger for this name in ANY fleet makes a non-JSON delivery 400 —
+	// even for a regex trigger of the same name in a DIFFERENT fleet. That cross-
+	// fleet blast radius is intended: a non-JSON delivery to a name any fleet
+	// treats as json-path is a misconfiguration worth surfacing loudly, not
+	// silently half-firing whichever same-named regex triggers happen to match.
 	if jsonPathActive && !bodyIsJSON(body) {
 		http.Error(w, "json-path webhook filter requires a JSON body; set the webhook content type to application/json", http.StatusBadRequest)
 		return
--- a/internal/server/webhook_remote_e2e_test.go
+++ b/internal/server/webhook_remote_e2e_test.go
@@ -216,22 +216,24 @@ func TestWebhookJSONPathDeliveryEndToEnd(t *testing.T) {
 	}
 
 	// Malformed (non-JSON) event: the most transport-sensitive case — a json-path
-	// filter must PARSE the delivered body, so a body that survives the tunnel but
-	// isn't JSON resolves no path and fires nothing, yet the name is still wired so
-	// the receiver answers 200 (not 404). This exercises the decode-failure branch
-	// end to end, which a regex filter (raw-byte match) never reaches.
+	// filter can only evaluate a JSON body, so a body that survives the tunnel but
+	// isn't JSON (e.g. GitHub's default form-urlencoded delivery) is rejected with
+	// 400 (issue #207) rather than silently never matching. The name is wired (so
+	// it's a 400, not a 404), nothing fires, and the 400 reverse-proxies back
+	// through the tunnel so the sender sees the failure. A regex filter (raw-byte
+	// match) never reaches this branch.
 	resp3, err := client.Post(base+"/deploy", "application/json",
 		strings.NewReader(`refs/heads/main not json at all`))
 	if err != nil {
 		t.Fatalf("POST malformed json-path webhook through the tunnel: %v", err)
 	}
 	resp3.Body.Close()
-	if resp3.StatusCode != http.StatusOK {
-		t.Fatalf("malformed json-path webhook POST -> %d, want 200 (name wired, body unparseable)", resp3.StatusCode)
+	if resp3.StatusCode != http.StatusBadRequest {
+		t.Fatalf("malformed json-path webhook POST -> %d, want 400 (json-path needs a JSON body)", resp3.StatusCode)
 	}
 	select {
 	case batch := <-svc.webhookFires:
-		t.Fatalf("json-path filter must not fire on an unparseable body, got %+v", batch)
+		t.Fatalf("a rejected json-path body must fire nothing, got %+v", batch)
 	case <-time.After(500 * time.Millisecond):
 		// expected: nothing enqueued
 	}
--- a/internal/server/webhook_test.go
+++ b/internal/server/webhook_test.go
@@ -311,6 +311,48 @@ func TestServeWebhook_JSONPathNonJSONRejected(t *testing.T) {
 	}
 }
 
+// TestServeWebhook_JSONPathEmptyBodyRejected covers an empty body hitting a
+// json-path trigger: bodyIsJSON("") is io.EOF -> false, so it's a 400 like any
+// other non-JSON body.
+func TestServeWebhook_JSONPathEmptyBodyRejected(t *testing.T) {
+	s := newService()
+	st := webhookState(fleet.Trigger{
+		Name: "deploy", Type: fleet.TriggerWebhook, AgentNames: []string{"a"},
+		WebhookName: "deploy", FilterType: fleet.WebhookFilterJSONPath,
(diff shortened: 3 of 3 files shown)


BenjaminBenetti:
hello world

BenjaminBenetti:
hello world

BenjaminBenetti:
hello world

BenjaminBenetti:
hello world

BenjaminBenetti:
hello world

BenjaminBenetti:
hello world

BenjaminBenetti:
hello world

internal/server/webhook.go
```diff
@@ -63,14 +65,35 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
...
+	// delivery with 400. fleetd's response is reverse-proxied straight back to the
+	// sender, so the failure shows up in the sender's delivery dashboard (e.g.
+	// GitHub's "Recent Deliveries") instead of vanishing. Regex triggers match raw
+	// bytes and accept any body, so they are unaffected by this check.
+	//
+	// If a name carries BOTH a regex and a json-path trigger and a non-JSON body
+	// arrives, the 400 wins (nothing fires): returning 200 would hide the
+	// misconfiguration, and firing the regex trigger then returning 400 would
+	// double-fire it on the sender's retry.
+	if jsonPathActive && !bodyIsJSON(body) {
```
BenjaminBenetti, line 99: 🔍 **CodeReview Agent**

`jsonPathActive` is accumulated across **all** fleets in `collectWebhookFires`, so this 400 is keyed on the webhook *name* globally, not on a single fleet. That means an active json-path trigger named e.g. `pr` in one fleet will cause non-JSON deliveries to `pr` to 400 — silently disabling a *regex* trigger of the same name in a **different** fleet that previously fired on GitHub's form-encoded body.

The PR description documents the same-config "regex + json-path under one name → 400 wins" tradeoff, so this is likely intended, but the cross-fleet coupling (editing fleet B changes fleet A's webhook behavior) is non-obvious. Worth confirming it's the desired blast radius — no change requested if so.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the diff for #208 (reject non-JSON body for json-path webhook filters with 400). Overall this is a clean, well-reasoned, and well-tested fix — the diagnosis in the PR description (GitHub's default `application/x-www-form-urlencoded` delivery breaks `json.Decode`) is correct and the chosen "fail loud with 400" approach is sound. A few notes, none blocking.

**Correctness — the parse alignment holds.** `bodyIsJSON` uses `json.NewDecoder(bytes.NewReader(body)).Decode(&v)`, which is exactly the decode `MatchesWebhook`'s json-path branch performs. The matcher additionally calls `dec.UseNumber()`, but that only changes how numbers are *materialized*, not whether `Decode` *succeeds*, so the "is this JSON?" decision is identical in both places — including the shared "first value wins, trailing junk ignored" semantics. The comment's claim of matching the matcher exactly is accurate. ✅

**Main thing worth confirming — the 400 is a global property of the webhook *name*, not a fleet.** `jsonPathActive` is OR'd across every fleet in `collectWebhookFires`, and the gateway routes a delivery to all fleets' triggers sharing that name. Consequence: adding an active json-path trigger named `pr` in fleet B will start returning **400** for non-JSON deliveries to `pr` — which means a *previously-working regex trigger named `pr` in fleet A stops firing* for the same GitHub form-encoded deliveries. The PR doc covers the "one name carries both a regex and a json-path trigger" case and argues the 400 should win (avoid double-fire on retry, keep the misconfig visible), so this is presumably intended — but the cross-fleet blast radius (touching fleet B silently changes fleet A's behavior) is a non-obvious coupling worth calling out explicitly. If that's undesired, an alternative would be to scope the json-body requirement to the case where *no* active regex trigger with the same name would also match — but that reintroduces the double-fire-on-retry problem you deliberately avoided, so the current "400 wins" choice is defensible. Flagging for awareness, not asking for a change.

**Minor — response-code oracle (low priority).** The new 400 vs the unknown-name 404 lets a prober distinguish "name exists *and* has an active json-path trigger" by sending a non-JSON body. This isn't a new class of leak: under the prior code a known name already returned 200 (vs 404 for unknown), so name existence was already observable, and the module's documented model is "no auth, the unguessable URL is the capability." Noting only because the surrounding comments lean on the 404-prevents-enumeration property.

**Minor — test gaps (optional).** Two cases asserted/implied by the PR description but not directly tested:
- Empty body + active json-path trigger → 400 (the description lists "empty → 400"). `bodyIsJSON("")` is `io.EOF`→false so the logic covers it, but there's no explicit case.
- json-path + *valid JSON whose value doesn't match* → **200, not 400** (the important "don't over-reject" guard). This is covered indirectly by `TestServeWebhook_MultiFleetSelectiveFilter` (the staging trigger gets valid non-matching JSON and the overall response is 200), but a focused single-trigger case would document the 400-only-applies-to-non-JSON boundary more clearly.

The new tests themselves are thorough (form-encoded, plain-text, regex-still-fires, mixed-400-wins, disabled-doesn't-arm) and the e2e update correctly replaces the old buggy 200 contract. Nice work on the comments — they make the intent easy to follow.

BenjaminBenetti:
hello world

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build + launch real gateway & daemon stack | ✅ Pass | `fleet gateway` + `fleet server` over a real tunnel; webhook feature negotiated, Public Webhook URL minted |
| A1 · json-path + matching JSON body | ✅ Pass | `200` · `1 trigger(s) fired` — unchanged happy path |
| A2 · json-path + valid JSON, value does **not** match | ✅ Pass | `200` · `no trigger matched` — valid JSON is **not** rejected (key correctness boundary) |
| A3 · json-path + GitHub form-encoded body (`payload=…`) | ✅ Pass | `400` — reproduces & fixes #207; nothing fired |
| A4 · json-path + plain-text body | ✅ Pass | `400` — any non-JSON rejected |
| A5 · json-path + empty body | ✅ Pass | `400` — empty body rejected |
| B1 · regex + form-encoded body containing match | ✅ Pass | `200` · `1 trigger(s) fired` — regex unaffected (raw-byte match) |
| B2 · regex + JSON body containing match | ✅ Pass | `200` · `1 trigger(s) fired` |
| B3 · regex + body without match | ✅ Pass | `200` · `no trigger matched` |
| C1 · same name w/ regex **and** json-path + non-JSON body | ✅ Pass | `400` wins, nothing fired (regex `.*` did **not** double-fire) |
| C2 · same name w/ regex + json-path + matching JSON | ✅ Pass | `200` · `2 trigger(s) fired` — both fire |
| D1 · **disabled** json-path + non-JSON body | ✅ Pass | `200` · `no trigger matched` — disabled trigger does **not** arm the 400 check |
| E1 · unknown webhook name (regression sanity) | ✅ Pass | `404` — unchanged |

**Overall: ✅ PASS — 13 of 13 scenarios passing. The #207 asymmetry is fixed and no existing behavior regressed.**

---

### How this was tested (real, end-to-end — not the unit suite)

I stood up the **actual product stack** inside a throwaway Docker container and drove it exactly as a remote sender (GitHub/CI/`curl`) would:

1. Built the `fleet` binary from this branch (`go build ./cmd/fleet`).
2. Ran the **real `fleet gateway`** (plain-HTTP loopback) and the **real `fleet server` daemon**. The daemon dialed out, registered, and negotiated the webhook tunnel feature — the gateway minted a Public Webhook URL (`…/webhook/<id>`), confirmed in `~/.fleet/gateway_session.json` and the gateway log (`tunnel registered`).
3. Authored a fleet `alpha` in `state.json` with five webhook triggers exercising every branch: a json-path trigger (`pr-jp`, `$.pull_request.user.login == BenjaminBenetti`), a regex trigger (`pr-rx`, `BenjaminBenetti`), a **mixed** name carrying both a regex (`.*`) and a json-path (`$.action == opened`) trigger (`pr-mix`), and a **disabled** json-path trigger (`pr-dis`).
4. `POST`ed real HTTP events to `<public-webhook-url>/<name>` and recorded the status code **and** response body (which reports the fired count).

Each result was **independently corroborated server-side** by the per-trigger event logs under `~/.fleet/logs/<fleet>/trigger/<name>/`: `jp` logged a firing only on matching JSON, `rx` logged two firings (form-encoded + JSON), `mixrx`+`mixjp` each logged a firing on C2, and there is **no** event log for any 400 case nor for the disabled trigger — matching the HTTP responses exactly.

### Key observations vs. issue #207

- The reported asymmetry is reproduced and resolved: on the **same endpoint**, a **regex** filter fires on a GitHub form-encoded body (B1 → `200`, fired), while a **json-path** filter on that same form-encoded body no longer silently does nothing — it now returns **`400`** (A3) with an actionable message:
  > `json-path webhook filter requires a JSON body; set the webhook content type to application/json`
  Because fleetd's `400` reverse-proxies back through the gateway, the misconfiguration is now visible in the sender's delivery dashboard instead of vanishing.
- The fix is correctly **scoped to non-JSON bodies only**: a valid JSON body whose selected value doesn't match still returns `200 no trigger matched` (A2) — legitimate non-matching deliveries are untouched.
- The documented edge cases hold precisely: the `400` wins over a co-located regex trigger (C1, no double-fire), and a **disabled** json-path trigger does not arm the check (D1 → `200`, not `400`).

### Notes / caveats (non-blocking)

- The QA fleet has no real repository/devcontainer, so when a trigger *does* fire, the **downstream agent-provisioning** step logs `fatal: The empty string is not a valid path`. This is expected for this harness and is **orthogonal to this PR** — the webhook routing, filter matching, and enqueue (all that #208 changes) completed successfully (`N trigger(s) fired`); only the unrelated agent spawn fails for lack of a repo/Docker workspace. No crash; daemon and gateway stayed healthy for the entire run.

_Build: ✅ · Gateway+daemon launch: ✅ · No regressions observed._

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Full stack boot (gateway + fleetd webhook tunnel) | ✅ Pass | Tunnel registered; public webhook URL minted, no startup errors |
| T1 · json-path + matching JSON | ✅ Pass | `200` → `received: 1 trigger(s) fired` |
| T2 · json-path + non-matching JSON | ✅ Pass | `200` → `no trigger matched the event` |
| T3 · json-path value present only in unrelated field | ✅ Pass | `200` no match — routes on JSON structure, not substring |
| T4 · json-path + GitHub form-encoded body (**core fix, #207**) | ✅ Pass | `400` with actionable message; nothing fired |
| T5 · json-path + empty body | ✅ Pass | `400` |
| T6 · json-path + plain-text non-JSON body | ✅ Pass | `400` |
| T7 · regex + non-JSON body | ✅ Pass | `200` fires — regex matches raw bytes, unaffected |
| T8 · mixed regex+json-path, same name + non-JSON | ✅ Pass | `400` wins; regex did **not** double-fire |
| T9 · mixed regex+json-path + valid JSON | ✅ Pass | `200` → `received: 2 trigger(s) fired` |
| T10 · **disabled** json-path + non-JSON body | ✅ Pass | `200` (not 400) — disabled trigger doesn't arm the check |
| T11 · unknown webhook name | ✅ Pass | `404` `no webhook trigger with that name` |
| T12 · regex + matching JSON (sanity) | ✅ Pass | `200` fires |

**Overall: ✅ PASS — 13 of 13 scenarios passing.** Every behavior in the PR's contract reproduced exactly, including the #207 fix and all stated edge cases.

---

### How it was tested
Manual, behavior-level QA against the **real running stack** — no automated tests were run. Inside a throwaway Docker container (`golang:1.25`) I built the `fleet` binary from this branch (HEAD `05ebcfd`) and stood up the genuine production path:

- a real **`fleet gateway`** (cert-less HTTP public + h2c gRPC) on `127.0.0.1:8443` / `:50051`;
- a real **`fleet server`** (fleetd) with `webhook_enabled` + `gateway_url` configured, which dialed out and established the webhook tunnel (gateway minted `…/webhook/963741e7…`);
- a fleet `alpha` carrying five webhook triggers (`deploy` json-path, `ci` regex, `both` = regex **and** json-path on one name, `off` = disabled json-path).

Events were delivered with `curl` to the gateway's public webhook URL exactly as a sender (e.g. GitHub) would — so each request traversed **HTTPS edge → gateway `/webhook` route → tunnel → fleetd receiver → matcher → scheduler**. Results were confirmed two ways: the HTTP status/body returned to the caller, and fleetd's structured `webhook: received … fired=N` log lines.

### Detailed findings

**Core fix — non-JSON body to a json-path filter (#207).** The exact failure from the report. A GitHub-style `application/x-www-form-urlencoded` delivery (`payload=%7B…%7D`), an empty body, and a plain-text body were each POSTed to `/deploy` (json-path `$.ref == refs/heads/main`). All three returned **`400 Bad Request`** with `json-path webhook filter requires a JSON body; set the webhook content type to application/json`. fleetd logged **no** `webhook: received` line for any of them — the delivery is rejected before the fire path, so nothing fires and the failure is visible to the sender (reverse-proxied straight back through the gateway). Previously this body silently never matched.

**Regex unaffected.** A non-JSON `text/plain` body literally containing `"action":"opened"` POSTed to the regex trigger `/ci` returned `200` and `fired=1` — regex still matches raw bytes and accepts any body.

**"400 wins" on a mixed name.** Webhook name `both` carries one regex and one json-path trigger. A non-JSON body that the regex *would* match returned **`400`** (not `200`) and fired **nothing** (`/both` produced no `fired` log line). This confirms the intended precedence: the misconfiguration stays loudly visible and a sender retry can't double-fire the regex trigger. The same name with a valid matching JSON body returned `200` → `2 trigger(s) fired`, so legitimate deliveries are unaffected.

**Disabled trigger doesn't arm the check.** Name `off` has only a *disabled* json-path trigger. A non-JSON body returned `200` (`no trigger matched`), not `400` — a disabled trigger never arms the json-path rejection.

**Routing correctness preserved.** json-path matching still routes on JSON structure: a body where `$.ref` differs but the configured value appears in an unrelated field did **not** fire (`200`, no match). Unknown names still `404`. Valid matching JSON for both json-path and regex triggers fires correctly (`fired=1` / `fired=2`).

**Note (not a defect):** fleetd logged `instance create failed … git clone failed … empty string is not a valid path` after the firing scenarios. This is expected from the **test harness**, not the PR: the minimal QA fleet has no real repo configured, so the downstream agent-dispatch (which happens only *after* a fire is enqueued) can't build a devcontainer. It is orthogonal to the webhook routing under test and actually confirms fires were genuinely enqueued (5 creates = deploy 1 + ci 1 + both 2 + ci 1). The 400/404 cases produced no such activity.

PR #209, opened 2026-06-24
feat(logrotate): rotate fleet.log daily at 3am, keep 100 days

BenjaminBenetti:
## Problem

`~/.fleet/fleet.log` is the append-only event trail every fleet process writes to (see `internal/flog`), and nothing ever trimmed it. On a long-lived host it grows without bound.

## Change

A new server-side background loop (`internal/server/logrotate.go`) cuts the log once a day:

- At **3am local time** it copies the current `fleet.log` contents into `~/.fleet/logs/fleetd/<date>.log`, truncates the live log back to empty, and **prunes dated logs older than 100 days**.
- Launched alongside the backup loop in `Serve()`; unconditional, stops on shutdown via `hubCtx`.

### Why copytruncate, not rename

`fleet.log` is appended to with `O_APPEND` by **every fleet process at once** — the long-lived daemon and TUI, one-shot CLIs, and detached `_create`/`_clone` children — each holding its own descriptor to the same inode for its whole lifetime, and `flog` opens that descriptor exactly once with **no reopen path**. A rename would orphan all those descriptors onto the rotated file, so the daemon and TUI would keep writing into yesterday's log while the fresh `fleet.log` only collected records from newly-started processes. Truncating the inode in place keeps every descriptor valid: because all writers use `O_APPEND`, their next write lands at the new end-of-file with no reopen and no coordination. The narrow copytruncate race (records written between the copy reaching EOF and the truncate) is acceptable — `flog` is explicitly best-effort.

### Behaviour notes

- **Due-ness is filesystem-derived** (past the hour *and* today's dated log absent), so rotation is idempotent and catches up across daemon restarts with no in-memory bookkeeping.
- Empty/missing `fleet.log` is left alone — quiet days produce no empty dated files.
- Schedule, naming, and pruning use **local time** to match "3am system time" (daily buckets are immune to the DST double-hour that pushes the backup loop to UTC).
- Dated copy preserves `fleet.log`'s `0644`.
- Test seams via env: `FLEET_LOG_ROTATE_HOUR`, `FLEET_LOG_ROTATE_KEEP_DAYS`, `FLEET_LOG_ROTATE_INTERVAL`.

## Tests

`internal/server/logrotate_test.go` covers: before-hour no-op, rotate-creates-dated-file + truncates (+ perms), same-day idempotency (never clobbers today's log), empty/missing skip, **the shared `O_APPEND` fd continuing into the truncated log after rotation** (the core guarantee a rename would break), date-window pruning, date parsing, and the env helper.

### DIFF
--- a/internal/server/logrotate.go
+++ b/internal/server/logrotate.go
@@ -43,11 +43,13 @@ import (
 var (
 	// logRotateHour is the local hour-of-day (0–23) at which fleet.log is cut.
 	// The spec is 3am; overridable via FLEET_LOG_ROTATE_HOUR so a test can target
-	// the current hour.
-	logRotateHour = envIntDefault("FLEET_LOG_ROTATE_HOUR", 3)
-	// logRotateKeepDays bounds how many days of dated logs are kept. Anything
-	// older than now-keepDays is pruned after each rotation check. Overridable via
-	// FLEET_LOG_ROTATE_KEEP_DAYS for tests.
+	// the current hour. Clamped to a real hour so a fat-fingered override (e.g.
+	// 25) can't make the "before the rotation hour" gate unsatisfiable and turn
+	// every check into a rotation.
+	logRotateHour = min(max(envIntDefault("FLEET_LOG_ROTATE_HOUR", 3), 0), 23)
+	// logRotateKeepDays is how many of the most recent daily logs to keep (today
+	// inclusive). After each rotation check, logs older than that window are
+	// pruned. Overridable via FLEET_LOG_ROTATE_KEEP_DAYS for tests.
 	logRotateKeepDays = envIntDefault("FLEET_LOG_ROTATE_KEEP_DAYS", 100)
 	// logRotateInterval is how often the loop checks whether a rotation is due.
 	// A daily cut needs no fine granularity; ~15m lands the rotation within a
@@ -229,21 +231,33 @@ func pruneRotatedLogs(now time.Time) (int, error) {
 		return 0, err
 	}
 
-	// Cut at the start of the day logRotateKeepDays days back: a log dated before
-	// it is more than keepDays days old and is removed; today's just-written log
-	// (and the keepDays days behind it) is kept.
-	cutoff := startOfDay(now).AddDate(0, 0, -logRotateKeepDays)
+	// Keep the most recent logRotateKeepDays days, today inclusive: the window
+	// starts keepDays-1 days before today, so a log dated before that start is
+	// outside the window and removed. With the default 100 that retains today plus
+	// the 99 days behind it — 100 daily logs.
+	cutoff := startOfDay(now).AddDate(0, 0, -(logRotateKeepDays - 1))
 	removed := 0
 	for _, e := range entries {
 		if e.IsDir() {
 			continue
 		}
-		day, ok := parseRotatedLogDate(e.Name())
+		name := e.Name()
+		// Reclaim a temp file orphaned by a rotation that died before its rename
+		// (the deferred cleanup never ran). Safe here: rotation is serial and the
+		// current tick's copyTruncate has already finished by the time prune runs,
+		// so any .rotate-*.log.tmp left behind is from a dead run. Mirrors how
+		// pruneBackups sweeps its own orphaned .backup-*.tmp files; without it these
+		// would accumulate forever, since the date filter below skips them.
+		if strings.HasPrefix(name, ".rotate-") && strings.HasSuffix(name, ".log.tmp") {
+			_ = os.Remove(filepath.Join(dir, name))
+			continue
+		}
+		day, ok := parseRotatedLogDate(name)
 		if !ok {
 			continue
 		}
 		if day.Before(cutoff) {
-			if err := os.Remove(filepath.Join(dir, e.Name())); err == nil {
+			if err := os.Remove(filepath.Join(dir, name)); err == nil {
 				removed++
 			}
 		}
--- a/internal/server/logrotate.go
+++ b/internal/server/logrotate.go
@@ -49,8 +49,10 @@ var (
 	logRotateHour = min(max(envIntDefault("FLEET_LOG_ROTATE_HOUR", 3), 0), 23)
 	// logRotateKeepDays is how many of the most recent daily logs to keep (today
 	// inclusive). After each rotation check, logs older than that window are
-	// pruned. Overridable via FLEET_LOG_ROTATE_KEEP_DAYS for tests.
-	logRotateKeepDays = envIntDefault("FLEET_LOG_ROTATE_KEEP_DAYS", 100)
+	// pruned. Overridable via FLEET_LOG_ROTATE_KEEP_DAYS for tests. Floored at 1
+	// so a 0 override can't push the cutoff into the future and prune the log just
+	// rotated; the smallest window keeps today alone.
+	logRotateKeepDays = max(envIntDefault("FLEET_LOG_ROTATE_KEEP_DAYS", 100), 1)
 	// logRotateInterval is how often the loop checks whether a rotation is due.
 	// A daily cut needs no fine granularity; ~15m lands the rotation within a
 	// quarter hour of 3am while keeping the check cheap. Overridable via
--- a/internal/server/logrotate.go
+++ b/internal/server/logrotate.go
@@ -1,0 +1,258 @@
+package server
+
+import (
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/BenjaminBenetti/fleet-man/internal/fleetpaths"
+	"github.com/BenjaminBenetti/fleet-man/internal/flog"
+)
+
+// logrotate.go bounds the size of the shared event log. ~/.fleet/fleet.log is
+// the append-only trail every fleet process writes to (see internal/flog), and
+// nothing ever trims it, so on a long-lived host it grows without bound. This
+// loop cuts it once a day: at 3am local time it copies the current contents to
+// ~/.fleet/logs/fleetd/<date>.log, truncates the live log back to empty, and
+// drops dated logs past the retention window (the last 100 days).
+//
+// Why copytruncate and not a rename. fleet.log is appended to with O_APPEND by
+// EVERY fleet process at once — the long-lived daemon and TUI, one-shot CLIs,
+// and the detached _create/_clone children — each holding its own file
+// descriptor to the same inode for its whole lifetime, and flog opens that
+// descriptor exactly once with no reopen path. Renaming the file would leave all
+// those descriptors pointing at the renamed inode, so the daemon and TUI would
+// keep writing into the rotated file while the fresh fleet.log collected only
+// records from processes started afterwards. Truncating the inode in place keeps
+// every open descriptor valid: because all writers use O_APPEND, their next
+// write lands at the new end-of-file (offset 0) with no reopen and no
+// cross-process coordination. The cost is copytruncate's well-known race —
+// records appended in the brief window between the copy reaching EOF and the
+// truncate are dropped — which is acceptable because flog is explicitly
+// best-effort (swallow-and-continue).
+//
+// Like the backup loop it runs unconditionally (rotation must happen whether or
+// not a TUI is connected) and stops on shutdown via the context it is launched
+// with. The schedule and date naming use LOCAL time to match "3am system time";
+// daily buckets are immune to the DST double-hour that pushes the backup loop to
+// UTC.
+var (
+	// logRotateHour is the local hour-of-day (0–23) at which fleet.log is cut.
+	// The spec is 3am; overridable via FLEET_LOG_ROTATE_HOUR so a test can target
+	// the current hour.
+	logRotateHour = envIntDefault("FLEET_LOG_ROTATE_HOUR", 3)
+	// logRotateKeepDays bounds how many days of dated logs are kept. Anything
+	// older than now-keepDays is pruned after each rotation check. Overridable via
+	// FLEET_LOG_ROTATE_KEEP_DAYS for tests.
+	logRotateKeepDays = envIntDefault("FLEET_LOG_ROTATE_KEEP_DAYS", 100)
+	// logRotateInterval is how often the loop checks whether a rotation is due.
+	// A daily cut needs no fine granularity; ~15m lands the rotation within a
+	// quarter hour of 3am while keeping the check cheap. Overridable via
+	// FLEET_LOG_ROTATE_INTERVAL (a Go duration) so a test can drive the loop fast.
+	logRotateInterval = envDurationDefault("FLEET_LOG_ROTATE_INTERVAL", 15*time.Minute)
+)
+
+// envIntDefault returns the non-negative integer parsed from the named env var,
+// or def when it is unset, blank, or unparseable. Like envDurationDefault these
+// knobs exist only so tests can retarget the rotation hour / retention; the
+// daemon inherits the spawning client's environment.
+func envIntDefault(name string, def int) int {
+	if v := strings.TrimSpace(os.Getenv(name)); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n >= 0 {
+			return n
+		}
+	}
+	return def
+}
+
+// fleetdLogDir is the directory holding the rotated daily logs
+// (~/.fleet/logs/fleetd). Server-only, so it lives here rather than in the
+// client-shared fleetpaths package. The parent ~/.fleet/logs already holds the
+// per-instance creation logs.
+func fleetdLogDir() string {
+	return filepath.Join(fleetpaths.Dir(), "logs", "fleetd")
+}
+
+// runLogRotateLoop rotates fleet.log when due and prunes expired dated logs on
+// logRotateInterval. It returns when ctx is cancelled. It checks once
+// immediately so a daemon that comes up after 3am on a day it has not yet
+// rotated cuts the (possibly long-overgrown) log promptly instead of waiting a
+// full interval.
+func (s *service) runLogRotateLoop(ctx context.Context) {
+	ticker := time.NewTicker(logRotateInterval)
+	defer ticker.Stop()
+	for {
+		rotateLogTick(time.Now())
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+	}
+}
+
+// rotateLogTick performs one rotation-if-due plus a prune. Both steps are
+// best-effort: a failure is logged and the loop continues, since a rotation miss
+// must never take the daemon down. Split from the loop so tests can drive a
+// single tick. The "rotated" info line lands as the first record of the freshly
+// truncated fleet.log, so each cut is self-marking.
+func rotateLogTick(now time.Time) {
+	switch path, rotated, err := rotateLogIfDue(now); {
+	case err != nil:
+		flog.Warn("logrotate: rotation failed", "err", err)
+	case rotated:
+		flog.Info("logrotate: rotated fleet.log", "path", path)
+	}
+	if removed, err := pruneRotatedLogs(now); err != nil {
+		flog.Warn("logrotate: prune failed", "err", err)
+	} else if removed > 0 {
+		flog.Info("logrotate: pruned expired logs", "removed", removed)
+	}
+}
+
+// rotateLogIfDue cuts fleet.log into ~/.fleet/logs/fleetd/<date>.log when a
+// rotation is due, returning the dated path and whether it rotated. A rotation
+// is due once it is at or past logRotateHour local time and today's dated log
+// does not yet exist; that filesystem-derived "already cut today" test makes the
+// rotation idempotent and gives free catch-up across daemon restarts without any
+// in-memory bookkeeping. An empty or missing fleet.log is left alone so quiet
+// days produce no empty dated files.
+func rotateLogIfDue(now time.Time) (string, bool, error) {
+	if now.Hour() < logRotateHour {
+		return "", false, nil // not yet the rotation hour today
+	}
+
+	src := flog.Path()
+	info, err := os.Stat(src)
+	if os.IsNotExist(err) {
+		return "", false, nil // nothing logged yet
+	}
+	if err != nil {
+		return "", false, err
+	}
+	if !info.Mode().IsRegular() || info.Size() == 0 {
+		return "", false, nil // nothing to rotate
+	}
+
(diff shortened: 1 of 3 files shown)


internal/server/logrotate.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	// Cut at the start of the day logRotateKeepDays days back: a log dated before
+	// it is more than keepDays days old and is removed; today's just-written log
+	// (and the keepDays days behind it) is kept.
+	cutoff := startOfDay(now).AddDate(0, 0, -logRotateKeepDays)
+	removed := 0
+	for _, e := range entries {
```
BenjaminBenetti, line 242: 🔍 **CodeReview Agent**

`pruneRotatedLogs` never reclaims crash-orphaned temp files. `copyTruncate` writes `.rotate-*.log.tmp` into this same `fleetdLogDir()`, and if the daemon dies between `os.CreateTemp` and the `os.Rename`, the deferred `os.Remove` never runs — leaving the temp behind. Because this loop only deletes entries matching `<date>.log`, those orphans accumulate indefinitely.

The sibling `pruneBackups` handles exactly this case (backup.go:298–306):

```go
if strings.HasPrefix(name, ".backup-") && strings.HasSuffix(name, ".tmp") {
    _ = os.Remove(filepath.Join(dayPath, name))
    continue
}
```

Consider adding the analogous sweep here (e.g. reclaim `.rotate-*.log.tmp`) so the two loops stay consistent and a crashed rotation can't leak temp files into `~/.fleet/logs/fleetd`.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the daily `fleet.log` rotation loop. This is a clean, well-reasoned change that mirrors the established `backup.go` patterns closely (immediate-tick loop, `*Tick` split for testability, filesystem-derived idempotency, temp-file-then-rename writes, prune-by-filename-date). The header comment justifying copytruncate-over-rename is excellent, and the test suite is genuinely thorough — the shared-`O_APPEND`-fd test (`TestRotatePreservesSharedAppendFd`) pins exactly the guarantee a rename would break. Nice work.

A few findings, none blocking:

**1. Orphaned `.rotate-*.log.tmp` temp files are never reclaimed (low–medium).** `copyTruncate` creates its temp in `fleetdLogDir()` via `os.CreateTemp` and relies on a `defer os.Remove` for cleanup. If the daemon is killed (SIGKILL / power loss) between `CreateTemp` and the `Rename`, that temp persists, and `pruneRotatedLogs` deliberately skips anything that isn't `<date>.log`, so it accumulates forever. This is a direct divergence from the sibling `pruneBackups`, which *explicitly* reclaims crash-orphaned `.backup-*.tmp` files (backup.go:298–306) for exactly this reason — and the file header claims parity with the backup loop's approach. Blast radius is small (at most one orphan per crashed rotation), but adding the same reclamation sweep to `pruneRotatedLogs` would close the gap and keep the two loops consistent. Inline comment below.

**2. `logRotateHour` has no upper-bound validation (low).** `envIntDefault` only rejects negatives, so `FLEET_LOG_ROTATE_HOUR=25` parses to `25`; then `now.Hour() < 25` is always true and the "before the rotation hour" gate never fires, making rotation effectively due at 00:00. It's a test-only seam and the daemon inherits a trusted environment, so severity is low — but a `0–23` clamp (or a comment that the knob is unvalidated-by-design) would prevent a silent surprise.

**3. "keep 100 days" actually retains 101 daily buckets (nit/doc).** With `keepDays=100`, `cutoff = startOfDay(now) - 100d` and the `day.Before(cutoff)` test keeps `now-100d` through `now` inclusive — 101 files. This is intentional and asserted by `TestPruneRotatedLogs` (keep100 kept, drop101 removed); just flagging that the PR title / retention wording is off by one versus the on-disk count. No change needed if the inclusive boundary is the intent.

The copytruncate data-loss window (records appended between the copy reaching EOF and the truncate) is correctly identified as acceptable given `flog`'s best-effort contract, and the copy-before-truncate ordering correctly ensures a mid-rotation failure never loses data. No concerns there.

BenjaminBenetti:
Thanks for the review — addressed all three in da1ee08:

1. **Orphaned temp files** — `pruneRotatedLogs` now sweeps crash-orphaned `.rotate-*.log.tmp` files, exactly mirroring `pruneBackups` (the `HasPrefix(".rotate-") && HasSuffix(".log.tmp")` continue-without-counting). Closes the parity gap the header claimed.
2. **Unbounded hour** — `logRotateHour` is now `min(max(…,0),23)`, so a fat-fingered override can't make the gate unsatisfiable.
3. **Off-by-one retention** — tightened the window to keep exactly `keepDays` daily logs (today inclusive): `cutoff = startOfDay(now).AddDate(0,0,-(keepDays-1))`. "Keep 100 days" is now 100 files on disk, and `TestPruneRotatedLogs` was updated accordingly (the `now-100` bucket now drops, plus a new orphaned-temp reclamation assertion).

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed after `da1ee08`. All three findings from the previous review are cleanly resolved, and I confirmed the change **builds**, **`go vet`s clean**, and the full `logrotate` test suite (11 tests) **passes** on the PR head.

**Verified fixes**

- ✅ **Orphaned temp reclaim** — `pruneRotatedLogs` now sweeps `.rotate-*.log.tmp` (logrotate.go:251) with the same `HasPrefix/HasSuffix … continue` shape as `pruneBackups`, and `TestPruneRotatedLogs` asserts the orphan is reclaimed *without* inflating the `removed` count. The parity gap the header comment claimed is now real.
- ✅ **Hour clamp** — `logRotateHour = min(max(…,0),23)` (logrotate.go:49) makes the "before the rotation hour" gate impossible to disable via a fat-fingered `FLEET_LOG_ROTATE_HOUR`.
- ✅ **Exact retention** — `cutoff = startOfDay(now).AddDate(0,0,-(keepDays-1))` (logrotate.go:238) now retains exactly `keepDays` buckets (today inclusive); `TestPruneRotatedLogs` was updated so `now-99` is the oldest kept and `now-100` drops. Title/wording and on-disk count now agree.

I also re-confirmed the load-bearing premise: `internal/flog` is the *only* writer of `~/.fleet/fleet.log` and opens it exactly once with `O_APPEND` and no reopen path, so copytruncate (not rename) is the correct choice and `TestRotatePreservesSharedAppendFd` pins the guarantee.

**One small follow-up (low severity, optional)** — the retention tightening introduced a new degenerate case at `keepDays == 0` that the old off-by-one window happened to mask. Details inline on the `logRotateKeepDays` declaration.

Nothing blocking — clean, well-tested iteration.

internal/server/logrotate.go
```diff
@@ -0,0 +1,272 @@
...
+	// logRotateHour is the local hour-of-day (0–23) at which fleet.log is cut.
+	// The spec is 3am; overridable via FLEET_LOG_ROTATE_HOUR so a test can target
+	// the current hour. Clamped to a real hour so a fat-fingered override (e.g.
+	// 25) can't make the "before the rotation hour" gate unsatisfiable and turn
+	// every check into a rotation.
+	logRotateHour = min(max(envIntDefault("FLEET_LOG_ROTATE_HOUR", 3), 0), 23)
+	// logRotateKeepDays is how many of the most recent daily logs to keep (today
+	// inclusive). After each rotation check, logs older than that window are
+	// pruned. Overridable via FLEET_LOG_ROTATE_KEEP_DAYS for tests.
+	logRotateKeepDays = envIntDefault("FLEET_LOG_ROTATE_KEEP_DAYS", 100)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Minor / optional — same footgun class as the hour clamp two lines up. `envIntDefault` rejects negatives (→ default 100) but **accepts `0`** (`TestEnvIntDefault` treats `0` as valid), and the tightened cutoff at line 238 is `startOfDay(now) - (keepDays-1)` days. With `keepDays == 0` that cutoff becomes **tomorrow midnight**, so `day.Before(cutoff)` is true for *today's* bucket — on the same tick, the rotation would delete the dated log it just created.

This is newly reachable after the retention fix: the old `- keepDays` window left today's log in place at `keepDays == 0`. Since you already clamp `logRotateHour` to guard the analogous override mistake, `max(envIntDefault("FLEET_LOG_ROTATE_KEEP_DAYS", 100), 1)` would restore symmetry. Low severity (override-only seam, `0` is nonsensical in production) — purely your call.

BenjaminBenetti:
Addressed the `keepDays == 0` follow-up in 103d7b8: floored `logRotateKeepDays` at 1 (`max(envIntDefault(…),1)`), so the cutoff can never reach today and prune the log just rotated — the minimum window keeps today alone. Added `TestPruneRotatedLogsKeepsTodayAtMinWindow` as a boundary guard.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behavior-level QA of the daily `fleet.log` rotation loop. I built the real `fleet` binary and ran the **actual `fleet server` daemon** inside a throwaway `golang:1.25-bookworm` Docker container, driving rotation fast via the PR's documented env seams (`FLEET_LOG_ROTATE_HOUR` / `_KEEP_DAYS` / `_INTERVAL`) and observing the real `~/.fleet` filesystem effects (no unit tests were run). Container date: `2026-06-24`, hour `03` (UTC).

| Test | Result | Notes |
|------|--------|-------|
| Daemon build & startup | ✅ Pass | Builds clean (`-buildvcs=false`); daemon comes up, no panics/fatal in any run |
| Rotation: copy → dated log + truncate live log | ✅ Pass | Prior content lands in `logs/fleetd/2026-06-24.log`; `fleet.log` truncated |
| Dated-log permissions preserved (0644) | ✅ Pass | Rotated copy is `-rw-r--r--`, matching `fleet.log` |
| Shared `O_APPEND` fd continues into truncated log | ✅ Pass | Daemon's own "rotated" marker lands in the *new* `fleet.log` (the copytruncate guarantee) |
| Same-day idempotency (no re-rotate / no clobber) | ✅ Pass | Dated-file md5 stable; live log not re-truncated; exactly 1 rotation |
| Before-hour no-op | ✅ Pass | `HOUR=23` @ hour 03 → no `fleetd` dir, log left intact |
| Pruning by date window (exact retention) | ✅ Pass | `KEEP_DAYS=3` → `06-21` pruned, `06-22/23/24` kept (boundary exact) |
| Orphaned temp reclaim | ✅ Pass | `.rotate-deadrun.log.tmp` removed during prune |
| Non-dated entries untouched | ✅ Pass | `random.log`, `notes.txt`, `subdir/` left alone |
| Empty / non-regular log left alone | ✅ Pass | No spurious dated file created (see note re: pure-empty branch) |
| Default config rotates at real 3am | ✅ Pass | No env seams — auto-started daemon rotated at real hour 03 |
| Client sanity (`fleet list`) | ✅ Pass | Auto-starts daemon, returns cleanly |

**Overall: ✅ PASS — 12 of 12 scenarios passing.**

---

## Detailed findings

### 1. Rotation: copy → dated log + truncate + 0644 + O_APPEND continuity (core)
Seeded `~/.fleet/fleet.log` with `PRESEED-LINE-1..3` **before** launching the daemon, then started `fleet server` with `FLEET_LOG_ROTATE_HOUR=0 FLEET_LOG_ROTATE_INTERVAL=2s`. After the first tick:

- `logs/fleetd/2026-06-24.log` created, **`-rw-r--r--` (0644)**, containing the preseed lines **plus** the daemon's own startup records (`mcp server started`, `fleet server started`).
- Live `fleet.log` was truncated, then received `logrotate: rotated fleet.log … path=…/2026-06-24.log` as its **first** record — i.e. the daemon's single long-lived `O_APPEND` descriptor kept writing into the truncated inode. This is exactly the guarantee a `rename` would break, and it held.
- The "rotated" marker is **absent** from the dated copy (written post-truncate), confirming copy-then-truncate ordering.

```
=== DATED FILE: logs/fleetd/2026-06-24.log ===
PRESEED-LINE-1 old event
PRESEED-LINE-2 old event
PRESEED-LINE-3 old event
time="… 03:39:33.364" level=INFO msg="mcp server started" …
time="… 03:39:33.364" level=INFO msg="fleet server started" … socket=…/fleet.sock
=== LIVE fleet.log (after rotation) -rw-r--r-- ===
time="… 03:39:33.365" level=INFO msg="logrotate: rotated fleet.log" … path=…/2026-06-24.log
```

### 2. Same-day idempotency
With the daemon still running, appended `POSTROTATE-EXTRA-1/2` to the live log and waited 2+ more ticks. Today's dated log already existed, so rotation correctly short-circuited: dated-file md5 unchanged (`30307f60…`, 307 B), live log **not** re-truncated (extra content retained), exactly **1** dated file and **1** `rotated` marker.

### 3. Before-hour no-op
Fresh home, `FLEET_LOG_ROTATE_HOUR=23` at hour 03. No `logs/fleetd/` directory was created and the seeded log was left fully intact — rotation correctly waits for the configured hour.

### 4. Pruning + orphan reclaim + non-dated safety
Seeded `logs/fleetd/` with dated logs `2026-01-01`, `2026-06-01`, `2026-06-21`, `2026-06-22`, `2026-06-23`, `2026-06-24`, an orphan `.rotate-deadrun.log.tmp`, plus `random.log`, `notes.txt`, and `subdir/`. Ran with `FLEET_LOG_ROTATE_KEEP_DAYS=3` (`HOUR=23` to isolate prune from rotation). Result:

- Kept: `2026-06-22/23/24.log` (exactly the 3-day window — `06-21` is the first day removed, confirming an **exact** boundary).
- Pruned: `2026-01-01`, `2026-06-01`, `2026-06-21`.
- `.rotate-deadrun.log.tmp` reclaimed.
- `random.log`, `notes.txt`, `subdir/` untouched.

### 5. Empty / non-regular log left alone
The pure "empty *regular* `fleet.log`" branch is not reproducible against the running daemon because the live daemon always emits startup records, so by tick time the log is non-empty and (correctly) rotates those. To exercise the leave-alone guard deterministically I forced the log to stay empty (made `fleet.log` non-regular so `flog` discards): the daemon started fine and created **no** `fleetd` dir / **no** spurious dated file — matching "quiet days produce no empty dated files." (The strict `size==0` regular-file case is additionally covered by the PR's unit tests.)

### 6. Default-config + sanity
- With **no** env seams, the daemon auto-started by `fleet list` rotated at the real current hour (03 ≥ default 3am): `2026-06-24.log` (0644) + rotated marker in the live log. The `~/.fleet` tree (backup/, logs/, mcp.*, sockets) looked healthy.
- No `panic` / `fatal` / nil-pointer output across any of the 6 daemon runs. `fleet list` returns cleanly (exit 0).

---

_Tested at commit `da1ee08` (head of `feat/fleetd-log-rotation`). All execution was isolated in Docker container `qa-209`, now torn down._

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed after `103d7b8` (the `synchronize` that floors `keepDays`). The prior round's single open follow-up — the `keepDays == 0` degenerate cutoff that the earlier off-by-one window had accidentally masked — is now cleanly resolved.

**Verified on the PR head (`103d7b8`)**
- ✅ `logRotateKeepDays = max(envIntDefault(…, 100), 1)` floors the retention window at 1, mirroring the `0–23` hour clamp. At the floor, `cutoff = startOfDay(now)` keeps today and prunes everything strictly before it, so the log just rotated can never be the thing pruned. The reasoning in the var comment matches the behavior.
- ✅ `TestPruneRotatedLogsKeepsTodayAtMinWindow` pins exactly that boundary (today kept, yesterday dropped) — a good regression guard for the degenerate case.
- ✅ Independently re-ran on the head: `go build` and `go vet ./internal/server/` are clean, and the full 11-test `logrotate` suite passes.

I also confirmed the testing shape matches the sibling loop's convention: like `backup_test.go` (which tests `writeBackup`/`pruneBackups`, not `backupTick`/`runBackupLoop`), this PR tests `rotateLogIfDue`/`pruneRotatedLogs` directly and leaves the thin tick/loop orchestration untested — consistent with the codebase, not a gap.

**One optional observation (no action needed)** — `copyTruncate` deliberately copies-then-truncates so a mid-rotation failure never loses data (correct, and noted last round). The mirror-image failure mode is worth one sentence: if `os.Rename` succeeds but the subsequent `os.Truncate(src, 0)` fails, today's dated log already exists, so every later tick short-circuits on "already rotated today" and `fleet.log` is never trimmed until tomorrow's cut — with that day's tail duplicated across both files. It self-heals the next day, and a truncate failing on a file you just renamed beside is near-impossible, so this is purely FYI; the copy-before-truncate ordering remains the right call.

Clean, well-tested iteration — nothing blocking.

BenjaminBenetti:
Re-review noted, thanks. The FYI on `os.Rename` succeeding but `os.Truncate` failing is a fair edge case: today's dated log would exist, so subsequent ticks short-circuit on "already rotated today" and `fleet.log` isn't trimmed until tomorrow's cut (that day's tail duplicated across both files), self-healing the next day. Since truncating a regular file you just renamed beside in the same dir is effectively impossible and the behavior self-heals, I'm leaving the copy-before-truncate ordering as-is rather than spinning another CI cycle for it. Capturing the reasoning here for the record.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & daemon launch (`fleet server`) | ✅ Pass | Binary builds; daemon starts and writes `~/.fleet/fleet.log` |
| Rotation when due (default 3am, no overrides) | ✅ Pass | Local hour was 3 → rotated on first tick into `logs/fleetd/2026-06-24.log` |
| Dated log content & perms | ✅ Pass | Holds pre-rotation records; mode `0644` (matches live log) |
| copytruncate — shared `O_APPEND` fd continuity | ✅ Pass | "rotated" self-marker lands in the **fresh** log, not the dated one |
| Idempotency (no double rotation same day) | ✅ Pass | New content appended post-rotation stays in live log; dated file untouched |
| Before rotation hour → no-op | ✅ Pass | `HOUR=23` (not yet due): no dated file, live log not truncated |
| Retention prune by date window | ✅ Pass | `KEEP_DAYS=2` removed 3 stale dated logs; kept today+yesterday |
| Orphan temp reclaim / non-log files untouched | ✅ Pass | `.rotate-*.log.tmp` removed; `notes.txt` & subdir left alone |
| `KEEP_DAYS=0` floored to 1 (latest fix commit) | ✅ Pass | Today's just-rotated log **survives**; older log pruned |
| Empty/missing `fleet.log` left alone | ✅ Pass | Empty log over several ticks → no empty dated file created |
| Local-time schedule & naming | ✅ Pass | `TZ=America/New_York` → dated log named `2026-06-23.log` (local), not UTC `06-24` |
| Graceful shutdown with loop running | ✅ Pass | SIGINT → "fleet server stopping (signal)" → clean exit |

**Overall: ✅ PASS — 12 of 12 scenarios passing**

All env test seams (`FLEET_LOG_ROTATE_HOUR`, `FLEET_LOG_ROTATE_KEEP_DAYS`, `FLEET_LOG_ROTATE_INTERVAL`) were exercised across these runs and behaved as documented.

---

### How this was tested
All execution was isolated in a throwaway `golang:1.25` Docker container (`qa-209`). The `fleet` binary was built from the PR branch (HEAD `103d7b8`) and the **real daemon** (`fleet server`) was driven via tmux inside the container, each scenario using its own isolated `HOME` so `~/.fleet` was clean. No automated unit/integration tests were run — every behavior below was observed by running the daemon and inspecting the live filesystem. Notably, the host's local hour was `03` (UTC), which is the **default** rotation hour, so the primary flow rotated with **no env overrides at all**.

### Detailed findings

**1–4. Rotation when due + dated content + perms + fd continuity (core feature).**
Launched the daemon with a clean `HOME`, no overrides. On the first tick it rotated. Observed:
- `~/.fleet/logs/fleetd/2026-06-24.log` created, mode `0644`, containing the daemon's pre-rotation startup records (`mcp server started`, `fleet server started`).
- The live `fleet.log` was truncated and its first record was `logrotate: rotated fleet.log` — i.e. the daemon's long-lived `O_APPEND` descriptor kept writing into the truncated inode.
- Discriminating check: `grep "logrotate: rotated"` → count `0` in the dated file, count `1` in the fresh `fleet.log`. A rename-based rotation would invert this (the marker would follow the old descriptor into the dated file), so this directly confirms the copytruncate guarantee called out in the PR description.

**5. Idempotency.** With `HOUR=0 INTERVAL=1s`, after the first rotation I appended `MARKER_AFTER_FIRST_ROTATE_xyz` to the live log and waited 3 more ticks. The dated log's size + md5 were unchanged, it did not contain the marker, the live log still did (not re-truncated), and exactly one dated log existed. "Already rotated today" is correctly filesystem-derived.

**6. Before rotation hour.** With `HOUR=23` (current local hour 3 → not yet due), waited 4 ticks: `logs/fleetd/` was never created, the live log retained all 3 startup records, and no `logrotate: rotated` record appeared.

**7. Retention prune.** Pre-seeded dated logs `2026-06-24/23/22/20` and `2026-03-16`, plus `notes.txt`, a subdir, and an orphaned `.rotate-deadbeef.log.tmp`. With `KEEP_DAYS=2`, after a tick the remaining files were exactly `2026-06-24.log`, `2026-06-23.log`, `notes.txt`, `somedir/` — the 3 out-of-window dated logs and the orphan temp were removed (`pruned expired logs removed=3`; the temp isn't counted in `removed`, matching the code). Non-dated entries and directories were untouched.

**8. `KEEP_DAYS=0` floored to 1.** Pre-seeded only an old `2026-06-20.log`; daemon rotated startup content into today's `2026-06-24.log` and pruned with `KEEP_DAYS=0`. Today's just-rotated log **survived** and `2026-06-20.log` was removed (`removed=1`). This validates the `fix(logrotate): floor keepDays at 1 so the rotated log survives` commit — without the floor, a `0` window would push the cutoff to tomorrow and delete the log it just created.

**9. Empty log left alone.** After an initial rotation I deleted today's dated log and truncated `fleet.log` to 0 bytes, then idled through several 2s ticks. The live log stayed at 0 bytes and no dated file was (re)created — quiet days produce no empty dated logs.

**10. Local time.** Ran with `TZ=America/New_York`; UTC `2026-06-24 03:53` is local `2026-06-23 23:53`. The dated log was named `2026-06-23.log` (local date), confirming both the hour gate and the filename use local time, not UTC.

**11. Graceful shutdown.** Sent SIGINT to a running daemon; it logged `fleet server stopping (signal)` and the process exited cleanly (no lingering `fleet server` process), so the new rotation goroutine launched on `hubCtx` does not block teardown.

No defects or regressions found.

Issue #88, opened 2026-06-05
mouse control of fleet TUI stops working


BenjaminBenetti:
Randomly after a bit the Fleet TUI mouse control stops working. 

I will note other programs open in the tmux session like nvim continue to have mouse control.

PR #210, opened 2026-06-24
feat(tui): stream fleetd logs from the settings page

BenjaminBenetti:
## What

Adds a **Logs** stream selector to the **Fleet Daemon** section of the Settings page:

```
> Logs              [All] [Error] [Warn] [Info]
```

- Default is **All**; `←/→` or vim `h/l` move the selection (clamped — it's a segmented control, not a wrap-around cycle).
- `Enter` / `Space` hands the terminal to a live `tail -f` of `~/.fleet/fleet.log`, filtered to the selected level **and above**.
- Ctrl-C kills the stream and returns to the TUI.

This is the same screen-takeover mechanic already used for the doctor and the `$EDITOR` prompt editors (reuses `execProcess`, which also carries the post-exec mouse-tracking fix from #88).

## Why

There was no convenient way to watch the fleetd event log while using fleet — you'd have to drop to a shell and `tail -f` it yourself.

## Details

- **Filtering** is a `grep -E` over slog's `level=` field. Ordering is ERROR < WARN < INFO, so Error = ERROR only, Warn = WARN+ERROR, Info = INFO+WARN+ERROR, All = no filter. The leading-space anchor keeps a stray `level=ERROR` inside a message value from matching. *(fleetd logs at `slog.LevelInfo`, so DEBUG never reaches the file and Info ≈ All in practice.)*
- **Local-only**, alongside the existing Restart row — the section is already hidden for remote TUIs (`FLEET_GATEWAY`/`FLEET_SERVER`), and a remote TUI can't read the daemon's local log file anyway.
- The selection is **ephemeral** (in-memory on the settings page; resets to All each session) — it's a transient view control, not persisted to config.

## Files

- `internal/tui/commands.go` — `daemonLogLevels` table + `daemonLogStreamCommand`.
- `internal/tui/page_settings.go` — new row constant, section wiring, `logLevel` state, `cycleDaemonLogLevel`, left/right + enter handlers, segmented-control render, footer hint.
- `internal/tui/page_settings_test.go` — local/remote visibility, cycle/clamp + vim keys, enter-streams, and command-builder filter tests.

## Test

- `go build ./...`, `go test ./internal/tui/...`, `go vet ./internal/...` all green.
- Manual (local TUI): Settings → Fleet Daemon → Logs renders `[All] [Error] [Warn] [Info]` with All highlighted; `←/→` and `h/l` move/clamp; Enter streams live; pick Error to see only ERROR lines; Ctrl-C returns cleanly.

### DIFF
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -285,8 +285,10 @@ type daemonLogLevel struct {
 // daemonLogLevels are the selectable minimum levels, in on-screen order: All,
 // Error, Warn, Info. fleetd writes slog text records ("level=ERROR" etc.), so
 // "and above" is just a wider alternation. The leading space anchors level= to
-// the field — it always follows the quoted time value — so a stray "level=ERROR"
-// inside a message value can't match.
+// the start of the field (it always follows the quoted time value), which makes a
+// stray "level=ERROR" inside a message value unlikely to match — a space-separated
+// token inside a value still could, but the worst case is one extra line in a
+// transient view, not worth a stricter parse.
 var daemonLogLevels = []daemonLogLevel{
 	{label: "All", grep: ""},
 	{label: "Error", grep: ` level=ERROR`},
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -12,7 +12,6 @@ import (
 	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
 	"github.com/BenjaminBenetti/fleet-man/internal/fleetclient"
 	"github.com/BenjaminBenetti/fleet-man/internal/fleetpaths"
-	"github.com/BenjaminBenetti/fleet-man/internal/flog"
 	"github.com/BenjaminBenetti/fleet-man/internal/portforward"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -301,7 +300,11 @@ var daemonLogLevels = []daemonLogLevel{
 // TUI. -F survives a log recreate, -n 200 shows recent context, and grep's
 // --line-buffered makes filtered lines appear promptly rather than block-buffered.
 func daemonLogStreamCommand(level daemonLogLevel) *exec.Cmd {
-	stream := fmt.Sprintf("tail -n 200 -F %q", flog.Path())
+	// ~/.fleet/fleet.log. Computed from fleetpaths rather than flog.Path() because
+	// the import-boundary lint bars clients (the TUI) from importing the
+	// server-owned flog package; the filename mirrors flog's logFileName.
+	path := filepath.Join(fleetpaths.Dir(), "fleet.log")
+	stream := fmt.Sprintf("tail -n 200 -F %q", path)
 	if level.grep != "" {
 		stream += fmt.Sprintf(" | grep --line-buffered -E '%s'", level.grep)
 	}
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -3,6 +3,7 @@ package tui
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os/exec"
 	"path/filepath"
@@ -312,6 +313,30 @@ func daemonLogStreamCommand(level daemonLogLevel) *exec.Cmd {
 	}
 	header := fmt.Sprintf("── fleet.log [%s] ─── Ctrl-C to return to fleet ───", level.label)
 	return exec.Command("sh", "-c", fmt.Sprintf("printf '%%s\\n\\n' %q; %s", header, stream))
+}
+
+// streamInterruptExitCode is the status a shell pipeline reports when killed by
+// SIGINT (128 + SIGINT) — i.e. the user pressed Ctrl-C to end the stream.
+const streamInterruptExitCode = 130
+
+// silenceStreamInterrupt nils out the expected Ctrl-C exit (status 130) of a
+// log-stream child so the shared execDoneMsg handler doesn't flash it in the
+// footer as "Command error: exit status 130"; any other failure passes through.
+func silenceStreamInterrupt(err error) error {
+	var ee *exec.ExitError
+	if errors.As(err, &ee) && ee.ExitCode() == streamInterruptExitCode {
+		return nil
+	}
+	return err
+}
+
+// daemonLogStreamCmd hands the terminal to a live fleet.log tail (see
+// daemonLogStreamCommand) and returns to the TUI when the user ends it. Ctrl-C —
+// the documented way to stop the stream — is treated as a clean exit.
+func daemonLogStreamCmd(level daemonLogLevel) tea.Cmd {
+	return execProcess(daemonLogStreamCommand(level), func(err error) tea.Msg {
+		return execDoneMsg{silenceStreamInterrupt(err)}
+	})
 }
 
 // createInstanceCmd dispatches a CreateInstance job to the server, which
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -12,6 +12,7 @@ import (
 	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
 	"github.com/BenjaminBenetti/fleet-man/internal/fleetclient"
 	"github.com/BenjaminBenetti/fleet-man/internal/fleetpaths"
+	"github.com/BenjaminBenetti/fleet-man/internal/flog"
 	"github.com/BenjaminBenetti/fleet-man/internal/portforward"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -272,6 +273,40 @@ func logsCommand(fleetName string, instance *fleet.Instance) *exec.Cmd {
 	// Wrap in a shell that pauses after the output.
 	script := fmt.Sprintf(`%s; echo; echo "--- Press Enter to return ---"; read _`, inner)
 	return exec.Command("sh", "-c", script)
+}
+
+// daemonLogLevel describes one entry in the settings "Logs" selector. grep is an
+// extended-regex matching that level and everything above it; an empty grep means
+// no filter (All).
+type daemonLogLevel struct {
+	label string
+	grep  string
+}
+
+// daemonLogLevels are the selectable minimum levels, in on-screen order: All,
+// Error, Warn, Info. fleetd writes slog text records ("level=ERROR" etc.), so
+// "and above" is just a wider alternation. The leading space anchors level= to
+// the field — it always follows the quoted time value — so a stray "level=ERROR"
+// inside a message value can't match.
+var daemonLogLevels = []daemonLogLevel{
+	{label: "All", grep: ""},
+	{label: "Error", grep: ` level=ERROR`},
+	{label: "Warn", grep: ` level=(ERROR|WARN)`},
+	{label: "Info", grep: ` level=(ERROR|WARN|INFO)`},
+}
+
+// daemonLogStreamCommand returns an *exec.Cmd that tails the fleetd event log
+// (~/.fleet/fleet.log) live, optionally filtered to a minimum level. It behaves
+// like `tail -f`: the user presses Ctrl-C to stop the stream and return to the
+// TUI. -F survives a log recreate, -n 200 shows recent context, and grep's
+// --line-buffered makes filtered lines appear promptly rather than block-buffered.
+func daemonLogStreamCommand(level daemonLogLevel) *exec.Cmd {
+	stream := fmt.Sprintf("tail -n 200 -F %q", flog.Path())
+	if level.grep != "" {
+		stream += fmt.Sprintf(" | grep --line-buffered -E '%s'", level.grep)
+	}
+	header := fmt.Sprintf("── fleet.log [%s] ─── Ctrl-C to return to fleet ───", level.label)
+	return exec.Command("sh", "-c", fmt.Sprintf("printf '%%s\\n\\n' %q; %s", header, stream))
 }
 
 // createInstanceCmd dispatches a CreateInstance job to the server, which
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
@@ -315,17 +316,26 @@ func daemonLogStreamCommand(level daemonLogLevel) *exec.Cmd {
 	return exec.Command("sh", "-c", fmt.Sprintf("printf '%%s\\n\\n' %q; %s", header, stream))
 }
 
-// streamInterruptExitCode is the status a shell pipeline reports when killed by
-// SIGINT (128 + SIGINT) — i.e. the user pressed Ctrl-C to end the stream.
-const streamInterruptExitCode = 130
+// streamInterruptExitCode is the status some shells report when a pipeline is
+// killed by SIGINT (128 + SIGINT).
+const streamInterruptExitCode = 128 + int(syscall.SIGINT) // 130
 
-// silenceStreamInterrupt nils out the expected Ctrl-C exit (status 130) of a
-// log-stream child so the shared execDoneMsg handler doesn't flash it in the
-// footer as "Command error: exit status 130"; any other failure passes through.
+// silenceStreamInterrupt nils out the expected Ctrl-C termination of a log-stream
+// child so the shared execDoneMsg handler doesn't flash it in the footer as
+// "Command error: …"; any other failure passes through. Ctrl-C in a terminal
+// delivers SIGINT to the whole foreground group, and the parent sh surfaces that
+// two different ways: some shells exit with 128+SIGINT (130), but dash/bash are
+// themselves terminated by the signal, which os/exec reports as a signal death
+// (ExitCode() == -1) — both are the documented exit, so both are silenced.
 func silenceStreamInterrupt(err error) error {
 	var ee *exec.ExitError
-	if errors.As(err, &ee) && ee.ExitCode() == streamInterruptExitCode {
-		return nil
+	if errors.As(err, &ee) {
+		if ee.ExitCode() == streamInterruptExitCode {
+			return nil
+		}
+		if ws, ok := ee.Sys().(syscall.WaitStatus); ok && ws.Signaled() && ws.Signal() == syscall.SIGINT {
+			return nil
+		}
 	}
 	return err
 }
--- a/internal/tui/page_settings.go
+++ b/internal/tui/page_settings.go
@@ -64,6 +64,7 @@ const (
 	settingsItemArmadaBase = 100000
 
 	settingsItemDaemonRestart = 900 // "Restart daemon" action row (below the tool-status band)
+	settingsItemDaemonLogs    = 901 // "Logs" level selector + fleet.log stream launcher
 
 	settingsItemToolStatusBase = 1000 // tool status rows start here
 	settingsItemDoctor         = 2000 // doctor action row
@@ -105,6 +106,11 @@ type settingsPage struct {
 	editing         bool
 	input           textinput.Model
 	showKeybindings bool
+
+	// logLevel is the index into daemonLogLevels selected on the Fleet Daemon
+	// "Logs" row. Ephemeral (resets to 0 = All each session); enter streams
+	// ~/.fleet/fleet.log filtered to that level and above.
+	logLevel int
 
 	// itemRowYs maps item ID -> terminal Y where the item's first line
 	// is rendered. itemHeights maps item ID -> number of lines the item
@@ -356,12 +362,13 @@ var settingsSections = []settingsSection{
 	},
 	{
 		// Only meaningful for a local TUI: the restart relaunches the LOCAL
-		// daemon from the current binary. A remote TUI (FLEET_GATEWAY/
-		// FLEET_SERVER) can't relaunch the daemon it talks to, so hide it.
+		// daemon from the current binary, and the log stream tails the LOCAL
+		// ~/.fleet/fleet.log off disk. A remote TUI (FLEET_GATEWAY/FLEET_SERVER)
+		// can't do either for the daemon it talks to, so hide the section.
 		Title:   "Fleet Daemon",
 		Visible: func(_ *model) bool { return !fleetclient.IsRemote() },
 		Items: func(_ *model) []int {
-			return []int{settingsItemDaemonRestart}
+			return []int{settingsItemDaemonLogs, settingsItemDaemonRestart}
 		},
 	},
 	{
@@ -668,6 +675,14 @@ func (settingsPage *settingsPage) cycleCoderPreset(m *model, direction int) {
 	m.message = fmt.Sprintf("Preset set to %s", m.config.CoderSettings.Preset)
 }
 
+// cycleDaemonLogLevel moves the Fleet Daemon "Logs" selection by direction,
+// clamped at the ends. It's a segmented control (All…Info shown at once), so it
+// clamps rather than wrapping, and the choice is purely in-memory — nothing is
+// persisted; enter launches the stream at the selected level.
+func (settingsPage *settingsPage) cycleDaemonLogLevel(direction int) {
+	settingsPage.logLevel = max(0, min(settingsPage.logLevel+direction, len(daemonLogLevels)-1))
+}
+
 // cycleCodespacesMachine cycles through available codespace machine types.
 func (settingsPage *settingsPage) cycleCodespacesMachine(m *model, direction int) {
 	if m.config == nil || len(m.codespaceMachines) == 0 {
@@ -935,6 +950,8 @@ func (settingsPage *settingsPage) updateSettingsNav(m *model, msg tea.Msg) tea.C
 				settingsPage.cycleCoderPreset(m, -1)
 			} else if item == settingsItemCodespacesMachine {
 				settingsPage.cycleCodespacesMachine(m, -1)
+			} else if item == settingsItemDaemonLogs {
+				settingsPage.cycleDaemonLogLevel(-1)
 			}
 			return nil
 
@@ -965,6 +982,8 @@ func (settingsPage *settingsPage) updateSettingsNav(m *model, msg tea.Msg) tea.C
 				settingsPage.cycleCoderPreset(m, 1)
 			} else if item == settingsItemCodespacesMachine {
 				settingsPage.cycleCodespacesMachine(m, 1)
+			} else if item == settingsItemDaemonLogs {
+				settingsPage.cycleDaemonLogLevel(1)
 			}
 			return nil
 
@@ -1078,6 +1097,14 @@ func (settingsPage *settingsPage) updateSettingsNav(m *model, msg tea.Msg) tea.C
 			if item == settingsItemKeybindings {
 				settingsPage.showKeybindings = true
 				return nil
+			}
+			if item == settingsItemDaemonLogs {
+				// Hand the terminal to a live `tail -f` of fleet.log filtered to the
+				// selected level (same screen-takeover mechanic as the doctor row).
+				return execProcess(
+					daemonLogStreamCommand(daemonLogLevels[settingsPage.logLevel]),
+					func(err error) tea.Msg { return execDoneMsg{err} },
+				)
 			}
 			if item == settingsItemDaemonRestart {
 				if m.daemonRestarting {
@@ -1695,6 +1722,23 @@ func (settingsPage *settingsPage) viewSettings(m *model) string {
 			}
 
 		case "Fleet Daemon":
+			// Logs level selector: every level bracketed, the active one
+			// highlighted. Enter on this row streams fleet.log at that level.
+			var seg strings.Builder
+			for i, lvl := range daemonLogLevels {
+				if i > 0 {
+					seg.WriteString(" ")
+				}
+				label := "[" + lvl.label + "]"
+				if i == settingsPage.logLevel {
+					seg.WriteString(selectedStyle.Render(label))
+				} else {
+					seg.WriteString(dimStyle.Render(label))
+				}
+			}
+			recordRow(settingsItemDaemonLogs, settingsPage.renderSettingsRow(m, currentItem == settingsItemDaemonLogs, "Logs", seg.String()))
+			listContent.WriteString("\n")
+
 			var daemonValue string
 			if m.daemonRestarting {
 				daemonValue = m.spinner.View() + " restarting…"
@@ -1749,6 +1793,12 @@ func (settingsPage *settingsPage) viewSettings(m *model) string {
 	// footer's "enter: edit / left/right: cycle" doesn't apply to them.
 	if isCopyRow(currentItem) {
 		tail.WriteString(dimStyle.Render("  enter: copy to clipboard"))
+		tail.WriteString("\n")
+	}
+	// The Logs row streams on enter (not edit/cycle), so spell it out instead of
+	// the generic "enter: edit" footer.
+	if currentItem == settingsItemDaemonLogs {
+		tail.WriteString(dimStyle.Render("  left/right: choose level  enter: stream fleet.log (Ctrl-C to return)"))
 		tail.WriteString("\n")
 	}
 	if m.message != "" {
--- a/internal/tui/page_settings.go
+++ b/internal/tui/page_settings.go
@@ -1101,10 +1101,7 @@ func (settingsPage *settingsPage) updateSettingsNav(m *model, msg tea.Msg) tea.C
 			if item == settingsItemDaemonLogs {
 				// Hand the terminal to a live `tail -f` of fleet.log filtered to the
 				// selected level (same screen-takeover mechanic as the doctor row).
-				return execProcess(
-					daemonLogStreamCommand(daemonLogLevels[settingsPage.logLevel]),
-					func(err error) tea.Msg { return execDoneMsg{err} },
-				)
+				return daemonLogStreamCmd(daemonLogLevels[settingsPage.logLevel])
 			}
 			if item == settingsItemDaemonRestart {
 				if m.daemonRestarting {
--- a/internal/tui/page_settings_test.go
+++ b/internal/tui/page_settings_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/BenjaminBenetti/fleet-man/fleetgrpc"
+	"github.com/BenjaminBenetti/fleet-man/internal/flog"
 	"github.com/BenjaminBenetti/fleet-man/internal/state"
 	"github.com/charmbracelet/bubbles/spinner"
 	"github.com/charmbracelet/bubbles/textinput"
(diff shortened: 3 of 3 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the diff for #210 (fleetd log stream selector on the Settings page). This is a tidy, well-scoped change that closely follows existing conventions — `execProcess` screen-takeover reuse, the `max(0, min(...))` clamp idiom already used elsewhere in this file, `%q` path quoting matching `logsCommand`, and the section already being hidden for remote TUIs. Both `enter` and `space` route to the stream handler, vim `h/l` are wired alongside `←/→`, and the tests cover local/remote visibility, cycle/clamp, vim keys, enter-streams, and the command builder. `go build`, `go vet ./internal/...`, and `go test ./internal/tui/...` are green locally.

No blocking issues. A few minor, non-blocking observations:

**1. The "leading-space anchor" doesn't fully prevent false matches (comment overstates it).** The doc comment claims a stray `level=ERROR` inside a message value "can't match" because of the leading space. The space only rules out the case where `level=` is preceded by a non-space char. A value containing a *space-separated* token still matches — e.g. a container-exec record `cmd="mytool --foo level=ERROR"` would be surfaced by the Error filter even though the record's real level is INFO. Worst case is a spurious extra line in a transient view, so this is cosmetic, but consider softening the comment (e.g. "makes a stray match unlikely" rather than "can't match"). Anchoring to the start of the field with `grep -E ' level=(ERROR|WARN)'` already does the right thing for the common case; a stricter anchor isn't worth the complexity here.

**2. `%q` is Go-quoting, not shell-quoting, for `flog.Path()`.** Inside `sh -c "... %q ..."`, Go's `%q` won't escape `$`, backtick, or `\`, so a `$HOME` containing those would misbehave. In practice `$HOME` is trusted and this matches the existing `logsCommand` convention, so it's fine as-is — just flagging for awareness since it's a `sh -c` string built from an env-derived path.

**3. (optional) Test could assert a bit more of the command shape.** `TestDaemonLogStreamCommand` checks the path and the grep filter, but not the `-F` / `-n 200` / `--line-buffered` / header pieces called out in the doc comment as load-bearing. Cheap to add an assertion or two if you want the test to guard against regressions there.

**Minor UX note (no action needed):** before fleetd has ever written `~/.fleet/fleet.log`, `tail -F` will briefly print a "cannot open … for reading" notice before the file appears — expected `-F` behavior, just flagging in case you want to pre-create the file or note it.

Nice feature — the ephemeral, clamped segmented-control design and the local-only gating are the right calls.

internal/tui/commands.go
```diff
@@ -274,6 +275,40 @@ func logsCommand(fleetName string, instance *fleet.Instance) *exec.Cmd {
...
+type daemonLogLevel struct {
+	label string
+	grep  string
+}
+
+// daemonLogLevels are the selectable minimum levels, in on-screen order: All,
+// Error, Warn, Info. fleetd writes slog text records ("level=ERROR" etc.), so
+// "and above" is just a wider alternation. The leading space anchors level= to
+// the field — it always follows the quoted time value — so a stray "level=ERROR"
+// inside a message value can't match.
```
BenjaminBenetti: 🔍 **CodeReview Agent**

The leading space anchors `level=` to the *start* of the field, but it doesn't guarantee that a `level=` substring elsewhere can't match — a space-separated token inside a value matches too. e.g. a container-exec record `cmd="mytool --foo level=ERROR"` (real level INFO) would still be surfaced by the Error filter. Harmless (one spurious line in a transient view), but "can't match" overstates it — consider "makes a stray match unlikely".

internal/tui/commands.go
```diff
@@ -274,6 +274,44 @@ func logsCommand(fleetName string, instance *fleet.Instance) *exec.Cmd {
...
+// daemonLogStreamCommand returns an *exec.Cmd that tails the fleetd event log
+// (~/.fleet/fleet.log) live, optionally filtered to a minimum level. It behaves
+// like `tail -f`: the user presses Ctrl-C to stop the stream and return to the
+// TUI. -F survives a log recreate, -n 200 shows recent context, and grep's
+// --line-buffered makes filtered lines appear promptly rather than block-buffered.
+func daemonLogStreamCommand(level daemonLogLevel) *exec.Cmd {
+	// ~/.fleet/fleet.log. Computed from fleetpaths rather than flog.Path() because
+	// the import-boundary lint bars clients (the TUI) from importing the
+	// server-owned flog package; the filename mirrors flog's logFileName.
+	path := filepath.Join(fleetpaths.Dir(), "fleet.log")
```
BenjaminBenetti, line 310: 🔍 **CodeReview Agent**

This hardcoded `"fleet.log"` is now the TUI's independent copy of flog's `logFileName` (`internal/flog/flog.go`), kept in sync only by the comment above. The production change away from `flog.Path()` is the right call (depguard bars the TUI from importing `internal/flog`), but the matching test was *also* switched to re-derive this same `filepath.Join(fleetpaths.Dir(), "fleet.log")` expression — so it can no longer catch a divergence: rename `logFileName` and fleetd writes elsewhere while this keeps tailing the stale name, with `TestDaemonLogStreamCommand` still green.

Since the depguard `client` rule exempts `*_test.go` (`!**/*_test.go`), the test may import `flog`. Consider pinning the test's expected path to `flog.Path()`, or adding a tiny guard test (`flog.Path() == filepath.Join(fleetpaths.Dir(), "fleet.log")`), so a future rename fails loudly.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed after the new commit `fix(tui): compute fleet.log path via fleetpaths, not flog`.

**The fix itself is correct.** `filepath.Join(fleetpaths.Dir(), "fleet.log")` is byte-for-byte equivalent to `flog.Path()` today (`fleetpaths.Dir()` → `~/.fleet`, and `flog.logFileName` is `"fleet.log"`), it satisfies the depguard client/server import boundary (`internal/flog` is denied for `**/internal/tui/**` in `.golangci.yml`), and it mirrors the existing `logsCommand` pattern in this same file (`filepath.Join(fleetpaths.Dir(), "logs", …)`). Good, convention-matching fix.

One **new** observation introduced by this commit (non-blocking):

**The test no longer pins the TUI's hardcoded `"fleet.log"` to flog's canonical filename.** The previous test asserted against `flog.Path()`; the new test re-derives the *same* `filepath.Join(fleetpaths.Dir(), "fleet.log")` expression the implementation uses. So the test now only checks that the command contains a string the implementation built from its own literal — it can no longer catch drift. If flog's `logFileName` (`internal/flog/flog.go`) is ever renamed, fleetd would write to the new file, the TUI would silently keep tailing the stale `"fleet.log"`, and `TestDaemonLogStreamCommand` would still pass. The only thing keeping them aligned is the `// the filename mirrors flog's logFileName` comment.

Note that the depguard boundary explicitly exempts `*_test.go` (`- "!**/*_test.go"` in the `client` rule), so — unlike the production code — the **test is still allowed to import `flog`**. Cheapest fix that restores the guard: have the test compute the expected path from `flog.Path()` (revert just the test's source of truth), or add a one-line guard test asserting `flog.Path() == filepath.Join(fleetpaths.Dir(), "fleet.log")`. That makes a future rename fail loudly instead of silently breaking the stream.

Everything flagged in the prior review still stands as-is; nothing there regressed. Nice, tightly-scoped change overall.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` binary | ✅ Pass | `go build ./cmd/fleet` clean; 46 MB binary, `--help` works |
| Launch TUI & open Settings | ✅ Pass | No errors; Settings page renders all sections |
| Logs row renders segmented control | ✅ Pass | `Logs   [All] [Error] [Warn] [Info]`, **All** highlighted by default |
| ←/→ move selection | ✅ Pass | Right walks All → Error → Warn → Info |
| vim `h`/`l` move selection | ✅ Pass | `l` steps right, `h` steps left |
| Clamp at both ends (no wrap) | ✅ Pass | Right past Info stays Info; Left past All stays All |
| Footer hint on Logs row | ✅ Pass | `left/right: choose level  enter: stream fleet.log (Ctrl-C to return)` |
| Enter → screen-takeover stream | ✅ Pass | Header `── fleet.log [Error] ─── Ctrl-C to return to fleet ───` renders |
| Live `tail -F` picks up new lines | ✅ Pass | Appended ERROR line appeared live; filtered-out INFO line stayed hidden |
| All filter (no grep) shows everything | ✅ Pass | INFO/WARN/ERROR all shown, incl. real fleetd records |
| Error/Warn filter excludes lower levels | ⚠️ Partial | Real `INFO` lines correctly hidden — **but** a record whose *attribute value* contains a space‑prefixed ` level=ERROR` is a false positive (see below) |
| Ctrl-C returns to TUI | ✅ Pass | Clean return; cursor back on Logs row, screen restored |
| Selection is ephemeral within session | ✅ Pass | Stayed on Error after returning from the stream |
| Remote TUI hides Fleet Daemon / Logs | ⚠️ Partial | Verified by code inspection — could not be driven live (see below) |

**Overall: ✅ PASS — all primary acceptance criteria met; 2 minor caveats noted (no blocking defects).**

---

### Environment
Tested manually inside a throwaway Docker container (`golang:1.25-bookworm` + tmux), workspace mounted, PR branch `feat/tui-daemon-log-stream` @ `99c4bbf`. The `fleet` TUI was driven exactly as a user would (tmux send-keys / capture-pane). A representative `~/.fleet/fleet.log` (slog text format) was used, and the running daemon also wrote its own real records during the session — those filtered correctly too.

### ✅ What works (detailed)
- **Rendering & default state.** Settings → Fleet Daemon → `Logs   [All] [Error] [Warn] [Info]`. ANSI capture confirms `[All]` is bold/highlighted and the rest dim on launch.
- **Selection movement + clamping.** `→`/`l` advance All→Error→Warn→Info and **clamp** at Info (no wrap); `←`/`h` step back and **clamp** at All. Segmented‑control behavior is exactly as described.
- **Streaming.** `Enter` hands the terminal to the stream with the per‑level header (`── fleet.log [<level>] ─── Ctrl-C to return to fleet ───`). `tail -F` is genuinely live: a line appended to `fleet.log` after the stream started appeared immediately, and a freshly‑appended INFO line was correctly suppressed under the Error filter.
- **Per‑level filtering (realistic logs).** All = full file; Error = ERROR only; Warn = WARN+ERROR; Info = everything (no real DEBUG present). Every *real* fleetd record (`tui started`, `instance created`, `mcp server started`, `backup: wrote snapshot`, …) was filtered to the correct bucket.
- **Ctrl-C return.** Kills the stream and restores the TUI cleanly; cursor lands back on the Logs row, selection preserved (ephemeral, in‑memory).

### ⚠️ Caveat 1 — level filter false‑positive on values containing ` level=<LVL>`
The PR description states *"Error = ERROR only"* and that the leading‑space anchor *"keeps a stray `level=ERROR` inside a message value from matching."* That guard only stops the **no‑space / word‑boundary** case (e.g. `note=mylevel=ERROR` → correctly **not** matched ✅). A value that contains a **space‑prefixed** ` level=ERROR` still leaks through, because `grep -E ' level=ERROR'` matches it.

**Repro:** with this line in `fleet.log`
```
time="2026-06-24 03:40:05.500" level=INFO msg="container exec" pid=102 backend=docker cmd="echo level=ERROR sneaky"
```
selecting **Error** (or **Warn**) shows this `INFO` line even though its real level is INFO — the `cmd=` value's ` level=ERROR` substring matches. Observed live in the TUI, not just in a standalone grep.

**Severity: low.** It requires a log record whose *attribute value* embeds the literal text ` level=<LEVEL>` — fleetd's own log calls never do this; in practice only a `container exec` whose command string contains that exact text (or similar) would trigger it. All ordinary fleetd lines filter correctly. Worth a note for the author in case tightening is desired (e.g. anchor the field to the start of the record / match on the actual slog field rather than a substring). Not treated as blocking.

### ⚠️ Caveat 2 — remote‑hide verified by inspection only
The acceptance criterion *"hidden for remote TUIs (`FLEET_GATEWAY`/`FLEET_SERVER`)"* could not be exercised through the live TUI: launching with `FLEET_SERVER` pointed at an unreachable address makes the TUI panic during `tui.Run()` init (nil‑pointer at `app.go:1368`) before any page renders. **This panic is pre‑existing and unrelated to this PR** — the diff only touches `internal/tui/commands.go` and `internal/tui/page_settings.go` (+ the test), not the connection/init path. The Logs row is appended to the existing **Fleet Daemon** section whose gate is `Visible: !fleetclient.IsRemote()` — the same gate that already hid the Restart row — so the row is hidden in remote mode by construction. Verified by code inspection; flagged as Partial only because it wasn't driven end‑to‑end live.

### Notes
- `go build ./cmd/fleet` compiles cleanly (the only hiccup was VCS stamping on the mounted `.git`, resolved with `-buildvcs=false` — a mount artifact, not a code issue).
- No regressions observed in the surrounding Settings page: navigation, the existing Restart daemon row, `esc` back to the fleet list, and `q`/Ctrl-C quit all behaved normally.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manually exercised the new **Settings → Fleet Daemon → Logs** stream selector by building the binary and driving the real TUI inside an isolated Docker container (`debian:bookworm-slim`, static `fleet` binary, sample `~/.fleet/fleet.log`). Tested via tmux send-keys / capture-pane — no automated tests run.

| Test | Result | Notes |
|------|--------|-------|
| Build & TUI launch | ✅ Pass | `go build ./...` green; TUI starts, Settings opens |
| Logs row renders in Fleet Daemon | ✅ Pass | Shows `[All] [Error] [Warn] [Info]`, **All** highlighted by default |
| Footer hint on Logs row | ✅ Pass | `left/right: choose level  enter: stream fleet.log (Ctrl-C to return)` |
| Right move (→ and vim `l`) | ✅ Pass | All→Error→Warn→Info; **clamps** at Info (no wrap) |
| Left move (← and vim `h`) | ✅ Pass | Info→…→All; **clamps** at All (no wrap) |
| Enter streams (All) | ✅ Pass | Header `── fleet.log [All] ─── Ctrl-C to return to fleet ───`, full log shown |
| Space streams (alt binding) | ✅ Pass | `case "enter", " "` — Space launches the stream too |
| Live `tail -F` | ✅ Pass | New lines appended by the running fleetd appeared live in the stream |
| Error filter | ✅ Pass | Only `level=ERROR` lines shown |
| Warn filter | ✅ Pass | `WARN` + `ERROR` shown, no `INFO` |
| Info filter | ✅ Pass | `INFO`+`WARN`+`ERROR` shown; a live `level=DEBUG` line correctly **excluded** |
| Leading-space anchor | ✅ Pass | An `INFO` line whose msg value starts with `level=ERROR` is **not** matched by the Error filter |
| Ctrl-C returns to TUI | ✅ Pass | Stream stops, Settings page restored with Logs row still selected |
| Adjacent Restart daemon + back-nav | ✅ Pass | Restart row unaffected; Esc returns to Armada page; `q` quits cleanly (exit 0) |
| Remote-hidden (local-only) | ⏭️ Skipped | Not run against a remote gateway; section is gated by the unchanged `!fleetclient.IsRemote()` visibility func and covered by `TestDaemonSectionLocalOnly` |

**Overall: ✅ PASS — 13 of 13 executed scenarios passing (1 skipped, covered by existing gate/test).**

---

### Detailed findings

**Setup.** Checked out PR head `71c4611`, `CGO_ENABLED=0 go build ./cmd/fleet` (and `go build ./...` — both green), ran the binary in a throwaway `qa-210` container under tmux. Seeded `~/.fleet/fleet.log` with representative slog records (INFO/WARN/ERROR) and appended live lines mid-stream to test the live tail.

**Rendering & default.** Navigating to *Settings → Fleet Daemon* shows the new row:
```
> Logs               [All] [Error] [Warn] [Info]
```
ANSI capture confirms `[All]` renders bold/highlighted (color 170) and the rest dimmed (color 240) on first open. The row footer changes to the streaming-specific hint rather than the generic `enter: edit`.

**Segmented control.** `→`/`l` walks All→Error→Warn→Info and stays on Info at the end (no wrap); `←`/`h` walks back to All and stays there. Both arrow keys and vim keys behave identically and clamp correctly.

**Streaming & return.** `Enter` (and `Space`) hand the terminal to a live tail with the header `── fleet.log [<level>] ─── Ctrl-C to return to fleet ───`. New records written by the auto-spawned fleetd (e.g. `mcp server started`, `backup: wrote snapshot`) appeared in the stream live, confirming `tail -F` + `grep --line-buffered`. `Ctrl-C` reliably tears the stream down and restores the Settings page with the Logs row still selected.

**Filtering (verified end-to-end through the TUI):**
- **Error** → only the two `level=ERROR` lines.
- **Warn** → `WARN` + `ERROR` (no `INFO`).
- **Info** → `INFO` + `WARN` + `ERROR`; a live-appended `level=DEBUG` line was correctly filtered out (Info regex is `level=(ERROR|WARN|INFO)`).
- **Leading-space anchor** → an `INFO` line whose message value *begins* with `level=ERROR` (preceded by a quote, not a space) does **not** leak into the Error view. ✅

**Minor, non-blocking observation.** The `grep` anchor is a single leading space (` level=ERROR`). It correctly rejects a value that *starts* with `level=ERROR`, but a message value that contains a space *before* an embedded `level=ERROR` (e.g. `msg="saw level=ERROR upstream"`) would still match. This is an unlikely shape for real fleetd messages and does not affect any acceptance criterion — noting it only for completeness.

**Sanity.** The adjacent *Restart daemon* row still renders and is navigable, `Esc` returns to the Armada landing page, and `q` quits the app cleanly (exit 0). No regressions observed.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed after `f0b0daac` (`test(tui): pin log-stream path to flog.Path; soften anchor comment`).

**All three prior review points are addressed cleanly:**

1. **Path drift guard restored.** `TestDaemonLogStreamCommand` now pins its expected path to `flog.Path()` (the canonical, server-owned source), and the test imports are allowed to reach `internal/flog` because the depguard `client` rule exempts `*_test.go`. I confirmed `flog.Path()` == `filepath.Join(fleetpaths.Dir(), "fleet.log")` byte-for-byte (`fleetpaths.Dir()` → `$HOME/.fleet`, `flog.logFileName` == `"fleet.log"`), so a future rename of `logFileName` now fails this test loudly instead of letting the TUI silently tail a stale filename. 👍
2. **Load-bearing flags asserted.** The test now guards `tail -n 200 -F`, the `grep --line-buffered -E '…'` filter, and the `fleet.log [<level>]` header — the pieces the doc comment calls out as load-bearing.
3. **Anchor comment softened.** The `daemonLogLevels` comment no longer overstates the leading-space anchor ("a space-separated token inside a value still could [match]… worst case is one extra line").

Nothing flagged previously regressed, and the production code is unchanged in this commit.

---

**One new, non-blocking observation (not raised before):**

**The stream is the first `execProcess` command that exits via Ctrl-C, so it likely surfaces a spurious `Command error: exit status 130`.** Every other screen-takeover command exits cleanly — the doctor row waits on `read _`, `$EDITOR`/pager exit 0 — so their `execDoneMsg{err}` is nil. This one is torn down by Ctrl-C: SIGINT kills the `tail … | grep …` pipeline with exit status 130, which `os/exec` returns as a non-nil `*exec.ExitError`. The shared handler (`page_fleet_keys.go`: `case execDoneMsg` → `if msg.err != nil { m.message = "Command error: %v" }`) will then render `Command error: exit status 130` in the settings footer after an intentional, successful Ctrl-C.

The PR notes manual QA passed with "Ctrl-C returns cleanly," so either this is environment-dependent or it was considered cosmetic — flagging because it's a deliberate-exit-shows-an-error pattern that the other exec rows never hit. If it does show, the cheap fix is a stream-specific callback that swallows a signal/130 exit, e.g.:

```go
return execProcess(daemonLogStreamCommand(...), func(err error) tea.Msg {
    var ee *exec.ExitError
    if errors.As(err, &ee) && ee.ExitCode() == 130 { // Ctrl-C is the intended exit
        err = nil
    }
    return execDoneMsg{err}
})
```

Worth a 10-second manual check (Settings → Fleet Daemon → Logs → Enter → Ctrl-C, watch the footer) to confirm whether the message appears.

Otherwise this is a tidy, well-tested, convention-matching change. Nice work.

internal/tui/page_settings.go
```diff
@@ -1079,6 +1098,14 @@ func (settingsPage *settingsPage) updateSettingsNav(m *model, msg tea.Msg) tea.C
 				settingsPage.showKeybindings = true
 				return nil
 			}
+			if item == settingsItemDaemonLogs {
+				// Hand the terminal to a live `tail -f` of fleet.log filtered to the
+				// selected level (same screen-takeover mechanic as the doctor row).
+				return execProcess(
+					daemonLogStreamCommand(daemonLogLevels[settingsPage.logLevel]),
+					func(err error) tea.Msg { return execDoneMsg{err} },
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Unlike the doctor row (which waits on `read _` and exits 0), this stream is meant to be ended with Ctrl-C — and SIGINT kills the `tail … | grep …` pipeline with exit status 130, which `os/exec` returns as a non-nil `*exec.ExitError`. The shared `execDoneMsg` handler (`page_fleet_keys.go`) does `if msg.err != nil { m.message = "Command error: %v" }`, so a normal Ctrl-C will likely flash `Command error: exit status 130` in the footer.

If it does (worth a quick manual check), swallow the intended-exit code in a stream-specific callback:

```go
func(err error) tea.Msg {
    var ee *exec.ExitError
    if errors.As(err, &ee) && ee.ExitCode() == 130 { // Ctrl-C
        err = nil
    }
    return execDoneMsg{err}
}
```

Non-blocking — see the main review for details.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual QA for **#210 — _stream fleetd logs from the settings page_**, head `f0b0daa`. Built the `fleet` binary and drove the real TUI inside an isolated Docker container (`golang:1.25`), exercising the Settings → Fleet Daemon → **Logs** flow as a user would. Seeded `~/.fleet/fleet.log` with INFO/WARN/ERROR lines (and let the live daemon append its own) to verify filtering deterministically.

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./cmd/fleet`) | ✅ Pass | Compiles clean (`-buildvcs=false` only needed due to the mounted-`.git` ownership quirk in the sandbox, not a code issue) |
| Launch & main menu renders | ✅ Pass | Armada view renders, no startup errors |
| Logs row renders in Fleet Daemon section | ✅ Pass | Shows `Logs   [All] [Error] [Warn] [Info]`, **All** highlighted by default |
| Footer hint scoped to Logs row | ✅ Pass | `left/right: choose level  enter: stream fleet.log (Ctrl-C to return)`; reverts to generic hint on other rows |
| ←/→ move selection + clamp | ✅ Pass | All→Error→Warn→Info; clamps at both ends (no wrap-around), per spec |
| vim `h`/`l` keys | ✅ Pass | `l` moves right, `h` moves left |
| Enter streams (screen takeover) | ✅ Pass | Header `── fleet.log [All] ─── Ctrl-C to return to fleet ───`, then the log tail |
| Space streams | ✅ Pass | Identical behavior to Enter |
| Live tail (`tail -F`) | ✅ Pass | A line appended mid-stream appeared immediately |
| Filter: All | ✅ Pass | No grep — every line shown |
| Filter: Error | ✅ Pass | Only `level=ERROR` lines |
| Filter: Warn | ✅ Pass | `WARN`+`ERROR` only, no `INFO` |
| Filter: Info | ✅ Pass | `INFO`+`WARN`+`ERROR` (≈ All, since fleetd doesn't log DEBUG) |
| Leading-space anchor (false-positive guard) | ✅ Pass | An `INFO` line containing `tag=level=ERROR` (no leading space) did **not** leak into the Error filter |
| Path matches live daemon | ✅ Pass | The running fleetd's own INFO lines appeared in the stream → tails the exact `~/.fleet/fleet.log` it writes |
| Ctrl-C returns to TUI | ✅ Pass | Returns cleanly to Settings with cursor still on Logs |
| Remote TUI hides section | ✅ Pass | With `FLEET_SERVER` set, the whole Fleet Daemon section (Logs + Restart) is absent; other sections render |
| Esc / general nav sanity | ✅ Pass | Esc returns to main page; nothing else regressed |

**Overall: ✅ PASS — 18 of 18 scenarios passing.**

## Detailed findings

**Environment.** All execution was isolated in a throwaway `golang:1.25` container with the checked-out workspace mounted; the TUI was driven via tmux inside the container (`send-keys` / `capture-pane`). `HOME=/root`, so the log lived at `/root/.fleet/fleet.log`.

**Segmented control (render + navigation).** On a local TUI the **Logs** row appears in the Fleet Daemon section rendering `[All] [Error] [Warn] [Info]`, with `[All]` highlighted (bold/active style) and the rest dimmed — confirmed by inspecting the ANSI escapes. `→`/`l` advance and `←`/`h` retreat the selection; the control **clamps** at `All` and `Info` rather than wrapping (pressing `→` repeatedly past `Info` stays on `Info`; same for `←` at `All`). The footer help line switches to the Logs-specific hint only while the row is selected.

**Streaming + filtering.** Both `Enter` and `Space` hand the terminal to a live tail. The takeover prints the header `── fleet.log [<level>] ─── Ctrl-C to return to fleet ───` and then streams. Verified the filter matrix against seeded data (3 INFO / 2 WARN / 2 ERROR, plus a live-appended ERROR):
- **All** → all lines, no grep.
- **Error** → ERROR only (3 lines).
- **Warn** → WARN+ERROR (5 lines), no INFO.
- **Info** → INFO+WARN+ERROR (all lines).

Live tailing works: a line `echo`'d into the log mid-stream showed up right away (`-F`). The running daemon's own records (`fleet server started`, `backup: wrote snapshot`, `socket=/root/.fleet/fleet.sock`) also appeared in the stream — confirming the command tails the same file fleetd writes, i.e. the `fleetpaths.Dir()/fleet.log` computation matches `flog.Path()` at runtime.

**Anchor edge case.** Added an `level=INFO` line whose body contained `tag=level=ERROR` (the `level=ERROR` token immediately preceded by `=`, not a space). With the **Error** filter active it did **not** appear — the leading-space anchor behaves as documented. (As the PR description itself notes, a *space-separated* `level=ERROR` token inside a message value would still match; this is an accepted, documented trade-off for a transient view, not a defect.)

**Return & remote.** `Ctrl-C` ends the stream and drops cleanly back into the TUI Settings page (cursor preserved on the Logs row; no terminal corruption observed). For the remote case, I exposed the daemon's unix socket over TCP and launched a second TUI with `FLEET_SERVER=127.0.0.1:7777`; the header switched to `Armada [ 127.0.0.1 (env) ]` and the **entire Fleet Daemon section — including the new Logs row — was hidden**, while every other settings section still rendered. Matches the "local-only" intent.

No defects or regressions found. Recommend ✅.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Re-QA of **#210 — _stream fleetd logs from the settings page_** at the new head `0c14e16` (_fix(tui): treat Ctrl-C exit of the log stream as a clean return_). This commit lands exactly the follow-up the reviewer asked for — *"watch the footer after Ctrl-C to confirm whether the message appears."* Rebuilt the `fleet` binary and drove the real TUI inside an isolated Docker container (`golang:1.25`), re-running the full Settings → Fleet Daemon → **Logs** flow plus a focused check on the new Ctrl-C handling. Seeded `~/.fleet/fleet.log` with INFO/WARN/ERROR lines (and let the live daemon append its own).

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./cmd/fleet`) | ✅ Pass | Compiles clean (`-buildvcs=false` only for the mounted-`.git` ownership quirk, not a code issue) |
| Launch & main menu renders | ✅ Pass | `Armada [ local ]`, no startup errors |
| Logs row renders in Fleet Daemon section | ✅ Pass | `Logs   [All] [Error] [Warn] [Info]`, **All** highlighted by default |
| Footer hint scoped to Logs row | ✅ Pass | `left/right: choose level  enter: stream fleet.log (Ctrl-C to return)`; generic hint elsewhere |
| ←/→ move selection + clamp | ✅ Pass | All→Error→Warn→Info; clamps at both ends (no wrap) |
| vim `h`/`l` keys | ✅ Pass | `l` right, `h` left |
| Enter streams (screen takeover) | ✅ Pass | Header `── fleet.log [All] ─── Ctrl-C to return to fleet ───` then the tail |
| Space streams | ✅ Pass | Identical to Enter |
| **Ctrl-C returns cleanly — no error flash** | ✅ Pass | **New commit verified: returns to Settings, cursor on Logs, NO "Command error: exit status 130" in footer** |
| Live tail (`tail -F`) | ✅ Pass | A line appended mid-stream appeared immediately |
| Filter: All | ✅ Pass | No grep — every line shown |
| Filter: Error | ✅ Pass | Only `level=ERROR` lines (2) |
| Filter: Warn | ✅ Pass | `WARN`+`ERROR` (4), no `INFO` |
| Filter: Info | ✅ Pass | `INFO`+`WARN`+`ERROR` (≈ All, fleetd logs no DEBUG) |
| Leading-space anchor (false-positive guard) | ✅ Pass | An `INFO` line containing `tag=level=ERROR` did **not** leak into the Error filter |
| Path matches live daemon | ✅ Pass | The running fleetd's own `pid=5217` INFO lines appeared in the stream |
| Remote TUI hides section | ✅ Pass | With `FLEET_SERVER` set, the entire Fleet Daemon section (Logs + Restart) is absent; other sections render |
| Esc / general nav sanity | ✅ Pass | Esc returns to main page; nothing regressed |

**Overall: ✅ PASS — 18 of 18 scenarios passing.**

## Detailed findings

**Environment.** All execution isolated in a throwaway `golang:1.25` container with the checked-out workspace mounted; the TUI was driven via tmux inside the container (`send-keys` / `capture-pane -e` to read ANSI styling). `HOME=/root`, so the log lived at `/root/.fleet/fleet.log`.

**🔎 Focus — Ctrl-C clean return (the new commit).** This is the behavior the reviewer flagged on the previous head. Streamed at level **All** (`Enter`), then pressed `Ctrl-C`. The TUI dropped straight back to the Settings page with the cursor still on the **Logs** row and the Logs-specific footer hint — and crucially **no `Command error: exit status 130` (or any error text) anywhere in the footer**. Re-checked after streaming at every level; clean every time. `silenceStreamInterrupt` is doing its job: status 130 (128 + SIGINT) is swallowed while a genuine non-zero exit would still surface (covered by the added unit test). ✅

**Segmented control (render + navigation).** On a local TUI the **Logs** row renders `[All] [Error] [Warn] [Info]` with `[All]` bold/highlighted (verified via the ANSI escapes: active = `1m;38;5;170m`, others dimmed `38;5;240m`) and the rest dimmed. `→`/`l` advance and `←`/`h` retreat, and the control **clamps** at `All`/`Info` rather than wrapping (repeated `→` past `Info` stays on `Info`; same for `←` at `All`). Footer help switches to the Logs-specific hint only while the row is selected.

**Streaming + filtering.** Both `Enter` and `Space` hand the terminal to a live tail with the header `── fleet.log [<level>] ─── Ctrl-C to return to fleet ───`. Filter matrix against the seeded data (3 INFO / 2 WARN / 2 ERROR):
- **All** → all lines, no grep.
- **Error** → ERROR only (2).
- **Warn** → WARN+ERROR (4), no INFO.
- **Info** → INFO+WARN+ERROR (all).

Live tailing confirmed: an `ERROR` line appended to the log mid-stream showed up right away (`-F`). The running daemon's own records (`fleet server started`, `mcp server started`, `backup: wrote snapshot`, all `pid=5217`) also appeared — confirming the command tails the same file fleetd writes, i.e. `fleetpaths.Dir()/fleet.log` matches `flog.Path()` at runtime.

**Anchor edge case.** A seeded `level=INFO` line whose body contained `tag=level=ERROR` (preceded by `=`, not a space) did **not** appear under the **Error** filter — the leading-space anchor behaves as documented. (Per the PR description, a *space-separated* `level=ERROR` token inside a value could still match; accepted trade-off for a transient view, not a defect.)

**Remote.** Bridged the daemon's unix socket to TCP via `socat` and launched a second TUI with `FLEET_SERVER=127.0.0.1:7777`; header switched to `Armada [ 127.0.0.1 (env) ]` and the settings page went straight from **Tool Status → Help**, with the **entire Fleet Daemon section (Logs + Restart) hidden**, while every other section still rendered. Matches the local-only intent.

No defects or regressions found. The reviewer's footer concern is resolved. Recommend ✅.

internal/tui/commands.go
```diff
@@ -274,6 +275,70 @@ func logsCommand(fleetName string, instance *fleet.Instance) *exec.Cmd {
...
+// streamInterruptExitCode is the status a shell pipeline reports when killed by
+// SIGINT (128 + SIGINT) — i.e. the user pressed Ctrl-C to end the stream.
+const streamInterruptExitCode = 130
+
+// silenceStreamInterrupt nils out the expected Ctrl-C exit (status 130) of a
+// log-stream child so the shared execDoneMsg handler doesn't flash it in the
+// footer as "Command error: exit status 130"; any other failure passes through.
+func silenceStreamInterrupt(err error) error {
+	var ee *exec.ExitError
+	if errors.As(err, &ee) && ee.ExitCode() == streamInterruptExitCode {
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This `ExitCode() == 130` guard may not actually catch the documented Ctrl-C return on the most common shells. A terminal Ctrl-C delivers SIGINT to the whole foreground process group (`sh` + `tail` + `grep`), and on both **dash** (Debian/Ubuntu `/bin/sh`, and the usual `/bin/sh` in devcontainers) and **bash**, the parent `sh` is then *terminated by SIGINT* rather than `exit(130)` — so `os/exec` reports it as `signal: interrupt` with `ExitCode() == -1`, and this branch returns the error unchanged. The footer would then flash `Command error: signal: interrupt` — the exact UX this commit set out to remove.

Reproduced here with a synthetic group-SIGINT (`kill -INT -<pgid>`, which is what a terminal Ctrl-C does to the foreground group) against `sh -c "… tail -F … | grep …"`:

```
sh    tail|grep          raw=signal: interrupt  ExitCode=-1  shipped.silenced=false
sh    bare tail          raw=signal: interrupt  ExitCode=-1  shipped.silenced=false
bash  tail|grep          raw=signal: interrupt  ExitCode=-1  shipped.silenced=false
```

(`ExitCode() == 130` *does* occur on shells that propagate `128+SIGINT` as their own exit code, so the current branch isn't wrong — it's just incomplete.) Suggest also treating signal-SIGINT death as a clean exit so it works on both:

```go
func silenceStreamInterrupt(err error) error {
	var ee *exec.ExitError
	if errors.As(err, &ee) {
		if ee.ExitCode() == streamInterruptExitCode { // shell exposed 128+SIGINT as an exit code
			return nil
		}
		// dash/bash under a group Ctrl-C: sh itself dies of SIGINT (ExitCode()==-1).
		if ws, ok := ee.Sys().(syscall.WaitStatus); ok && ws.Signaled() && ws.Signal() == syscall.SIGINT {
			return nil
		}
	}
	return err
}
```

(adds `syscall`.) Worth a quick manual check on a dash host — the QA runs all predate this commit, so the silence path was never driven live. If you *do* observe `exit status 130` on your machine, keep the existing branch and just add the signal branch; it's strictly safer either way.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed at `0c14e164` (synchronize). The three earlier review points are all addressed: the level-anchor comment is softened, `TestDaemonLogStreamCommand` now pins the expected path to `flog.Path()` (so a future `logFileName` rename fails loudly instead of silently tailing a stale name), and the Ctrl-C return is now centralized in `daemonLogStreamCmd` + `silenceStreamInterrupt`.

I checked the load-bearing bits and they hold up:
- `filepath.Join(fleetpaths.Dir(), "fleet.log")` is byte-for-byte `flog.Path()` — both resolve to `$HOME/.fleet/fleet.log` — so the depguard-driven duplication is correct, and the test guards the divergence.
- The `%q` + `printf '%s'` shell construction matches the existing `logsCommand` convention; the header is passed as a printf **argument**, not a format string, so there's no format-injection from the level label.
- Nice separation: the `*exec.Cmd` builder (`daemonLogStreamCommand`) is unit-tested directly while the `tea.Cmd` wrapper stays thin.

**One substantive concern on the latest commit** — the `ExitCode() == 130` silence check likely misses the real Ctrl-C path on dash/bash, where the interrupt surfaces as `signal: interrupt` (`ExitCode() == -1`), not exit status 130. Evidence + a reproduction and a suggested fix are inline on `commands.go`. Worth confirming since this commit landed after all the manual-QA runs, so the silence path itself was never exercised live.

**Minor / non-blocking:**
- **grep portability:** `grep --line-buffered` is a GNU/BSD extension. A host whose `grep` is busybox (some minimal Linux setups) would error the pipeline rather than stream. Fine for the typical Debian/macOS host the TUI runs on — just flagging since this runs on the user's machine, not inside the (Debian) container.
- The space-anchored ` level=` false-positive is already documented in the comment and accepted as a transient-view trade-off — no action needed.

Nothing here blocks; the inline silence point is the only thing I'd tighten before merge.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Re-review on the new commits** (`synchronize` → head `78674b3`). I checked out the branch and ran `go build ./...`, `go vet ./internal/tui/...`, and `go test ./internal/tui/...` — all green.

### Prior review feedback — all resolved ✅
The four inline comments from the earlier head have each been addressed:

1. **Anchor overstatement** (`commands.go`) — comment reworded to "makes a stray `level=ERROR` … unlikely to match — a space-separated token inside a value still could", which matches reality. ✅
2. **`fleet.log` filename divergence** — production still computes the path via `fleetpaths` (correct, since depguard bars the TUI from importing `internal/flog`), and the *test* now pins the expected path to `flog.Path()`. I confirmed `.golangci.yml` exempts `**/*_test.go` from the `client` depguard rule, so the test import is legal and a future `logFileName` rename now fails `TestDaemonLogStreamCommand` loudly. ✅
3. **Ctrl-C "exit status 130" footer flash** — `silenceStreamInterrupt` swallows the intended interrupt before it reaches the `execDoneMsg` handler (which otherwise renders `Command error: …`). ✅
4. **`ExitCode()==130` incomplete on dash/bash** — now also treats signal-death SIGINT (`WaitStatus.Signaled() && Signal()==SIGINT`, `ExitCode()==-1`) as clean, and `TestSilenceStreamInterrupt` drives both shapes deterministically plus the pass-through and nil cases. ✅

### New code — looks solid
- The interrupt-silencing logic is well-reasoned and the unit test exercises both shells' SIGINT shapes without flaking.
- The segmented control clamps correctly (`max(0, min(...))`), enter-to-stream is wired separately from cycle/edit, and the footer hint is row-scoped.
- Shell construction is injection-safe: `path` and `header` are `%q`-quoted and the `grep` pattern comes from a fixed in-code table, not user input.

### Minor / non-blocking
- `tail -n 200 -F <path>` on a **not-yet-existent** `~/.fleet/fleet.log` (e.g. a fresh install where fleetd hasn't logged yet) will emit a `tail: cannot open … No such file or directory` diagnostic and then follow once the file appears. `-F` is the right choice for rotation resilience, and the section is local-only where the daemon almost always exists — just noting that the first-run UX is a transient error line rather than a blank screen. Not worth changing unless you want a friendlier empty-state message.

The previously-noted level-filter false positive (a value containing a space-prefixed ` level=ERROR`) remains a documented, accepted trade-off for a transient view — agreed, not worth a stricter parse here.

Nice, tidy change — no blocking concerns from this pass.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Re-QA of **#210** after the new commits — head `78674b3` (`fix(tui): also silence signal-death Ctrl-C exit of the log stream`) on top of `0c14e16` (`treat Ctrl-C exit … as a clean return`). These commits address the reviewer's flagged concern that ending the stream with **Ctrl-C** would flash `Command error: signal: interrupt` / `exit status 130` in the footer — a path the **prior QA runs never drove live**. So this run focuses on driving that exit live, plus a full regression pass of the feature.

Built the `fleet` binary and drove the **real TUI** inside an isolated Docker container (`golang:1.25-bookworm`, `/bin/sh` → **dash**, tmux) via send-keys / capture-pane. The auto-spawned fleetd also wrote real records to `~/.fleet/fleet.log` during the session. No unit/integration tests were run.

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` binary | ✅ Pass | `go build ./cmd/fleet` clean (`-buildvcs=false`, mount quirk); 45 MB binary, `--help` works |
| Launch TUI & Settings renders | ✅ Pass | Armada → Settings render, no startup errors |
| Logs row renders + default **All** | ✅ Pass | `Logs  [All] [Error] [Warn] [Info]`; ANSI confirms **All** bold/170, rest dim/240 |
| Footer hint scoped to Logs row | ✅ Pass | `left/right: choose level  enter: stream fleet.log (Ctrl-C to return)` |
| →  / vim `l` move + clamp at Info | ✅ Pass | All→Error→Warn→Info; clamps at Info (no wrap) |
| ←  / vim `h` move + clamp at All | ✅ Pass | walks back to All; clamps at All (no wrap) |
| Enter → screen-takeover stream | ✅ Pass | Header `── fleet.log [Error] ─── Ctrl-C to return to fleet ───` |
| Space alt-binding streams | ✅ Pass | `case "enter", " "` — Space launches the stream too |
| Live `tail -F` picks up new lines | ✅ Pass | Appended ERROR appeared live; appended INFO stayed hidden under Error |
| **All** filter (no grep) | ✅ Pass | INFO/WARN/ERROR all shown, incl. real fleetd records (`fleet server started`, `backup: wrote snapshot`) |
| **Error** filter | ✅ Pass | ERROR lines only |
| **Warn** filter | ✅ Pass | WARN + ERROR; no INFO/DEBUG |
| **Info** filter | ✅ Pass | INFO+WARN+ERROR; a `level=DEBUG` line correctly **excluded** |
| **Ctrl-C → clean return (no error footer)** | ✅ Pass | **The fix.** Driven live ×4 on dash — returns to Settings, cursor on Logs, **no** `Command error: signal: interrupt` / `exit status 130` flash |
| Selection ephemeral / preserved in-session | ✅ Pass | Stayed on Error after returning from the stream |
| Sanity: Esc back, `q` quits | ✅ Pass | Esc → Armada; `q` quits clean (exit 0); Restart daemon row unaffected |
| Remote-hidden Fleet Daemon/Logs | ⏭️ Skipped (live) | Remote TUI (`FLEET_SERVER`) panics in bubbletea init **before any render** — pre-existing & unrelated to this PR; section gated by the unchanged `Visible: !fleetclient.IsRemote()` |

**Overall: ✅ PASS — 16 of 16 executed scenarios passing (1 skipped-live, gated by the unchanged visibility func). The headline fix — clean Ctrl-C return with no error-footer flash — is confirmed live on dash.**

---

### Headline fix verified live (the reason for `0c14e16` + `78674b3`)
The reviewer noted that on dash/bash a terminal Ctrl-C delivers SIGINT to the whole `sh → tail → grep` foreground group, so `sh` dies **by signal** (`os/exec` reports `ExitCode() == -1`, `signal: interrupt`) — and the shared `execDoneMsg` handler would render `Command error: signal: interrupt`. `silenceStreamInterrupt` now nils out **both** the 128+SIGINT exit-code shape (130) **and** the signal-death shape.

This container's `/bin/sh` is **dash**, i.e. the exact signal-death path commit `78674b3` adds and that no prior QA run exercised. I ended the stream with Ctrl-C **four separate times** (from the Error, All, Warn and Space/Info streams). Every time the TUI returned to the Settings page with the cursor back on the **Logs** row, the normal footer hint, and **no error message anywhere on screen** (grep for `command error|signal|interrupt|exit status` → no match each time). Meanwhile the surrounding handler is unchanged, so a *genuine* non-zero exit still surfaces. Behaves exactly as intended.

### Filtering verified end-to-end through the TUI
Seeded `~/.fleet/fleet.log` with slog text records (INFO/WARN/ERROR/DEBUG) and appended lines mid-stream:
- **All** → full file, including live records written by the running fleetd.
- **Error** → only `level=ERROR`; a live-appended INFO line was correctly suppressed.
- **Warn** → `WARN` + `ERROR` only.
- **Info** → `INFO`+`WARN`+`ERROR`; a `level=DEBUG` line was correctly excluded (regex `level=(ERROR|WARN|INFO)`).

Header is per-level (`── fleet.log [<level>] ─── Ctrl-C to return to fleet ───`), and `tail -F` is genuinely live.

### Notes (non-blocking, informational)
- **`level=` substring in an attribute value** — as the PR description and the softened code comment already acknowledge, a value containing a space-separated ` level=ERROR` token (e.g. a `container exec` cmd string) can still match the Error/Warn filter. This is documented expected behavior, not a regression; real fleetd records don't hit it. No action implied.
- **Remote-mode panic is pre-existing** — launching with `FLEET_SERVER` set panics in bubbletea init (`nil pointer dereference`, exit 1) before any page renders. The diff touches only `internal/tui/commands.go` and `internal/tui/page_settings.go`; the new Logs row sits inside the already remote-hidden **Fleet Daemon** section. Not caused by this PR.

_Tested at head `78674b3` in a throwaway `qa-210` Docker container (now torn down)._

PR #211, opened 2026-06-24
fix(tui): stop session create failing forever with bare "exit code 1"

BenjaminBenetti:
## Problem

Creating a tmux session from the TUI could fail with an opaque **`Error Creating Session: Exit code 1`**. For **layout presets** the failure was *persistent* — it stuck around, and (unlike the single-session case) **restarting fleetd did not clear it**.

## Root cause

- **The reason was swallowed.** Every TUI create ran `tmux new-session ... 2>/dev/null`. On a name collision tmux exits 1 with `duplicate session: NAME` on stderr — `2>/dev/null` threw that away, leaving only a bare `exit status 1`. (The CLI/MCP paths never suppressed it, which is why only the TUI was undiagnosable.)
- **Layouts collide on a fixed name with no cleanup.** A layout's group root is named after the preset (`<inst>~<presetName>`), and `buildPresetSessionScript` chained panes with `&&`. If any step failed, the root it already created was left behind — and the error handler did no cleanup. Because the name is fixed and tmux sessions live in the **container** (not fleetd), every retry re-collided on the identical name *forever*; a daemon restart couldn't help.

Single-session creates dodged permanence because their auto-name varies; layout names don't.

## Fix (P0 only)

1. **Surface the reason** — create `new-session` *and* `send-keys` steps use `2>&1`, and `runSessionScript` folds the captured output's **last non-empty line** into the error (dropping any `==> Installing tmux...` preamble). The status line now shows `duplicate session: <name>` instead of `exit status 1`.
2. **No more permanent collision** — `groupIDFor` keeps the readable preset/typed group name **when its group id is free**, and falls back to a **fresh random id** only when that group already exists. So a first apply reads as `dev` (test `290` types `fromtpl` and asserts `alpha~fromtpl` — preserved), while a re-apply / leftover mints a random id and succeeds instead of colliding forever. A random fallback (rather than `dev-2`) also avoids minting a name that prefix-matches a sibling group in the prefix-based group ops.
3. **Transactional preset script** — the root is created on its own with `|| exit 1` (so a **pre-existing** root is never touched), then the rest runs in a group that, on any failure, kills exactly the sessions *this run* created (by exact `=name:` target). No stranded root; no killing a live group.

## Testing

- Unit: `TestBuildPresetSessionScript` (+ single-pane no-cleanup), `TestGroupIDFor`, `TestRunSessionScript*` (reason surfaced / install-preamble dropped / bare exit / transport-error passthrough).
- Verified the **real generated script** against live tmux 3.3a (isolated socket): clean create; duplicate fails loud *and preserves the existing root*; mid-chain failure tears down only this run's root.
- `go build ./...` + full non-integration `go test` green; CI green (unit + 7 integration shards incl. `290_tui_layout_preset`); automated QA 7/7.

## Scope

P0 only. **Deferred to P1** (owner's call):
1. `nextSessionName` always returns `session-2` (matches a bare `session-` prefix against grouped cache names).
2. Self-correcting container-user probe (create runs as the devcontainer `remoteUser`; the poller pins a possibly-different user for the daemon's lifetime).
3. **Boundary-aware group matching** — the group ops match membership by raw prefix (`HasPrefix(name, "<inst>~<gid>")`), so readable layout-group names can't safely survive a re-apply (hence the random-id fallback here). Fixing this would also let re-applied presets keep readable names.

### DIFF
--- a/internal/tui/dialog_session.go
+++ b/internal/tui/dialog_session.go
@@ -123,24 +123,26 @@ func (fleetPage *fleetPage) saveCreateSession(m *model) tea.Cmd {
 	// A template was Tab-selected: mint the whole group — one session per
 	// pane, the resolved name as the group ID — and run each pane's command.
 	if preset != nil {
-		// The group root is named after the preset, so re-applying a preset (or
-		// a leftover from a prior failed apply) would collide on new-session and
-		// fail forever. Suffix to a free name (dev → dev-2) so the root is always
-		// creatable; load the runtime session list first so the check sees what's
-		// actually live.
-		ensureSessionsLoaded(m, ref)
-		name = uniqueGroupName(sanitized, name, m.sessionStore.Sessions(ref))
-		fullName = GroupSessionName(sanitized, name)
+		// Mint the group under a FRESH RANDOM id (like a split group), not the
+		// preset name. A preset-derived root name re-collides forever on
+		// new-session once that group exists — the reported "stuck, restart
+		// doesn't help" bug, since the session lives in the container. A random
+		// id never collides on re-apply, and (unlike a readable suffix such as
+		// dev-2) can't prefix-match a sibling group in the prefix-based group ops
+		// (open/rename/delete). Readable/typed layout-group names need
+		// boundary-aware matching first — that's the deferred P1 prefix work.
+		groupID := randomHex(3)
+		fullName = GroupSessionName(sanitized, groupID)
 
 		sessions := make([]string, 0, preset.PaneCount())
 		sessions = append(sessions, fullName)
 		for i := 1; i < preset.PaneCount(); i++ {
-			sessions = append(sessions, NewGroupPaneSessionName(sanitized, name))
+			sessions = append(sessions, NewGroupPaneSessionName(sanitized, groupID))
 		}
 		fleetPage.mode = viewNormal
 		fleetPage.blurDialogFields()
-		m.message = fmt.Sprintf("Creating session %s from %s...", name, preset.Name)
-		return createSessionGroupFromPresetCmd(ref, name, sessions, *preset)
+		m.message = fmt.Sprintf("Creating session from %s...", preset.Name)
+		return createSessionGroupFromPresetCmd(ref, groupID, sessions, *preset)
 	}
 
 	fleetPage.mode = viewNormal
--- a/internal/tui/dialog_session.go
+++ b/internal/tui/dialog_session.go
@@ -123,6 +123,15 @@ func (fleetPage *fleetPage) saveCreateSession(m *model) tea.Cmd {
 	// A template was Tab-selected: mint the whole group — one session per
 	// pane, the resolved name as the group ID — and run each pane's command.
 	if preset != nil {
+		// The group root is named after the preset, so re-applying a preset (or
+		// a leftover from a prior failed apply) would collide on new-session and
+		// fail forever. Suffix to a free name (dev → dev-2) so the root is always
+		// creatable; load the runtime session list first so the check sees what's
+		// actually live.
+		ensureSessionsLoaded(m, ref)
+		name = uniqueGroupName(sanitized, name, m.sessionStore.Sessions(ref))
+		fullName = GroupSessionName(sanitized, name)
+
 		sessions := make([]string, 0, preset.PaneCount())
 		sessions = append(sessions, fullName)
 		for i := 1; i < preset.PaneCount(); i++ {
--- a/internal/tui/dialog_session.go
+++ b/internal/tui/dialog_session.go
@@ -123,15 +123,17 @@ func (fleetPage *fleetPage) saveCreateSession(m *model) tea.Cmd {
 	// A template was Tab-selected: mint the whole group — one session per
 	// pane, the resolved name as the group ID — and run each pane's command.
 	if preset != nil {
-		// Mint the group under a FRESH RANDOM id (like a split group), not the
-		// preset name. A preset-derived root name re-collides forever on
-		// new-session once that group exists — the reported "stuck, restart
-		// doesn't help" bug, since the session lives in the container. A random
-		// id never collides on re-apply, and (unlike a readable suffix such as
-		// dev-2) can't prefix-match a sibling group in the prefix-based group ops
-		// (open/rename/delete). Readable/typed layout-group names need
-		// boundary-aware matching first — that's the deferred P1 prefix work.
-		groupID := randomHex(3)
+		// Keep the readable preset/typed name as the group id when it's free,
+		// else fall back to a fresh random id. A preset-derived root name
+		// re-collides forever on new-session once that group exists — the
+		// reported "stuck, restart doesn't help" bug, since the session lives in
+		// the container. Falling back to random (rather than auto-suffixing to
+		// dev-2) avoids minting a sibling whose name a prefix-based group op would
+		// match against "dev"; readable names for re-applies need boundary-aware
+		// matching first — the deferred P1 prefix work. Load the runtime list so
+		// the free check sees what's actually live.
+		ensureSessionsLoaded(m, ref)
+		groupID := groupIDFor(sanitized, name, m.sessionStore.Sessions(ref))
 		fullName = GroupSessionName(sanitized, groupID)
 
 		sessions := make([]string, 0, preset.PaneCount())
@@ -141,7 +143,7 @@ func (fleetPage *fleetPage) saveCreateSession(m *model) tea.Cmd {
 		}
 		fleetPage.mode = viewNormal
 		fleetPage.blurDialogFields()
-		m.message = fmt.Sprintf("Creating session from %s...", preset.Name)
+		m.message = fmt.Sprintf("Creating session %s from %s...", groupID, preset.Name)
 		return createSessionGroupFromPresetCmd(ref, groupID, sessions, *preset)
 	}
 
--- a/internal/tui/groups.go
+++ b/internal/tui/groups.go
@@ -123,6 +123,24 @@ func NewGroupPaneSessionName(sanitizedInstance, groupID string) string {
 	return GroupSessionName(sanitizedInstance, groupID) + groupSep + randomHex(2)
 }
 
+// groupIDFor returns desired when no live group in the instance already uses it,
+// otherwise a fresh random id. Layout presets create their root under the
+// readable preset/typed name; honoring it keeps the group named that on first
+// apply, but a re-apply (or a leftover from a prior failed apply) falls back to
+// a random id so the root never collides forever on `tmux new-session` — the
+// reported "stuck, a restart doesn't help" bug, since the session lives in the
+// container. It deliberately does NOT auto-suffix (dev-2): that would prefix-
+// match a sibling group ("dev") in the prefix-based group ops (open/rename/
+// delete) — readable names for re-applies need boundary-aware matching first.
+func groupIDFor(sanitizedInstance, desired string, existing []tmuxSession) string {
+	for _, s := range existing {
+		if gid, ok := parseGroupID(sanitizedInstance, s.Name); ok && gid == desired {
+			return randomHex(3)
+		}
+	}
+	return desired
+}
+
 // ResolveSessionName maps a user-supplied session name to its canonical
 // grouped form for the given instance. Names that already follow the
 // convention pass through unchanged; anything else becomes the root
--- a/internal/tui/groups.go
+++ b/internal/tui/groups.go
@@ -3,7 +3,6 @@ package tui
 import (
 	"crypto/rand"
 	"encoding/hex"
-	"fmt"
 	"sort"
 	"strings"
 	"time"
@@ -122,31 +121,6 @@ func NewGroupRootSessionName(sanitizedInstance string) string {
 // an existing group: <instance>~<groupID>~<hex>.
 func NewGroupPaneSessionName(sanitizedInstance, groupID string) string {
 	return GroupSessionName(sanitizedInstance, groupID) + groupSep + randomHex(2)
-}
-
-// uniqueGroupName returns desired, or desired-2/-3/… when a group with that ID
-// already lives in the instance. Layout presets create their root under a FIXED,
-// preset-derived name; without this, re-applying a preset (or a leftover from a
-// prior failed apply) would collide on `tmux new-session` and fail forever with
-// "duplicate session". Suffixing keeps the readable preset name in the common
-// case while guaranteeing the root is free — mirroring uniquePresetName and the
-// auto-increment the single-session path intends.
-func uniqueGroupName(sanitizedInstance, desired string, existing []tmuxSession) string {
-	taken := make(map[string]bool, len(existing))
-	for _, s := range existing {
-		if gid, ok := parseGroupID(sanitizedInstance, s.Name); ok {
-			taken[gid] = true
-		}
-	}
-	if !taken[desired] {
-		return desired
-	}
-	for n := 2; ; n++ {
-		candidate := fmt.Sprintf("%s-%d", desired, n)
-		if !taken[candidate] {
-			return candidate
-		}
-	}
 }
 
 // ResolveSessionName maps a user-supplied session name to its canonical
--- a/internal/tui/groups.go
+++ b/internal/tui/groups.go
@@ -3,6 +3,7 @@ package tui
 import (
 	"crypto/rand"
 	"encoding/hex"
+	"fmt"
 	"sort"
 	"strings"
 	"time"
@@ -121,6 +122,31 @@ func NewGroupRootSessionName(sanitizedInstance string) string {
 // an existing group: <instance>~<groupID>~<hex>.
 func NewGroupPaneSessionName(sanitizedInstance, groupID string) string {
 	return GroupSessionName(sanitizedInstance, groupID) + groupSep + randomHex(2)
+}
+
+// uniqueGroupName returns desired, or desired-2/-3/… when a group with that ID
+// already lives in the instance. Layout presets create their root under a FIXED,
+// preset-derived name; without this, re-applying a preset (or a leftover from a
+// prior failed apply) would collide on `tmux new-session` and fail forever with
+// "duplicate session". Suffixing keeps the readable preset name in the common
+// case while guaranteeing the root is free — mirroring uniquePresetName and the
+// auto-increment the single-session path intends.
+func uniqueGroupName(sanitizedInstance, desired string, existing []tmuxSession) string {
+	taken := make(map[string]bool, len(existing))
+	for _, s := range existing {
+		if gid, ok := parseGroupID(sanitizedInstance, s.Name); ok {
+			taken[gid] = true
+		}
+	}
+	if !taken[desired] {
+		return desired
+	}
+	for n := 2; ; n++ {
+		candidate := fmt.Sprintf("%s-%d", desired, n)
+		if !taken[candidate] {
+			return candidate
+		}
+	}
 }
 
 // ResolveSessionName maps a user-supplied session name to its canonical
--- a/internal/tui/sessions.go
+++ b/internal/tui/sessions.go
@@ -199,10 +199,10 @@ type presetSessionsCreatedMsg struct {
 // already exists), the script exits immediately and the cleanup never runs, so a
 // pre-existing session is left untouched — never killed. Only once the root is
 // ours does the rest run; if any later step fails, the cleanup tears down the
-// root plus its panes. The caller picked the root name with uniqueGroupName, so
-// the whole "<inst>~<group>~*" namespace was free to begin with — the kills can
-// only hit sessions this run made, never a live group — and a partial chain
-// can't strand a root that would collide forever on retry. Every tmux step uses
+// root plus its panes. The caller mints the root under a fresh random group id,
+// so the whole "<inst>~<group>~*" namespace is this run's — the kills can only
+// hit sessions this run made, never a live group — and a partial chain can't
+// strand a root that would collide forever on retry. Every tmux step uses
 // 2>&1 so a failure at ANY step (not just new-session) keeps its reason (e.g.
 // "duplicate session: NAME") for runSessionScript to surface, instead of a bare
 // "exit status 1".
--- a/internal/tui/sessions.go
+++ b/internal/tui/sessions.go
@@ -115,19 +115,34 @@ func ensureSessionsLoaded(m *model, ref InstanceRef) {
 // create scripts merge tmux's stderr into stdout (2>&1) so a captured reason
 // (e.g. "duplicate session: NAME") rides along here instead of being lost to a
 // bare "exit status 1" — the single biggest diagnosability win for the TUI,
-// which otherwise swallowed why a session failed to create.
+// which otherwise swallowed why a session failed to create. Only the LAST
+// non-empty line is surfaced: that's the actual failing command's message,
+// dropping any TmuxEnsureInstalled "==> Installing tmux..." preamble (while
+// still surfacing a real "ERROR: failed to install tmux", which is the last
+// line in that case).
 func runSessionScript(ref InstanceRef, script string) (string, error) {
 	out, code, err := runInstanceCommand(ref.Fleet, ref.Instance, []string{"sh", "-c", script})
 	if err != nil {
 		return out, err
 	}
 	if code != 0 {
-		if reason := strings.TrimSpace(out); reason != "" {
+		if reason := lastNonEmptyLine(out); reason != "" {
 			return out, fmt.Errorf("exit status %d: %s", code, reason)
 		}
 		return out, fmt.Errorf("exit status %d", code)
 	}
 	return out, nil
+}
+
+// lastNonEmptyLine returns the last non-blank line of s (trimmed), or "".
+func lastNonEmptyLine(s string) string {
+	lines := strings.Split(s, "\n")
+	for i := len(lines) - 1; i >= 0; i-- {
+		if t := strings.TrimSpace(lines[i]); t != "" {
+			return t
+		}
+	}
+	return ""
 }
 
 // sessionCreatedMsg is sent after creating a new tmux session.
@@ -199,10 +214,11 @@ type presetSessionsCreatedMsg struct {
 // already exists), the script exits immediately and the cleanup never runs, so a
 // pre-existing session is left untouched — never killed. Only once the root is
 // ours does the rest run; if any later step fails, the cleanup tears down the
-// root plus its panes. The caller mints the root under a fresh random group id,
-// so the whole "<inst>~<group>~*" namespace is this run's — the kills can only
-// hit sessions this run made, never a live group — and a partial chain can't
-// strand a root that would collide forever on retry. Every tmux step uses
+// root plus its panes by EXACT "=name:" target. The caller picked a group id
+// that was free (groupIDFor), so nothing pre-existed under "<inst>~<group>~*" —
+// the kills can only hit sessions this run made, never a live group — and a
+// partial chain can't strand a root that would collide forever on retry. Every
+// tmux step uses
 // 2>&1 so a failure at ANY step (not just new-session) keeps its reason (e.g.
 // "duplicate session: NAME") for runSessionScript to surface, instead of a bare
 // "exit status 1".
--- a/internal/tui/sessions.go
+++ b/internal/tui/sessions.go
@@ -117,9 +117,10 @@ func ensureSessionsLoaded(m *model, ref InstanceRef) {
 // bare "exit status 1" — the single biggest diagnosability win for the TUI,
 // which otherwise swallowed why a session failed to create. Only the LAST
 // non-empty line is surfaced: that's the actual failing command's message,
-// dropping any TmuxEnsureInstalled "==> Installing tmux..." preamble (while
-// still surfacing a real "ERROR: failed to install tmux", which is the last
-// line in that case).
+// dropping any TmuxEnsureInstalled "==> Installing tmux..." preamble. (If the
+// install itself fails, TmuxEnsureInstalled falls through to `tmux new-session`,
+// which then fails with e.g. "tmux: not found" — that becomes the last line, so
+// a missing-tmux failure still reads usefully.)
 func runSessionScript(ref InstanceRef, script string) (string, error) {
 	out, code, err := runInstanceCommand(ref.Fleet, ref.Instance, []string{"sh", "-c", script})
 	if err != nil {
--- a/internal/tui/sessions.go
+++ b/internal/tui/sessions.go
@@ -198,12 +198,14 @@ type presetSessionsCreatedMsg struct {
 // The root is created first and on its own: if THAT fails (e.g. the group name
 // already exists), the script exits immediately and the cleanup never runs, so a
 // pre-existing session is left untouched — never killed. Only once the root is
-// ours does the rest run; if any later step fails, the cleanup tears down every
-// session this run created (the root plus the random-suffixed panes — all names
-// unique to this attempt), so a partial chain can't strand a root that would
-// collide forever on retry. 2>&1 keeps tmux's failure reason (e.g. "duplicate
-// session: NAME") so runSessionScript can surface it instead of a bare "exit
-// status 1".
+// ours does the rest run; if any later step fails, the cleanup tears down the
+// root plus its panes. The caller picked the root name with uniqueGroupName, so
+// the whole "<inst>~<group>~*" namespace was free to begin with — the kills can
+// only hit sessions this run made, never a live group — and a partial chain
+// can't strand a root that would collide forever on retry. Every tmux step uses
+// 2>&1 so a failure at ANY step (not just new-session) keeps its reason (e.g.
+// "duplicate session: NAME") for runSessionScript to surface, instead of a bare
+// "exit status 1".
 //
 // send-keys targets use "=<name>:" — exact session match (the group's pane names
 // extend the root name, so prefix matching would be ambiguous) and the trailing
@@ -221,8 +223,8 @@ func buildPresetSessionScript(sessions []string, commands []string) string {
 	sendKeys := func(name, command string) []string {
 		target := shQuote("=" + name + ":")
 		return []string{
-			fmt.Sprintf("tmux send-keys -t %s -l -- %s", target, shQuote(command)),
-			fmt.Sprintf("tmux send-keys -t %s Enter", target),
+			fmt.Sprintf("tmux send-keys -t %s -l -- %s 2>&1", target, shQuote(command)),
+			fmt.Sprintf("tmux send-keys -t %s Enter 2>&1", target),
 		}
 	}
 
--- a/internal/tui/sessions.go
+++ b/internal/tui/sessions.go
@@ -111,13 +111,20 @@ func ensureSessionsLoaded(m *model, ref InstanceRef) {
 
 // runSessionScript runs a shell one-liner inside ref's container and folds a
 // non-zero exit into the returned error — matching the semantics of the old
-// local cmd.Run()/cmd.Output() call sites these commands were built on.
+// local cmd.Run()/cmd.Output() call sites these commands were built on. The
+// create scripts merge tmux's stderr into stdout (2>&1) so a captured reason
+// (e.g. "duplicate session: NAME") rides along here instead of being lost to a
+// bare "exit status 1" — the single biggest diagnosability win for the TUI,
+// which otherwise swallowed why a session failed to create.
 func runSessionScript(ref InstanceRef, script string) (string, error) {
 	out, code, err := runInstanceCommand(ref.Fleet, ref.Instance, []string{"sh", "-c", script})
 	if err != nil {
 		return out, err
 	}
 	if code != 0 {
+		if reason := strings.TrimSpace(out); reason != "" {
+			return out, fmt.Errorf("exit status %d: %s", code, reason)
+		}
 		return out, fmt.Errorf("exit status %d", code)
 	}
 	return out, nil
@@ -157,7 +164,10 @@ type sessionDeletedMsg struct {
 // path) and then creates a detached session inside the container.
 func createSessionCmd(ref InstanceRef, sessionName string) tea.Cmd {
 	return func() tea.Msg {
-		script := tmuxEnsureInstalled + fmt.Sprintf(`tmux new-session -d -s %s 2>/dev/null`, shQuote(sessionName))
+		// 2>&1 (not 2>/dev/null): on a name collision tmux exits 1 with
+		// "duplicate session: NAME" on stderr — keep it so runSessionScript can
+		// surface the reason instead of a bare "exit status 1".
+		script := tmuxEnsureInstalled + fmt.Sprintf(`tmux new-session -d -s %s 2>&1`, shQuote(sessionName))
 		if _, err := runSessionScript(ref, script); err != nil {
 			return sessionCreatedMsg{ref: ref, err: err}
 		}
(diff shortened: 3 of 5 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary** — Reviewed the diff (5 files, +215/−25). The fix is well-targeted and the layered defense is sound: `uniqueGroupName` avoids the collision in the common case, and the transactional script's `|| exit 1` guard means even a TOCTOU collision on the root fails safely (pre-existing session preserved, reason surfaced) rather than killing a live group. `go build` + the new/updated tests pass locally (`go test ./internal/tui/`). A few notes below — none blocking.

**Things I verified**
- The `2>&1` change is more than cosmetic. The local exec path (`runInstanceCommand`) uses `cmd.Output()`, which captures **stdout only**; tmux writes `duplicate session: …` to **stderr**, so even without `2>/dev/null` the reason would have been dropped. Merging stderr into stdout is what actually lets `runSessionScript` surface it. Good call, and the comment explains it well.
- `recordPresetGroupLayout` keys off `msg.groupID` (the uniquified name) and `msg.sessions[0] == GroupSessionName(sanitized, name)`, so the saved layout stays consistent with the real session names after suffixing — no orphaned layout. It's recorded only on success, so a torn-down partial group leaves nothing behind. 👍
- Cleanup is `;`-separated and ends in `exit 1`, so a swallowed `kill-session` ("can't find session") still yields a guaranteed non-zero exit with the failing step's reason intact in stdout.

**Minor — the reason is surfaced for `new-session` failures, but not `send-keys` failures** (inline comment on `sessions.go`)
The inner `send-keys` steps don't merge stderr, so if the failing step is a `send-keys` rather than a `new-session`, the user is back to a bare `exit status 1`. The headline case (duplicate session = a `new-session`) is fully covered, so this is low priority — but it's a gap relative to the PR's "surface the reason" goal.

**Very minor / optional — cleanup kills panes unconditionally**
The root is guarded so a pre-existing root is never killed, but the pane `kill-session`s run regardless of who created them. The comment calls the pane names "all unique to this attempt"; that holds only modulo a `randomHex(2)` (~65k space) collision with a pre-existing session, in which case cleanup would kill a foreign session. Practically negligible and not a regression — flagging only because the comment states it as a guarantee.

Nice, focused P0, and the deferred `nextSessionName` / poller items in the description look like the right things to leave for P1.

internal/tui/sessions.go
```diff
@@ -178,30 +188,69 @@ type presetSessionsCreatedMsg struct {
...
-			b.WriteString(" && ")
+	if len(sessions) == 0 {
+		return ""
+	}
+	root := sessions[0]
+
+	sendKeys := func(name, command string) []string {
+		target := shQuote("=" + name + ":")
+		return []string{
+			fmt.Sprintf("tmux send-keys -t %s -l -- %s", target, shQuote(command)),
```
BenjaminBenetti: 🔍 **CodeReview Agent**

These `send-keys` steps don't redirect stderr, unlike the `new-session` calls (`2>&1`). Because the local `runInstanceCommand` path uses `cmd.Output()` (stdout only), a failure *here* — rather than at a `new-session` — loses its reason and surfaces as a bare `exit status 1`, the exact symptom this PR fixes. Appending `2>&1` to both `send-keys` invocations would close the gap. Low priority: the headline "duplicate session" case is a `new-session`, so it's already covered.

BenjaminBenetti: Good catch — fixed in 4feb120. Both `send-keys` invocations now use `2>&1`, so a failure at a send-keys surfaces its reason via runSessionScript instead of a bare `exit status 1`. Verified against a real tmux that the pane command still runs.

BenjaminBenetti:
Thanks for the review — both notes addressed in 4feb120:

- **send-keys stderr (inline):** both `send-keys` steps now use `2>&1`, closing the gap so a send-keys failure surfaces its reason rather than a bare `exit status 1`. Re-verified end-to-end against a real tmux (the pane command still executes).
- **Cleanup-uniqueness comment (very minor):** reworded — the safety guarantee is now tied to `uniqueGroupName` having confirmed the whole `<inst>~<group>~*` namespace was free, rather than overclaiming "all names unique to this attempt".

`nextSessionName` and the self-correcting container-user probe remain deferred to P1 per the issue owner's call.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` from PR branch + TUI launches | ✅ Pass | Static binary builds clean; TUI renders, discovers the running instance |
| Provision a real instance (`fleet up`) | ✅ Pass | `repo/qa-inst` running, tmux 3.3a present in container |
| Single-session create (happy path) | ✅ Pass | `a` → name `dev` → "Session created"; appears in list |
| **Duplicate session → reason surfaced** | ✅ Pass | Status line: `Failed to create session: exit status 1: duplicate session: qa-inst~dev` (no longer a bare `exit status 1`) |
| Existing session preserved on failed create | ✅ Pass | Original `dev` untouched after the collision |
| Create layout preset (capture group) | ✅ Pass | Captured `qa-inst / dev` as preset `dev (1 pane)` |
| **Apply preset → root uniquified** | ✅ Pass | Preset `dev` collides with live `dev` group → mints `dev-2` instead of failing |
| Re-apply preset again (no forever-collision) | ✅ Pass | Third apply mints `dev-3`; `dev`/`dev-2` preserved — the persistent-failure bug is gone |
| Transactional script — clean 2-pane create | ✅ Pass | Generated script vs real tmux 3.3a: exit 0, root + pane both created |
| Transactional script — pre-existing root preserved | ✅ Pass | Root create fails `\|\| exit 1` before cleanup → leftover root **untouched**; reason `duplicate session` surfaced |
| Transactional script — mid-chain failure cleanup | ✅ Pass | Pane create fails mid-chain → cleanup tears down the root **this run created**; **no stranded root** |
| Open session as split (observation) | ⚠️ Note | Pre-existing, out-of-scope: opening `dev` attaches `dev`/`dev-2`/`dev-3` as one group (prefix grouping — the P1 item this PR explicitly defers). Does not affect the P0 fixes. |

**Overall: ✅ PASS — all 11 tested scenarios behaved as expected (1 informational, pre-existing/out-of-scope note).**

---

## How this was tested

All app execution was isolated in a throwaway **privileged Docker-in-Docker container** (`qa-211`) — its own `dockerd` + `devcontainer` CLI + tmux — so the QA never touched the host session. The PR branch (`9b3da6d`) was built into a static `fleet` binary and used to provision a **real** devcontainer instance (`repo/qa-inst`, debian-12 + tmux 3.3a) from the integration fixture. The TUI was driven inside the container via `tmux send-keys` / `capture-pane`, exactly as a user would.

## Detailed findings

### ✅ Headline fix — duplicate session now reports the real reason
1. `a` on `qa-inst` → typed `dev` → **"Session created"** (verified `dev` in the list).
2. `a` again → typed `dev` (same name) → status line showed:
   ```
   Failed to create session: exit status 1: duplicate session: qa-inst~dev
   ```
   This is the core fix (`2>&1` + `runSessionScript` folding the captured stderr into the error). Previously this surfaced only as a bare `exit status 1`. The original `dev` session was preserved.

### ✅ Layout-preset root uniquification (the "fails forever" bug)
1. Created a layout preset from the live `dev` group → saved as `dev (1 pane)`.
2. Applied it with an empty name (defaults to preset name `dev`). Because a `dev` group already exists, it auto-suffixed → **`Session dev-2 created (1 pane)`**.
3. Applied a third time → **`Session dev-3 created (1 pane)`**.
   Result: `dev`, `dev-2`, `dev-3` all live and intact (confirmed via `tmux ls` inside the container). Re-applying a preset never collides permanently anymore — a daemon restart is no longer needed because the fixed name is gone.

### ✅ Transactional preset script (validated against the instance's real tmux 3.3a)
The cleanup branch can't be triggered by ordinary UI actions (a user can't make tmux fail mid-chain), so I validated the **exact script shape `buildPresetSessionScript` generates** for a 2-pane preset directly against the instance's tmux:
- **Clean create:** `exit 0`; both `qa-inst~ok` and `qa-inst~ok~p1` created.
- **Pre-existing root (leftover from a prior failed apply):** root `new-session` fails → `|| exit 1` short-circuits **before** the cleanup block, so the pre-existing root is **never killed**; the captured reason `duplicate session: qa-inst~ex` rides along. `sessions after: qa-inst~ex` (preserved).
- **Mid-chain failure:** root created, then pane `new-session` collides → cleanup runs and kills the root **this run** created → `sessions after:` (empty). No stranded root to collide forever on retry.

### ⚠️ Out-of-scope observation (not a regression from this PR)
Opening the `dev` session as a split pane attached to `dev`, `dev-2`, and `dev-3` together as if one group — consistent with the prefix-matching limitation the PR explicitly **defers to P1** ("`nextSessionName` … matches a bare prefix against grouped cache names"). It does not affect any of the P0 behaviors verified above. Flagging it only for visibility.

---
_Container `qa-211` was torn down after testing._

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Re-review of `4feb120`** (synchronize). Re-ran the checks against the new head: `go build ./...` ✅, full `go test ./...` ✅, and I validated the generated script shape against a live tmux 3.3a on an isolated socket.

**Prior feedback — resolved ✅**
- The send-keys stderr gap is closed: both `send-keys` steps now end in `2>&1`, so a failure at a *send-keys* (not just `new-session`) surfaces its reason through `runSessionScript`. Confirmed on real tmux that the pane command still types and runs.
- The cleanup-safety comment was tightened, and the reasoning now holds end-to-end: `uniqueGroupName` rejects a desired name whenever *any* `<inst>~<group>~*` session exists (`parseGroupID` maps a pane name back to its group), so the whole `<inst>~<group>~*` namespace really is free before the run — the kills can only hit sessions this run created, never a live group.

**Verified on live tmux (3.3a)**
- A mid-chain failure tears down **both** the root and the already-created pane — the `=<name>:` `kill-session` target form resolves correctly, so nothing is stranded.
- A pre-existing root **survives** a colliding retry: the root's `|| exit 1` exits *before* the cleanup block, leaving the existing session untouched, and `duplicate session: <name>` rides out on stdout for the status line.

**One new minor note (non-blocking)**
On the cold-start path where tmux isn't installed yet, `TmuxEnsureInstalled` echoes `==> Installing tmux...` to unredirected stdout. Because the create scripts merge tmux stderr into that same stdout (`2>&1`) and `runSessionScript` folds the *entire* trimmed output into the error, a collision in that narrow window yields a multi-line `exit status 1: ==> Installing tmux...\n…\nduplicate session: <name>` — the install banner leads and partly buries the reason this PR works to surface. It's rare (a duplicate-on-preset almost always means tmux is already present), so just flagging it: if you want the status line to stay crisp, surfacing the last non-empty line of `out`, or routing the install banner to stderr→/dev/null, would close it.

Nice, tightly-scoped P0 — deferring `nextSessionName` and the poller self-correction to P1 reads as the right call.

BenjaminBenetti:
Followed up on the QA "opening dev attaches dev/dev-2/dev-3 as one group" observation — it's not just cosmetic, and it was caused by this PR.

**Why:** the `uniqueGroupName` suffixing (`dev → dev-2 → dev-3`) created group IDs where one is a *prefix* of another, and several group ops match membership by raw prefix (`HasPrefix(name, "<inst>~<groupID>")`) — `splitpane.go:690` (open), `sessions.go:340` (delete), `:299` (rename), `:364` (Contains). So `<inst>~dev` prefix-matches `<inst>~dev-2`: opening `dev` grabs all three, and **deleting group `dev` would also kill `dev-2`/`dev-3`** — a data-loss path my own naming introduced.

**Fix (adeb135):** mint the layout group root under a **fresh random group id** (`randomHex(3)`), exactly like split-created groups already do. A random id is collision-proof on re-apply *and* can never be a prefix of a sibling, so it doesn't feed the prefix matching the wrong thing. Dropped `uniqueGroupName` + its test.

Tradeoff: a layout group now shows a hex id (consistent with split groups) instead of the preset name. Making readable/typed layout-group names safe requires boundary-aware group matching — that belongs with the deferred P1 prefix work (`nextSessionName` et al.), owner's call.

The three P0 behaviors are unchanged: reason surfaced on duplicate, no forever-collision on re-apply, transactional cleanup on partial failure.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the P0 set on `fix/session-create-exit1-p0` (head `adeb135`). Checked out the branch locally — `go build ./...` and `go test ./internal/tui/` are green. The root-cause analysis is correct and the fix is well-targeted. Findings below; nothing blocking.

### What looks right
- **Transactional preset script is sound.** Creating the root on its own with `... 2>&1 || exit 1` *before* the `{ inner } || { kills; exit 1; }` group is the key correctness property: a collision on a pre-existing root exits before the cleanup ever runs, so a live session is never killed. The cleanup `kill-session -t '=<name>:'` calls target exact sessions, and since the root is now minted under a fresh `randomHex(3)` group id the whole `<inst>~<group>~*` namespace belongs to this run — kills can only hit sessions this run created. The empty-`inner` short-circuit (single bare session → no `|| exit 1`/cleanup scaffolding) is handled and tested.
- **Random group id is the right P0 call.** It side-steps both the forever-collision (container-resident session under a fixed name) and the prefix-match foot-gun (`HasPrefix(name, "<inst>~dev")` matching `<inst>~dev-2` in the group open/rename/delete ops) that the `dev-2` suffix approach would have fed. Consistent with how split groups (`NewGroupRootSessionName`) already mint ids.
- **`send-keys ... 2>&1`** closes the gap from the earlier review round — a failure at a `send-keys` step (not just `new-session`) now surfaces its reason rather than regressing to a bare `exit status 1`.
- **Test coverage is good** — `TestRunSessionScriptSurfacesReason` / `…BareExitWhenSilent` / `…PropagatesTransportError` cover the three branches of the fold, and `…SinglePaneNoCleanup` pins the empty-`inner` shape. `uniqueGroupName` and its test are fully removed (no dangling refs).

### Notes (non-blocking)

1. **User-visible: layout groups now surface under an opaque hex id.** With the random `groupID`, the success toast becomes `Session 3f9ac1 created (…)` and the group renders in the session list under the hex rather than the readable preset name (e.g. `dev`). This is an intentional tradeoff (documented as deferred-to-P1 boundary-aware matching) and matches split-group behavior — but it's a real regression in the common case that the PR title/description frames around the *error* path, not this side effect. Flagging only so it's a conscious sign-off, not a surprise.

2. **PR description is now stale.** Fix item #2 ("Uniquify the layout root — `dev → dev-2 → …` against the live session list") and the Testing bullet listing `TestUniqueGroupName` describe the approach the final commit (`adeb135`) replaced with the random-id strategy and deleted. Worth updating the body so the description matches the merged diff.

3. **(Minor) Multi-line reason can ride into the single-line status.** `runSessionScript` folds the *entire* trimmed `out` into the error. In practice the failing tmux command emits one line (`duplicate session: NAME`) and the earlier successful steps emit nothing, so this is clean today. The only noisy path is when `TmuxEnsureInstalled` echoes (e.g. `==> Installing tmux...` / `ERROR: failed to install tmux`) precede a failure — that text would be merged into the TUI status line. Left an inline suggestion to take the last non-empty line for display robustness. Low priority.

Nice work — the transactional shape and the "leave a pre-existing root untouched" property are exactly right.

internal/tui/sessions.go
```diff
@@ -111,13 +111,20 @@ func ensureSessionsLoaded(m *model, ref InstanceRef) {
...
+// bare "exit status 1" — the single biggest diagnosability win for the TUI,
+// which otherwise swallowed why a session failed to create.
 func runSessionScript(ref InstanceRef, script string) (string, error) {
 	out, code, err := runInstanceCommand(ref.Fleet, ref.Instance, []string{"sh", "-c", script})
 	if err != nil {
 		return out, err
 	}
 	if code != 0 {
+		if reason := strings.TrimSpace(out); reason != "" {
+			return out, fmt.Errorf("exit status %d: %s", code, reason)
```
BenjaminBenetti, line 131: 🔍 **CodeReview Agent**

Minor robustness nit: this folds the whole trimmed `out` into the error, and since the create scripts now merge stderr via `2>&1`, on the rare path where `TmuxEnsureInstalled` echoes (`==> Installing tmux...`, `ERROR: failed to install tmux`) precede a failure, that text rides into the single-line TUI status. The common `duplicate session: NAME` case is one line, so this is low-priority — but consider showing just the last non-empty line, e.g.:

```go
if reason := strings.TrimSpace(out); reason != "" {
    if i := strings.LastIndexByte(reason, '\n'); i >= 0 {
        reason = strings.TrimSpace(reason[i+1:])
    }
    return out, fmt.Errorf("exit status %d: %s", code, reason)
}
```

(Keeps the full `out` as the returned string; trims only what's interpolated into the message.)

BenjaminBenetti: Fixed in 62ab80c — runSessionScript now surfaces only the last non-empty output line, so the `==> Installing tmux...` preamble is dropped while a genuine `ERROR: failed to install tmux` (the last line in that case) still shows. Added a unit test for it.

BenjaminBenetti:
Heads-up: the previous push (random group id) regressed integration test `290_tui_layout_preset`, which types `fromtpl` and asserts the session is `alpha~fromtpl` — a real typed-name contract I'd traded away.

Fixed in `62ab80c`: `groupIDFor` honors the preset/typed name **when its group id is free**, and only falls back to a random id when that group already exists. So:
- First apply (free name) → readable group `fromtpl` — test 290 green, common-case UX preserved.
- Re-apply / leftover (name taken) → random id — no forever-collision, and no auto-minted `dev-2` prefix sibling for the prefix-based group ops to mismatch.

Also addressed the install-preamble nit in the same commit (last-non-empty-line). CI re-running.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| `go build ./...` | ✅ Pass | Compiles clean (exit 0) |
| TUI launch & fleet/instance render | ✅ Pass | `qa/inst1` renders, no startup errors |
| Single-session create (`alpha`) | ✅ Pass | Created `inst1~alpha`; appears in TUI tree |
| Duplicate create surfaces reason (**fix #1**) | ✅ Pass | Status line: `…exit status 1: duplicate session: inst1~alpha` — no longer a bare `exit status 1` |
| Layout-preset create (**fix #3** transactional, success) | ✅ Pass | `dev (2 panes)` → root `inst1~dev` + pane `inst1~dev~576f`; pane-2 startup cmd ran |
| Re-apply preset → unique name (**fix #2**) | ✅ Pass | `dev` → `dev-2` → `dev-3` on repeat applies; never collides "forever" |
| Pre-existing groups untouched on re-apply (**fix #3** safety) | ✅ Pass | All prior roots **and** panes intact after re-applies; no cross-group kills / stranded roots |
| Graceful quit | ✅ Pass | `q` exits the TUI cleanly, no crash |
| Cleanup-on-mid-chain-failure branch (fix #3) | ⏭️ Skipped | Couldn't inject a mid-chain tmux failure through the UI; success path + "pre-existing root never touched" verified instead |

**Overall: ✅ PASS — 8 of 8 executed scenarios passing (1 path not forceable via the UI, noted below).**

---

### How it was tested (manual, behavior-level — not unit tests)

All app execution was isolated in a throwaway **privileged DinD container** (`qa-211`), never in the QA runner's own session. Inside it I built the **PR branch** binary (`4feb120`, static), brought up a real **devcontainer-backed instance** (`devcontainer up`, the same `devcontainer exec --workspace-folder …` path the TUI uses for session creation), seeded a fleet `qa` / instance `inst1` with a 2-pane layout preset `dev`, then drove the **actual `fleet` TUI** with tmux `send-keys` / `capture-pane`, verifying each result against the instance's real tmux (`tmux list-sessions`). Container torn down afterward.

### Detailed findings

**1. Reason surfacing on a name collision (fix #1) — ✅**
- Created session `alpha` (→ `inst1~alpha`, confirmed in the instance's tmux), then opened *New session* again and submitted the same name `alpha`.
- TUI status line showed: **`Failed to create session: exit status 1: duplicate session: inst1~alpha`**.
- Pre-fix this was an opaque `exit status 1`. The captured `duplicate session: <name>` reason now rides through `runSessionScript` (the `2>&1` + folded output). ✔️

**2. Layout-preset apply, success path (fix #3) — ✅**
- *New session* → `[tab]` to select template `[ dev ]` → submit (empty name → defaults to preset name).
- Status: `Session dev created (2 panes)`. Instance tmux showed the group **root `inst1~dev` + pane `inst1~dev~576f`**.
- The pane-2 startup command from the preset (`echo hello-pane2`) actually executed in that pane (`hello-pane2` printed), so the transactional `{ … } || { kill…; exit 1 }` shape and the `send-keys … 2>&1` changes don't break normal pane command typing. ✔️

**3. Re-applying the same preset → unique root name (fix #2) — ✅**
- Re-applied `dev` twice more. Results: **`dev-2`** then **`dev-3`** (`Session dev-2/dev-3 created (2 panes)`), each minting its own root + pane (`inst1~dev-2~9d7c`, `inst1~dev-3~7c48`).
- This is the heart of the fix: the fixed preset-derived root name no longer collides on `new-session` and "fail forever" — `uniqueGroupName` auto-increments against the live session list while keeping the readable name. ✔️

**4. Safety: pre-existing groups never touched (fix #3) — ✅**
- After all re-applies, the full instance session set was intact and non-overlapping:
  `inst1~alpha`, `inst1~dev (+~576f)`, `inst1~dev-2 (+~9d7c)`, `inst1~dev-3 (+~7c48)`.
- No earlier root/pane was killed and no stranded/duplicate root appeared — confirming the cleanup can only ever hit sessions the current run created.

**5. Smoke / regression — ✅**
- TUI launched cleanly, navigation (`j`/`space`/`tab`), the session poller reflected created sessions, and `q` quit without a crash.

### Notes / transparency
- **Cleanup-on-failure branch (fix #3) not directly exercised:** forcing a *mid-chain* tmux failure (so the group's own root+panes get torn down) isn't reachable through normal TUI input. I verified the success path and the "pre-existing root is never touched" guarantee instead; the PR's own description covers the mid-chain teardown against live tmux.
- **Container-user probe (out of scope):** my first instance image (`ubuntu:24.04`, default `ubuntu` user) initially created sessions as `root` while the poller probed user `ubuntu`, so a created session didn't list. This is the **pre-existing P1 issue the PR explicitly defers** (poller container-user probe self-correction), not a regression from this change. I re-ran with `remoteUser: ubuntu` so create + poll agreed, and everything behaved as expected.

_Tested against head `4feb12046939b0a679926d8c7edddd704ce8d529`._

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` binary | ✅ Pass | `go build ./cmd/fleet` clean on PR head `adeb135` |
| Launch & main menu renders | ✅ Pass | Armada + `itest-fleet/alpha` render, no errors |
| Single-session create | ✅ Pass | "Session created"; `○ mysession` row; tmux session `alpha~mysession` exists |
| **Duplicate create surfaces reason (P0 #1)** | ✅ Pass | Status now reads `Failed to create session: exit status 1: duplicate session: alpha~mysession` — reason no longer swallowed |
| Layout preset capture + save | ✅ Pass | Preset `mysession (1 pane)` saved; `layoutPresets` persisted to `state.json` |
| Templated session create (apply #1) | ✅ Pass | Created group `23f66d` |
| **Layout preset re-apply ×3 (P0 #2)** | ✅ Pass | `23f66d` / `c6d6f3` / `eee6f4` — 3 distinct groups, **no forever "exit status 1"** |
| Preset pane command executes (P0 #3 happy path) | ✅ Pass | `echo preset-pane-ready` ran inside the new pane |
| Session delete + clean quit (sanity) | ✅ Pass | Deleted `alpha~23f66d`; TUI quit cleanly |
| Templated session naming | ⚠️ Note | Typed name is **ignored** for templates; group shows a random hex, not a readable name (see below) — non-blocking, deliberate trade-off |

**Overall: ✅ PASS — all P0 acceptance criteria met (9/9 functional scenarios pass); 1 non-blocking UX note.**

---

### Environment
- QA image `ghcr.io/benjaminbenetti/fleet-man/qa:latest` (Docker-in-Docker, `--privileged`), Go 1.25.8.
- Real devcontainer instance `itest-fleet/alpha` brought up via `fleet up` (image `mcr.microsoft.com/devcontainers/base:debian-12`, **tmux 3.3a** inside — matches the version called out in the PR).
- TUI driven headlessly via `tmux send-keys` / `capture-pane`, exactly as a user would.

### Detailed findings

**P0 #1 — failure reason is surfaced (the headline fix). ✅**
1. Created session `mysession` (succeeds → `alpha~mysession`).
2. Opened **New session** again, typed the same name `mysession`, Enter.
3. Status line: `Failed to create session: exit status 1: duplicate session: alpha~mysession`.
   The tmux stderr (`duplicate session: …`) now rides through `runSessionScript` instead of being thrown away by `2>/dev/null`. Pre-fix this was a bare `exit status 1`. Confirmed working.

**P0 #2 — layout root is uniquified, so re-apply never collides (the "stuck forever" fix). ✅**
1. Captured a preset from `mysession`, set pane 1 cmd `echo preset-pane-ready`, saved → `Layouts (1)`.
2. Created a session from the template **three times** in a row (New session → `Tab` to select `[ mysession ]` → Enter).
3. Each apply produced a **distinct** group and a "Session &lt;id&gt; created (1 pane)" status:
   ```
   alpha~23f66d   alpha~c6d6f3   alpha~eee6f4
   ```
   No repeated "exit status 1", nothing stranded. The old fixed-name collision-forever behaviour is gone.

**P0 #3 — transactional preset script (happy path). ✅**
   `tmux capture-pane` of `alpha~23f66d` shows the pane command actually executed:
   ```
   vscode ➜ /workspaces/itest-fleet (main) $ echo preset-pane-ready
   preset-pane-ready
   ```
   Root + pane send-keys chain ran cleanly; no partial/stranded sessions left behind.

**Sanity.** Session delete (`d` → `y`) removed `alpha~23f66d` from both the TUI list and the container; `q` quit the TUI cleanly. No regressions observed in the basic session flows.

### ⚠️ Non-blocking observation — templated session naming / stale PR description
With the final commit (`adeb135` "mint layout group under a random id, not a readable suffix"), creating a session from a template:
- **ignores the `Name:` field** — the New session dialog still shows and accepts a name (I typed `fromtpl`), but it is silently discarded; the group is minted under a random hex id (`23f66d`, …) and appears that way in the list.
- means the **PR description's fix #2 wording is now stale**: it says *"`dev → dev-2 …`"* and *"The readable preset name is kept in the common case"*, but the shipped code always uses a random hex (no readable name). The code comment correctly documents this and defers readable/boundary-aware names to **P1**.

This is a deliberate trade-off to kill the P0 collision bug, not a functional defect, so it does not block the verdict — flagging so the description can be reconciled and the discarded-name UX considered for the P1 follow-up. (Note: integration test `290_tui_layout_preset.sh` still asserts the old readable name — `Session fromtpl created (1 pane)`, `○ fromtpl`, `=alpha~fromtpl:`, and `grep '"fromtpl"'` in `state.json` — and was not updated in this PR, so it will likely fail in the integration CI against this change. `291` only captures a preset across a restart and is unaffected.)

BenjaminBenetti:
CI green on `62ab80c` (all 9 checks, including the full integration suite — test 290 passes again).

Re-running QA to confirm the final behavior, since it changed since the last pass:
- apply preset `dev` (free) → group `dev` (readable)
- re-apply `dev` → **random id** (not `dev-2`), so opening/deleting `dev` no longer touches the re-applied group (the conflation you flagged is gone)

/qa-manager test this PR.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary** — Reviewed the P0 fix for opaque / permanent layout-preset session-create failures. This is a clean, well-scoped, and well-tested change. Verified locally: the patch applies on `main`, and `go build ./...`, `go vet ./internal/tui/`, and the full `internal/tui` test suite all pass. I also dumped the real generated preset script to confirm its shape.

**What holds up well**
- **Transactional preset script is correct.** `tmuxEnsureInstalled` ends in `; }; `, so `tmux new-session … 2>&1 || exit 1` binds the guard to *only* the root create — a pre-existing-root collision exits **before** the cleanup block runs, so an existing session is never killed. The `{ inner; } || { kills…; exit 1; }` cleanup targets exact `=name:` sessions with `2>/dev/null`, and since `groupIDFor` hands out a fresh group id, those kills can only ever hit sessions *this run* created. The doc-comment reasoning checks out against the generated script.
- **`2>&1` + `lastNonEmptyLine`** is the right way to surface `duplicate session: …` instead of a bare `exit status 1`, and the bare-exit / transport-error passthrough paths are preserved and explicitly tested.
- **`groupIDFor`** correctly prefix-scopes via `parseGroupID` (other instances' sessions are ignored) and deliberately avoids the `dev-2` suffix that would foul prefix-based group ops — sensible to defer that to P1.
- **Good defense-in-depth:** even if `groupIDFor` is fooled by a stale/empty runtime cache, the `|| exit 1` + self-cleanup means a failure surfaces loudly and strands nothing, so the "stuck forever" recurrence can't come back through that path.

**Minor, non-blocking notes** (left inline):
1. The `runSessionScript` doc comment slightly overstates which line is surfaced on a tmux *install* failure.
2. `groupIDFor`'s collision check is only as strong as the runtime cache that feeds it — worth noting the first-re-apply edge.

No blocking issues. The live-tmux verification plus the focused unit tests give good confidence here. 👍

internal/tui/sessions.go
```diff
@@ -111,18 +111,40 @@ func ensureSessionsLoaded(m *model, ref InstanceRef) {
...
 // non-zero exit into the returned error — matching the semantics of the old
-// local cmd.Run()/cmd.Output() call sites these commands were built on.
+// local cmd.Run()/cmd.Output() call sites these commands were built on. The
+// create scripts merge tmux's stderr into stdout (2>&1) so a captured reason
+// (e.g. "duplicate session: NAME") rides along here instead of being lost to a
+// bare "exit status 1" — the single biggest diagnosability win for the TUI,
+// which otherwise swallowed why a session failed to create. Only the LAST
+// non-empty line is surfaced: that's the actual failing command's message,
+// dropping any TmuxEnsureInstalled "==> Installing tmux..." preamble (while
+// still surfacing a real "ERROR: failed to install tmux", which is the last
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Minor doc nit (behavior is fine): this says the last line "still surfac[es] a real `ERROR: failed to install tmux`". But `tmuxEnsureInstalled` ends with `; }; ` (not `&& exit`), so after the install block echoes that error, control falls through to `tmux new-session …`, which then fails with something like `sh: tmux: not found` (exit 127). That not-found message is the *last* non-empty line, so it — not `ERROR: failed to install tmux` — is what `lastNonEmptyLine` actually returns. Still a useful reason; the comment just slightly overstates which message wins.

internal/tui/dialog_session.go
```diff
@@ -123,15 +123,28 @@ func (fleetPage *fleetPage) saveCreateSession(m *model) tea.Cmd {
...
+		// Keep the readable preset/typed name as the group id when it's free,
+		// else fall back to a fresh random id. A preset-derived root name
+		// re-collides forever on new-session once that group exists — the
+		// reported "stuck, restart doesn't help" bug, since the session lives in
+		// the container. Falling back to random (rather than auto-suffixing to
+		// dev-2) avoids minting a sibling whose name a prefix-based group op would
+		// match against "dev"; readable names for re-applies need boundary-aware
+		// matching first — the deferred P1 prefix work. Load the runtime list so
+		// the free check sees what's actually live.
+		ensureSessionsLoaded(m, ref)
```
BenjaminBenetti, line 135: 🔍 **CodeReview Agent**

`groupIDFor`'s collision avoidance is only as good as what `ensureSessionsLoaded` → `runtimeSessions(ref)` returns here. For an instance whose row was never expanded and whose ~1s runtime poll hasn't cached sessions yet, `Sessions(ref)` is `nil`, so `groupIDFor` honors the desired name verbatim even when `<inst>~<name>` is already live in the container. The first apply then fails with the (now-visible) `duplicate session` error instead of transparently picking a fresh id.

It self-heals — the transactional script strands nothing, and the next attempt (after the cache populates) falls back to a random id — so this is *not* the old "stuck forever" bug, just a possible one-shot duplicate-error on first apply/re-apply. Consistent with the P1 deferral; flagging for awareness.

internal/tui/sessions.go
BenjaminBenetti: Correct — fixed the comment in a3a7ad8. The install block ends in `; }; ` so control falls through to `tmux new-session`, whose "tmux: not found" becomes the last line; the comment now says that instead of claiming the install ERROR line is surfaced. Behavior unchanged (last line is still a useful reason).

internal/tui/dialog_session.go
BenjaminBenetti: Accurate, and intentionally accepted for this P0. The server polls tmux for every *running* instance (~1s) and pushes it to m.runtime regardless of row expansion, so the cold window is narrow (just after TUI connect / instance start). When it does hit, the failure is graceful, not the original bug: the transactional root `... 2>&1 || exit 1` leaves the live session untouched and surfaces `duplicate session: <inst>~<name>`, and the next attempt (cache now warm) falls back to a random id. Fully closing the window needs either a synchronous tmux list at create time or detect-collision-and-retry-with-random — both beyond this focused P0; noting it alongside the deferred boundary-aware-matching P1 item.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & binary run | ✅ Pass | `go build ./cmd/fleet` clean; `fleet --help` runs (Go 1.25.11) |
| TUI launch / render / quit | ✅ Pass | Dependency check → Armada main view render → navigate → quit, no panic |
| Single session create — reason surfaced | ✅ Pass | Collision now shows `duplicate session: NAME`, not bare `exit status 1` |
| Layout preset — clean create | ✅ Pass | Root + pane sessions created, pane commands typed |
| Preset root-collision preserves existing root | ✅ Pass | Fails loud; pre-existing root + its window left untouched; no pane leaked |
| Preset mid-chain failure cleanup | ✅ Pass | Root this run created is torn down — nothing stranded to collide forever |
| Collision no longer permanent (groupIDFor) | ✅ Pass | Re-apply mints a fresh random id and succeeds instead of colliding forever |

**Overall: ✅ PASS — 7 of 7 scenarios passing.**

---

### Method & environment

Tested inside a throwaway Docker container (`golang:1.25-bookworm` + `tmux 3.3a` — the same tmux version cited in the PR's own testing), workspace mounted read-only.

The TUI binary was launched and driven as a user (dependency-check screen → main *Armada* view → create-instance guard → quit) and behaves correctly. The **in-TUI session-create flow itself requires a provisioned instance** (devcontainer CLI + Docker-in-Docker), which the minimal image doesn't carry — so the session-create behavior (the entirety of this PR's diff: `sessions.go`, `dialog_session.go`, `groups.go`) was verified by executing the **exact shell scripts the changed Go code generates** (`createSessionCmd` / `buildPresetSessionScript`) through a faithful port of `runSessionScript` against a real tmux, on an isolated socket. This mirrors the PR author's own verification method and directly exercises the changed logic. **11/11** behavioral assertions passed.

### Detailed findings

**1. Build & run** — `go build -o /tmp/fleet ./cmd/fleet` succeeded and `fleet --help` runs. (Needed `-buildvcs=false` only because the read-only bind mount trips git's dubious-ownership check inside the container — an environment quirk, not a code issue.)

**2. TUI launch** — Renders the Dependency Check screen (`found tmux`; `not found devcontainer CLI`, expected in this minimal image), `[enter]` advances to the `Armada [ local ]` view showing *"No instances. Press 'a'…"*. Navigation and quit work; no panic/stack trace observed.

**3. Single session create — reason surfaced** (the headline bug). Creating a duplicate session name:
```
OLD (2>/dev/null) surfaced : 'exit status 1'
NEW (2>&1 + fold) surfaced : 'exit status 1: duplicate session: myinst~session-2'
```
✅ The real reason now rides through to the status line instead of being swallowed.

**4. Layout preset — clean create.** Script created both `inst~dev` (root) and `inst~dev~ab` (pane); pane startup commands were sent. Live sessions after: `inst~dev inst~dev~ab`.

**5. Preset root-collision preserves existing root.** With a pre-existing `inst~dev` (window renamed `ORIGINAL`), re-applying the preset:
- Surfaced `exit status 1: duplicate session: inst~dev` (loud, not bare).
- Pre-existing root **preserved** — window still `ORIGINAL` after the failed apply.
- No pane session created → the `|| exit 1` short-circuited before cleanup, so the live root was never killed. ✅

**6. Preset mid-chain failure cleanup.** Forced a later step to fail (a colliding pane name, standing in for a leftover from a prior partial apply) while the root was freshly created this run:
- Surfaced `exit status 1: duplicate session: inst~dev~ab`.
- The root this run created (`inst~dev`) was **torn down** by the cleanup branch — no stranded root left to collide forever on retry. ✅

**7. Collision is no longer permanent.** With a leftover `inst~dev` sitting in the container:
- OLD behavior (fixed readable root) → `duplicate session: inst~dev` on every retry (the *"stuck, restart doesn't help"* bug).
- NEW behavior (`groupIDFor` mints a fresh random id when the readable name is taken) → re-apply under e.g. `inst~a1b2c3` **succeeds**, coexisting with the leftover. ✅

### Notes
- The PR description still describes the older `dev → dev-2` auto-suffix approach; the merged commits changed this to *"honor the free readable name, else fall back to a random id"* (`groupIDFor`). The QA above reflects the **actual** shipped behavior, which is correct and avoids the prefix-matching hazard the description's deferred-P1 note calls out.
- No regressions observed. The TUI startup/quit/dependency paths are outside this PR's diff (confirmed: only `internal/tui/{sessions,dialog_session,groups}.go` change), so the headless quit-code nuance seen during testing is pre-existing and unrelated.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Re-review of `a3a7ad8`** (synchronize, `62ab80c → a3a7ad8`).

The triggering push is **docs-only**: it corrects the `runSessionScript` comment about install failures. I verified the new wording is accurate — when `TmuxEnsureInstalled` can't install tmux, the snippet falls through to `tmux new-session`, whose `… 2>&1` merges the shell's `tmux: not found` onto stdout, making *that* the last non-empty line `lastNonEmptyLine` surfaces (not `ERROR: failed to install tmux`, as the old comment claimed). Confirmed empirically with `dash`. Good catch — the previous comment was wrong.

Re-ran on the head commit in a fresh worktree: `go build ./...` ✅ and `go test ./internal/tui/` ✅.

**On the P0 set as a whole** — solid and well-scoped. The transactional preset script (root via `… 2>&1 || exit 1` so a pre-existing root is never killed; inner steps wrapped in `{ … } || { kill-by-exact-target 2>/dev/null; exit 1; }`) is correct, and the kills are silenced + exact-target so they can only hit sessions this run created. `groupIDFor` cleanly threads the needle between the earlier `dev-2` prefix-match regression and the typed-name contract (integration test 290): a free name is honored, a taken name falls back to a random id. Test coverage for the new behavior is good.

A few **minor, non-blocking** notes (all arguably part of the already-deferred P1 boundary-matching work):

1. **`groupIDFor` doesn't protect names containing the group separator `~`.** `SanitizeSessionName` rewrites `.` `:` `/` but not `~`. A name like `a~b` becomes root `inst~a~b`, which `parseGroupID` reads as group **`a`** — so `gid == desired` (`"a~b"`) never matches the live session, `groupIDFor` returns `desired` unchanged, and a re-apply re-collides on the fixed root. The `2>&1` change makes this *diagnosable* rather than silently-stuck, so it isn't a P0 regression, but the uniquification quietly doesn't cover `~`-bearing names. (Inline note on the function.)

2. **Status toast shows the random fallback id.** When `groupIDFor` falls back, `m.message` reads `Creating session <hex> from <preset>…` — the user typed/saw a readable name but the toast shows a hex id. Cosmetic only; persisted group state stays consistent. Consider surfacing `preset.Name` there instead.

3. **Doc nit:** the PR description still references `TestUniqueGroupName`; the test now in the tree is `TestGroupIDFor` (renamed during iteration).

Nothing here blocks merge — the fix does what it claims and the diagnosability win is real.

internal/tui/groups.go
```diff
@@ -123,6 +123,24 @@ func NewGroupPaneSessionName(sanitizedInstance, groupID string) string {
...
+// readable preset/typed name; honoring it keeps the group named that on first
+// apply, but a re-apply (or a leftover from a prior failed apply) falls back to
+// a random id so the root never collides forever on `tmux new-session` — the
+// reported "stuck, a restart doesn't help" bug, since the session lives in the
+// container. It deliberately does NOT auto-suffix (dev-2): that would prefix-
+// match a sibling group ("dev") in the prefix-based group ops (open/rename/
+// delete) — readable names for re-applies need boundary-aware matching first.
+func groupIDFor(sanitizedInstance, desired string, existing []tmuxSession) string {
+	for _, s := range existing {
+		if gid, ok := parseGroupID(sanitizedInstance, s.Name); ok && gid == desired {
```
BenjaminBenetti, line 137: 🔍 **CodeReview Agent**

Edge case: `SanitizeSessionName` doesn't strip the group separator `~`, so a `desired` containing `~` (a typed name like `a~b`, or a `~`-bearing preset name) defeats this check. The live root would be `inst~a~b`, which `parseGroupID` parses as group **`a`** — so `gid == desired` (`"a~b"`) never matches, `groupIDFor` returns `desired` unchanged, and a re-apply re-collides on the fixed root. It's now *diagnosable* thanks to `2>&1`, but not actually uniquified for this input. If preset/typed names aren't guaranteed `~`-free, consider stripping/replacing `~` in `desired` (or comparing against the parsed group id) so the fallback still triggers.

PR #212, opened 2026-06-24
fix(devcontainer): resolve containerUser from the authoritative remoteUser

BenjaminBenetti:
## The bug

In a fleet, tmux **sessions can't be opened** — the TUI reports "session created" but nothing shows up, and automation agents fail to launch (`tmux: Permission denied`, `~/.augment` write `exit 1`).

Root cause: fleet creates/attaches tmux sessions via `devcontainer exec` (which runs as the devcontainer **`remoteUser`**), but lists/captures/automation-launches them via `docker exec -u $(containerUser)`. `containerUser()` *guessed* the user from `/tmp/tmux-[1-9]*/` socket ownership — and that guess is wrong whenever a **non-fleet user holds a tmux socket**.

Confirmed live: the affected container had two tmux sockets —

```
/tmp/tmux-1000  ->  vscode   ← fleet's sessions live here
/tmp/tmux-1654  ->  app      ← image's own user, empty
```

`devcontainer exec` runs as `vscode`, so sessions land in `/tmp/tmux-1000`. But `containerUser()` latched onto `app` (its socket existed at the first probe; the answer is cached permanently), so the poll ran `docker exec -u app`, read `/tmp/tmux-1654` (empty), and showed nothing. tmux sockets are **per-uid**, so create and list landing on different users never see each other.

## The fix

Make `containerUser()` **authoritative** instead of a guess: resolve the user `devcontainer exec` itself uses, so both transports share one tmux server. The fast `docker exec` transport (poller CPU fix #182) is untouched — only *which user* it targets changes. Resolution is cached by containerID, cheapest-and-surest first:

1. **`remoteUser` from the `devcontainer.metadata` container label** — pure `docker inspect`, **no Node**, safe on the ~1s poll path (runs only on a cache miss). Effective user = last non-empty `remoteUser` in the merged-metadata array, falling back to last `containerUser`.
2. **`devcontainer exec --default-user-env-probe none id -un`** — ground truth for a container whose metadata names no user (one cached Node call, never on the steady-state path; env probe disabled so the username is all that's on stdout).
3. **the legacy socket/home-ownership heuristic** — last resort only.

Also fixes the inverse case the old heuristic mishandled: a `remoteUser: root` container, where the old comment's "fleet's sessions never live under root" assumption was simply false (create as root → `/tmp/tmux-0/`, which the heuristic explicitly skipped).

## Why not the alternatives

- *Swap create/attach onto `docker exec -u containerUser`* — would drag the interactive shells onto a transport that loses the login-shell PATH, the auto-TTY, and the workspace cwd `devcontainer exec` gives for free (behind ~15 call sites). Big blast radius, real regression surface.
- *Put `devcontainer exec` on the list path* — reintroduces the Node cold-start CPU burn that #182 deliberately removed.

This keeps both transports as-is and just makes them agree on the user.

## Test

`TestRemoteUserFromMetadataJSON` covers the parse against a real label sample plus edge cases (last-wins merge, `containerUser` fallback, malformed/empty). Full `go build ./...` + backend/server/agentdetect suites pass.

### DIFF
--- a/internal/backend/devcontainer/devcontainer_backend.go
+++ b/internal/backend/devcontainer/devcontainer_backend.go
@@ -22,12 +22,21 @@ type DevcontainerBackend struct {
 	verbose bool
 
 	userCache   map[string]string
+	userMissAt  map[string]time.Time // last failed user resolution, to throttle re-resolves
 	userCacheMu sync.Mutex
 }
 
+// userResolveRetryInterval throttles re-resolving a container whose user
+// couldn't be determined, so a persistently-unresolvable container can't make
+// the per-second poll re-spawn the resolution subprocesses every tick.
+const userResolveRetryInterval = 15 * time.Second
+
 // New creates a new DevcontainerBackend.
 func New(opts ...Option) *DevcontainerBackend {
-	devcontainerBackend := &DevcontainerBackend{userCache: make(map[string]string)}
+	devcontainerBackend := &DevcontainerBackend{
+		userCache:  make(map[string]string),
+		userMissAt: make(map[string]time.Time),
+	}
 	for _, opt := range opts {
 		opt(devcontainerBackend)
 	}
@@ -39,26 +48,36 @@ func New(opts ...Option) *DevcontainerBackend {
 // they operate on the SAME tmux server as the `devcontainer exec` paths
 // (create/attach). That user is the devcontainer remoteUser — see
 // resolveContainerUser. The answer is cached to avoid re-resolving on every
-// poll. Only a non-empty answer is cached: an empty result (a just-started or
-// gone container, or a transient probe failure) must be re-resolved next call
-// so it self-heals. Caching "" would, now that the backend is reused across
-// poll ticks, pin every later exec to the wrong user forever.
+// poll. Only a non-empty answer is cached permanently: an empty result (a
+// just-started or gone container, or a transient probe failure) must be
+// re-resolved later so it self-heals, but re-resolution is throttled to
+// userResolveRetryInterval so a stuck container can't re-spawn the resolution
+// subprocesses every tick. Caching "" permanently would, now that the backend
+// is reused across poll ticks, pin every later exec to the wrong user forever.
 func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string) string {
 	devcontainerBackend.userCacheMu.Lock()
 	if cached, ok := devcontainerBackend.userCache[containerID]; ok {
 		devcontainerBackend.userCacheMu.Unlock()
 		return cached
 	}
+	if missAt, ok := devcontainerBackend.userMissAt[containerID]; ok && time.Since(missAt) < userResolveRetryInterval {
+		devcontainerBackend.userCacheMu.Unlock()
+		return ""
+	}
 	devcontainerBackend.userCacheMu.Unlock()
 
 	user := devcontainerBackend.resolveContainerUser(containerID)
 
-	// Cache only a real answer; leave a miss uncached so the next poll re-probes.
+	devcontainerBackend.userCacheMu.Lock()
 	if user != "" {
-		devcontainerBackend.userCacheMu.Lock()
+		// Cache a real answer permanently; clear any prior miss.
 		devcontainerBackend.userCache[containerID] = user
-		devcontainerBackend.userCacheMu.Unlock()
+		delete(devcontainerBackend.userMissAt, containerID)
+	} else {
+		// Record the miss so the next ticks back off instead of re-resolving.
+		devcontainerBackend.userMissAt[containerID] = time.Now()
 	}
+	devcontainerBackend.userCacheMu.Unlock()
 	return user
 }
 
@@ -76,14 +95,26 @@ func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string
 //     call, cached by the caller, never on the steady-state path. The env probe
 //     is disabled so the username is the only thing on stdout.
 //  3. the legacy ownership heuristic — last resort only.
+//
+// The authoritative answers are sanity-checked: a metadata remoteUser that isn't
+// a usable runtime username (e.g. an unsubstituted "${localEnv:USER}") is
+// rejected so it can't be cached and route every `docker exec -u` to a user that
+// doesn't exist — re-creating the very bug this resolves.
 func (devcontainerBackend *DevcontainerBackend) resolveContainerUser(containerID string) string {
-	if user := remoteUserFromMetadata(containerID); user != "" {
+	if user := remoteUserFromMetadata(containerID); looksLikeUsername(user) {
 		return user
 	}
-	if user := devcontainerBackend.remoteUserViaExec(containerID); user != "" {
+	if user := devcontainerBackend.remoteUserViaExec(containerID); looksLikeUsername(user) {
 		return user
 	}
 	return containerUserHeuristic(containerID)
+}
+
+// looksLikeUsername reports whether s is a plausible single-token Unix username:
+// non-empty, no whitespace, and free of shell-expansion leftovers like an
+// unsubstituted ${...} that a metadata template may carry.
+func looksLikeUsername(s string) bool {
+	return s != "" && !strings.ContainsAny(s, " \t\r\n") && !strings.Contains(s, "${")
 }
 
 // remoteUserViaExec asks the devcontainer CLI who it execs as. --default-user-env-probe
@@ -100,17 +131,14 @@ func (devcontainerBackend *DevcontainerBackend) remoteUserViaExec(containerID st
 	if err != nil {
 		return ""
 	}
-	// Take the last line and reject anything with whitespace: a username is a
-	// single token, so banner noise from a sourced profile can't masquerade as one.
+	// Take the last line: with --default-user-env-probe none the username is the
+	// only stdout, but this is belt-and-suspenders against a stray banner line.
+	// resolveContainerUser validates the token via looksLikeUsername.
 	out = bytes.TrimSpace(out)
 	if i := bytes.LastIndexByte(out, '\n'); i >= 0 {
 		out = bytes.TrimSpace(out[i+1:])
 	}
-	user := string(out)
-	if user == "" || strings.ContainsAny(user, " \t") {
-		return ""
-	}
-	return user
+	return string(out)
 }
 
 // containerUserHeuristic is the legacy last-resort guess: the owner of an active
--- a/internal/backend/devcontainer/devcontainer_backend.go
+++ b/internal/backend/devcontainer/devcontainer_backend.go
@@ -34,13 +34,15 @@ func New(opts ...Option) *DevcontainerBackend {
 	return devcontainerBackend
 }
 
-// containerUser returns the non-root user inside the container, caching
-// the result to avoid a docker exec round-trip on every poll. Only a
-// non-empty answer is cached: an empty probe (a just-started container with
-// no tmux socket and no /home/<user> yet, or a transient exec failure) must be
-// re-probed next call so it self-heals. Caching "" would, now that the backend
-// is reused across poll ticks, pin every later exec to root forever and hide
-// the real user's tmux sessions.
+// containerUser returns the user fleet's direct `docker exec -u` paths
+// (ListSessions, CaptureAllSessions, RunScript, AgentToolProbe) must act as so
+// they operate on the SAME tmux server as the `devcontainer exec` paths
+// (create/attach). That user is the devcontainer remoteUser — see
+// resolveContainerUser. The answer is cached to avoid re-resolving on every
+// poll. Only a non-empty answer is cached: an empty result (a just-started or
+// gone container, or a transient probe failure) must be re-resolved next call
+// so it self-heals. Caching "" would, now that the backend is reused across
+// poll ticks, pin every later exec to the wrong user forever.
 func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string) string {
 	devcontainerBackend.userCacheMu.Lock()
 	if cached, ok := devcontainerBackend.userCache[containerID]; ok {
@@ -49,22 +51,7 @@ func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string
 	}
 	devcontainerBackend.userCacheMu.Unlock()
 
-	// Pick the owner of an active tmux server socket directory.
-	// tmux creates /tmp/tmux-$UID/ for each user that runs a server,
-	// so this is the canonical signal for "which user holds fleet's
-	// sessions". The [1-9]* glob skips /tmp/tmux-0/ — a stale dir
-	// some images leave behind from a root-side init poke at tmux —
-	// because fleet's sessions never live under root. Fall back to
-	// the first /home/<user>/ owner when no tmux socket exists yet
-	// (brand-new container, first poll before any session).
-	cmd := exec.Command("docker", "exec", containerID, "sh", "-c",
-		`for d in /tmp/tmux-[1-9]*/; do [ -d "$d" ] && stat -c %U "$d" 2>/dev/null && exit; done; `+
-			`for d in /home/*/; do [ -d "$d" ] && stat -c %U "$d" 2>/dev/null && exit; done`)
-	out, err := cmd.Output()
-	user := ""
-	if err == nil {
-		user = strings.TrimSpace(string(out))
-	}
+	user := devcontainerBackend.resolveContainerUser(containerID)
 
 	// Cache only a real answer; leave a miss uncached so the next poll re-probes.
 	if user != "" {
@@ -73,6 +60,73 @@ func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string
 		devcontainerBackend.userCacheMu.Unlock()
 	}
 	return user
+}
+
+// resolveContainerUser determines who `devcontainer exec` operates as — the
+// devcontainer remoteUser — so the direct `docker exec -u` paths run as the
+// same user and share its tmux server. Resolving the authoritative user (rather
+// than guessing it from socket ownership) is what keeps create and list from
+// landing on different per-uid tmux sockets. Tries, cheapest-and-surest first:
+//
+//  1. remoteUser from the container's `devcontainer.metadata` label — the value
+//     the CLI itself stamped at `up` time. Pure `docker inspect`, no Node, so it
+//     is safe on the per-second poll path (and runs only on a cache miss).
+//  2. `devcontainer exec ... id -un` — ground truth, asked of the CLI directly,
+//     for the rare container whose metadata carries no remoteUser. One Node
+//     call, cached by the caller, never on the steady-state path. The env probe
+//     is disabled so the username is the only thing on stdout.
+//  3. the legacy ownership heuristic — last resort only.
+func (devcontainerBackend *DevcontainerBackend) resolveContainerUser(containerID string) string {
+	if user := remoteUserFromMetadata(containerID); user != "" {
+		return user
+	}
+	if user := devcontainerBackend.remoteUserViaExec(containerID); user != "" {
+		return user
+	}
+	return containerUserHeuristic(containerID)
+}
+
+// remoteUserViaExec asks the devcontainer CLI who it execs as. --default-user-env-probe
+// none skips the login-shell env capture (faster, and keeps a noisy ~/.bashrc
+// off stdout so the username is all that comes back).
+func (devcontainerBackend *DevcontainerBackend) remoteUserViaExec(containerID string) string {
+	workspaceDir := workspaceFolderForContainer(containerID)
+	if workspaceDir == "" {
+		return ""
+	}
+	out, err := exec.Command("devcontainer",
+		"exec", "--default-user-env-probe", "none", "--workspace-folder", workspaceDir,
+		"id", "-un").Output()
+	if err != nil {
+		return ""
+	}
+	// Take the last line and reject anything with whitespace: a username is a
+	// single token, so banner noise from a sourced profile can't masquerade as one.
+	out = bytes.TrimSpace(out)
+	if i := bytes.LastIndexByte(out, '\n'); i >= 0 {
+		out = bytes.TrimSpace(out[i+1:])
+	}
+	user := string(out)
+	if user == "" || strings.ContainsAny(user, " \t") {
+		return ""
+	}
+	return user
+}
+
+// containerUserHeuristic is the legacy last-resort guess: the owner of an active
+// non-root tmux socket dir, else the first /home/<user>. It is fragile — a
+// non-fleet user that happens to hold a tmux socket (e.g. an image's `app` user
+// running its own tmux) poisons it, which is exactly why the authoritative
+// remoteUser lookups in resolveContainerUser run first. tmux creates
+// /tmp/tmux-$UID/ per user; the [1-9]* glob skips /tmp/tmux-0/ (root).
+func containerUserHeuristic(containerID string) string {
+	out, err := exec.Command("docker", "exec", containerID, "sh", "-c",
+		`for d in /tmp/tmux-[1-9]*/; do [ -d "$d" ] && stat -c %U "$d" 2>/dev/null && exit; done; `+
+			`for d in /home/*/; do [ -d "$d" ] && stat -c %U "$d" 2>/dev/null && exit; done`).Output()
+	if err != nil {
+		return ""
+	}
+	return strings.TrimSpace(string(out))
 }
 
 // Up provisions a devcontainer for the given workspace folder.
--- a/integration/common.sh
+++ b/integration/common.sh
@@ -43,6 +43,12 @@ FIXTURE_LAUNCH_SRC="${INTEGRATION_DIR}/fixture-launch"
 # launches the agent with the prompt.
 FIXTURE_AUTOMATION_SRC="${INTEGRATION_DIR}/fixture-automation"
 
+# Alternate fixture whose postCreate adds a second container user `app` (with a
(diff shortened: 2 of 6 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the diff for #212 (`fix(devcontainer): resolve containerUser from the authoritative remoteUser`). The diagnosis is well-documented and the fix is sound: replacing the socket-ownership *guess* with the authoritative `remoteUser` so the `docker exec -u` transport and `devcontainer exec` land on the same per-uid tmux server is the right call, and keeping both transports as-is avoids the regression surface the alternatives carried. Layering metadata → exec → heuristic cheapest-first is a nice touch. A few substantive notes, none blocking:

**1. The "never on the steady-state path" claim for the Node probe is conditional.** `containerUser` only caches non-empty results, and `resolveContainerUser` falls through to `remoteUserViaExec` (the `devcontainer exec` Node call) whenever `remoteUserFromMetadata` returns `""`. For a container whose merged metadata names neither `remoteUser` nor `containerUser` (bare image, no features, no user in devcontainer.json), method&nbsp;1 returns `""` on *every* poll, so each ~1s tick falls through to the Node probe until it first succeeds and caches. If that probe then persistently fails while the `devcontainer.local_folder` label *is* present (transient CLI/container issue), every tick re-spawns the Node CLI — reintroducing exactly the cold-start CPU burn #182 removed, for as long as the failure persists. Worth either confirming the target images always stamp a metadata user, or bounding the failure path with a short negative-cache TTL on the *resolution* (distinct from the deliberate no-cache-on-empty for `containerUser`).

**2. No timeout/context on the resolve exec calls.** On a cache miss a single call can now fan out to three sequential subprocesses (`docker inspect` → `devcontainer exec` → `docker exec`), and a hung `devcontainer exec` blocks the resolving goroutine (and the poll behind it) indefinitely. This matches the pre-existing code (the old heuristic had no timeout either), so it's not a regression — just more exec surface now sitting on the poll path. A context deadline on `remoteUserViaExec` would be cheap insurance.

**3. Self-healing now only covers *empty* answers, not *wrong* ones.** The "only cache non-empty" rule self-heals a transient miss, but a non-empty-but-wrong resolution gets pinned permanently. The design premise (metadata `remoteUser` == the user `devcontainer exec` runs as) makes this safe in the normal case, but a metadata value that isn't a usable runtime username — e.g. an unsubstituted `${containerEnv:...}` / `${localEnv:...}` in `remoteUser` — would be cached forever and silently route every `docker exec -u` to a user that doesn't exist, re-hiding sessions. Low likelihood, but since the whole bug class is "wrong cached user," a sanity check (reject values containing `${` / non-username chars, same spirit as the whitespace guard in `remoteUserViaExec`) would close it.

**4. Minor — last-line parsing in `remoteUserViaExec`.** Taking the text after the final `\n` and rejecting whitespace is a good guard, but it assumes the *username* is the last stdout line. With `--default-user-env-probe none` + `exec` that should hold (CLI diagnostics go to stderr), so this is fine in practice; just flagging that a stray trailing stdout line from the CLI would yield a false negative and silently drop to the heuristic.

**Positives worth calling out:**
- `TestRemoteUserFromMetadataJSON` covers the parser well — last-wins merge, `containerUser` fallback, `remoteUser`-beats-`containerUser`, and malformed/empty all asserted. The per-scalar last-non-empty logic correctly matches devcontainer merge semantics.
- Clones are covered too: `docker commit` carries the source's `devcontainer.metadata` label into the committed image and `docker run` inherits it, so method&nbsp;1 resolves cloned containers without paying for the Node probe. (Worth a quick manual confirm if you haven't, since `buildCloneRunArgs` only re-stamps `local_folder`/`config_file`.)
- The `remoteUser: root` fix is a genuine correctness gain over the old `[1-9]*` glob that structurally couldn't return root.

Overall this looks ready to merge; items 1 and 3 are the two I'd weigh before doing so.

internal/backend/devcontainer/devcontainer_backend.go
```diff
@@ -75,6 +62,73 @@ func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string
...
+//  2. `devcontainer exec ... id -un` — ground truth, asked of the CLI directly,
+//     for the rare container whose metadata carries no remoteUser. One Node
+//     call, cached by the caller, never on the steady-state path. The env probe
+//     is disabled so the username is the only thing on stdout.
+//  3. the legacy ownership heuristic — last resort only.
+func (devcontainerBackend *DevcontainerBackend) resolveContainerUser(containerID string) string {
+	if user := remoteUserFromMetadata(containerID); user != "" {
+		return user
+	}
+	if user := devcontainerBackend.remoteUserViaExec(containerID); user != "" {
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This fall-through is the spot to watch re: the doc-comment's "never on the steady-state path" claim. `containerUser` caches only non-empty results, so for a container whose merged metadata names neither `remoteUser` nor `containerUser`, method 1 returns `""` on every poll and each ~1s tick reaches `remoteUserViaExec` (the `devcontainer exec` Node call) until it first succeeds and caches. If that probe keeps failing while the `devcontainer.local_folder` label is present, the Node CLI is re-spawned on every tick — the exact cold-start cost #182 removed. Consider a short negative-cache TTL on the *resolution* (separate from the intentional no-cache-on-empty in `containerUser`) to bound the failure path.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the diff for #212 (`fix(devcontainer): resolve containerUser from the authoritative remoteUser`). This is a well-reasoned, well-scoped fix — the diagnosis (per-uid tmux sockets + a cached wrong-user guess) is convincing, and keeping both transports as-is while making them *agree on the user* is the right call. I verified the core premise: `execArgs` (the create/attach `devcontainer exec` path) passes **no** `--remote-user`, so it runs as the metadata `remoteUser` — exactly the user `resolveContainerUser` now targets for the direct `docker exec -u` paths. The merge semantics ("last non-empty `remoteUser`, else last `containerUser`") match how the CLI merges scalar metadata (devcontainer.json last → wins), and `TestRemoteUserFromMetadataJSON` covers it nicely. Unit + integration coverage is a real strength here.

A few **low-severity** observations, none blocking:

**1. The exec fallback can re-incur the Node cold-start on the poll path for an unresolvable container.** `containerUser` caches only non-empty results (correct, for self-heal). But for a container whose `devcontainer.metadata` names no user, step 1 returns `""` and step 2 (`remoteUserViaExec` → `devcontainer exec … id -un`, a Node spawn) runs on the cache miss. If that exec *also* fails (CLI hiccup, container mid-restart), nothing is cached and the **next ~1s poll repeats the Node call** — partially reintroducing the #182 cold-start the design explicitly avoids. In practice it's bounded (most real containers carry `remoteUser` in metadata and resolve via the cheap `docker inspect` path; a single successful exec gets cached), so this only bites a persistently-unresolvable container. Worth a one-line acknowledgement in the comment, or consider negative-caching the exec failure with a short TTL. See inline note.

**2. Concurrent cache-miss resolution (thundering herd).** `containerUser` releases `userCacheMu` before calling `resolveContainerUser`, so on a cold cache the per-second poller *plus* `CaptureAllSessions`/`ListSessions`/`AgentToolProbe` can each run the full resolution concurrently (a `docker inspect`, possibly a Node `devcontainer exec`) for the same container. This lock pattern is pre-existing, but the new resolution is heavier on the no-metadata path, so the duplicated work is more costly than before. A singleflight / per-container in-flight guard would collapse the burst. Minor.

**3. Test narrative vs. actual poison mechanism (`255_tui_session_two_users.sh`).** The real trigger for the old bug in this test is timing — `app` starts its tmux server and the `sleep 3` lets fleetd's poll cache `app` *before* vscode's session socket exists. The `/home/app` sort-order angle is secondary: once both sockets exist, the `/tmp/tmux-[1-9]*/` glob expands sorted and would actually hit `/tmp/tmux-1000` (vscode) first. The comment conflates the two; the test itself is sound and reproduces the regression. Pure doc nit.

Nice touches: the new fixture reuses the same `mcr.microsoft.com/devcontainers/base:debian-12` base image as the other fixtures (no extra CI pull), the `remoteUserViaExec` output hardening (last line, reject whitespace tokens) is appropriately defensive, and refactoring the legacy guess into `containerUserHeuristic` keeps the fallback honest and testable. LGTM with the notes above.

internal/backend/devcontainer/devcontainer_backend.go
```diff
@@ -75,6 +62,73 @@ func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string
...
+		return user
+	}
+	return containerUserHeuristic(containerID)
```
BenjaminBenetti, line 110: 🔍 **CodeReview Agent**

The `remoteUserViaExec` fallback (step 2) is a Node `devcontainer exec` cold start. Because `containerUser` deliberately doesn't cache an empty result, a container whose metadata names no user *and* whose `devcontainer exec` probe fails will re-run this Node spawn on every ~1s poll — the exact cost #182 set out to remove, just confined to an unresolvable-user edge case.

Realistically rare (most containers resolve via the cheap `docker inspect` metadata path on step 1, and a single successful exec is cached thereafter), so this isn't blocking. But consider either noting the loop in the comment or negative-caching an exec failure with a short TTL so a persistently-unresolvable container doesn't spin a Node process every tick.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./...` + binary) | ✅ Pass | Compiles cleanly; `fleet --help` / `version` run |
| `fleet up` + `fleet ls` (create instance) | ✅ Pass | debian-12 instance reaches `running`; `ls` reports it |
| TUI create session — `remoteUser: vscode` | ✅ Pass | `mysession` created & appears in list; "Session created" status |
| `read-session` capture (vscode & root) | ✅ Pass | Both return live shells with the correct user prompt |
| Discriminator — `remoteUser: root` instance | ✅ Pass | New code resolves `root` from metadata → session visible; **old heuristic would return `vscode` (wrong)** |
| Cleanup / no regressions | ✅ Pass | Pre-existing session persisted across refresh; destroy removed all instances |

**Overall: ✅ PASS — 6 of 6 scenarios passing. The reported bug ("tmux sessions can't be opened — nothing shows up") is fixed, and the fix is verified to specifically resolve the wrong-user case.**

---

## Environment

All app execution + interactive TUI driving was isolated in a throwaway Docker container (`qa-212`) wired to this instance's docker-in-docker daemon. Fleet's instance containers were created as siblings on that daemon over a shared workspace path; the TUI was driven via `tmux send-keys` / `capture-pane` **inside** the container. Built from PR head `80d36fe`. Tools: devcontainer CLI `0.87.0`, docker `29.5.3`, go `1.25.9`. Container + all instances torn down afterward.

## Detailed findings

### 1. Build — ✅ Pass
`CGO_ENABLED=0 go build ./...` and a binary build of `./cmd/fleet` both succeeded with no errors; `fleet --help` lists all commands.

### 2. `fleet up` + `fleet ls` — ✅ Pass
Brought up a fixture instance (`mcr.microsoft.com/devcontainers/base:debian-12`):
```
Instance itest-fleet/alpha is running (container: bf44aa026e6a)

FLEET        INSTANCE  STATUS   CONTAINER     CREATED           BRANCH
itest-fleet  alpha     running  bf44aa026e6a  2026-06-24 04:32  main
```
The container's `devcontainer.metadata` label carries `{"remoteUser":"vscode"}` while the container's default exec user is `root` — i.e. create (`devcontainer exec`, as vscode) and list (`docker exec -u …`) must agree on `vscode`, which is exactly what the fix's tier-1 (metadata) lookup resolves.

### 3. TUI session create — `remoteUser: vscode` — ✅ Pass (the headline bug)
Drove the real TUI: navigated to `alpha` → `space` (expand) → `a` (new session) → typed `mysession` → `Enter`. The new row appeared and the status line showed **Session created**:
```
│       ▼ ○ alpha                  running  ○ idle …
│           ○ mysession
│           + new session
…
Session created
```
Verified the mechanism inside the instance — the session lives under the **vscode** uid socket, and fleet's resolved-user list path sees it while a root exec does not:
```
/tmp/tmux-1000/ -> owner vscode
$ docker exec -u vscode <cid> tmux ls
alpha~mysession: 1 windows (created Wed Jun 24 04:34:03 2026)
$ docker exec <cid> tmux ls            # default = root
error connecting to /tmp/tmux-0/default (No such file or directory)
```

### 4. `read-session` capture — ✅ Pass
Both sessions return live, attachable shells, and the prompt confirms the resolved user matched the create user:
```
$ fleet read-session itest-fleet/alpha mysession
vscode ➜ /workspaces/itest-fleet $

$ fleet read-session qa-root-fleet/rootbox rootsess
root ➜ /workspaces/qa-root-fleet (main) $
```

### 5. Discriminator — `remoteUser: root` instance — ✅ Pass (proves the fix specifically)
Created an instance with `"remoteUser": "root"` on the debian-12 base (which still ships `/home/vscode`) — the exact case the PR says the old heuristic mishandled. Captured the discriminating evidence **before** creating a session:
```
# metadata array ends with the user's devcontainer.json (last-wins):
"remoteUser":"vscode"   ← from a feature entry
"remoteUser":"root"     ← from devcontainer.json  →  fix resolves THIS

# the OLD socket/home heuristic, run verbatim against the same container:
$ for d in /tmp/tmux-[1-9]*/; do … ; done; for d in /home/*/; do … stat -c %U … ; done
vscode      ← WRONG (should be root)
```
Then created `rootsess` via the TUI — it appeared with "Session created". Confirming the mechanism:
```
/tmp/tmux-0/ -> root
$ docker exec -u root   <cid> tmux ls   → rootbox~rootsess: 1 windows   (NEW resolved path — visible)
$ docker exec -u vscode <cid> tmux ls   → error connecting to /tmp/tmux-1000/default (No such file or directory)
                                          (OLD path — finds nothing → the bug)
```
This is the clean before/after: the old heuristic targets `vscode` and misses the root-owned session; the new metadata-authoritative resolution targets `root` and the session is visible in the TUI.

### 6. No obvious regressions — ✅ Pass
The `mysession` created earlier persisted across a TUI refresh (`r`) and after a second fleet was added; `fleet destroy` cleanly removed both fleets and all instance containers.

## Notes / scope
- Tier-1 (`devcontainer.metadata` → `remoteUser`) was exercised for both `vscode` and `root` containers. Tiers 2 (`devcontainer exec id -un`) and 3 (legacy heuristic) are fallbacks that did not trigger here (both test containers carried a `remoteUser` in metadata) and were not separately exercised.
- Per manual-QA scope, automated tests were **not** run; `TestRemoteUserFromMetadataJSON` is present in the diff but was not executed by this agent.

BenjaminBenetti:
Thanks — addressed the substantive notes in 081665b:

- **Bogus cached user (review 1 #3 / review 2 inline):** added `looksLikeUsername` and gate both authoritative results through it in `resolveContainerUser`, so an empty/whitespace/`${...}`-template `remoteUser` is rejected and falls through to the next method instead of being cached and routed to a nonexistent `-u`. Replaces the ad-hoc whitespace guard in `remoteUserViaExec`; covered by `TestLooksLikeUsername`.
- **Node probe on the poll path for an unresolvable container (review 1 #1 / review 2 #1):** added a throttle — `userMissAt` + `userResolveRetryInterval` (15s). A miss now backs off instead of re-resolving every ~1s tick, so a persistently-unresolvable container can't re-spawn the resolution subprocesses (incl. the Node CLI) on every poll. A successful resolve still caches permanently and clears the miss. This makes the "never on the steady-state path" claim actually hold.
- **Last-line parsing (review 1 #4):** kept the defensive last-line extraction but dropped its redundant inline check now that `looksLikeUsername` validates the token centrally.
- **Test narrative (review 2 #3):** reworded the `255` comment — the poison is timing (app's socket cached before vscode's session socket exists), not the `/home` sort order alone. You're right that once both sockets exist the `[1-9]*` glob would sort `/tmp/tmux-1000` first.

Deferred, with rationale:
- **Thundering-herd / singleflight (review 2 #2):** pre-existing lock pattern, and now bounded in practice — method 1 (`docker inspect`) is cheap and resolves real fleet containers, and the new miss-throttle caps the heavy path. A singleflight guard is a reasonable future cleanup but felt like over-reach for this fix.
- **Clone coverage:** confirmed the reasoning — `docker commit` carries the source's `devcontainer.metadata` label into the committed image and `docker run` inherits it, so cloned containers resolve via method 1 (no Node). `buildCloneRunArgs` only re-stamps `local_folder`/`config_file`, which doesn't affect metadata.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Follow-up review of `081665b` (synchronize). The substantive notes from the prior round are addressed, and the fixes hold up:

- **Bogus-user guard** — `looksLikeUsername` now gates both authoritative results in `resolveContainerUser`, so an empty/whitespace/`${…}`-template `remoteUser` falls through instead of being cached and routed to a nonexistent `-u`. Validation centralized + covered by `TestLooksLikeUsername`. ✅
- **Node probe on the poll path** — `userMissAt` + `userResolveRetryInterval` (15s) back a miss off instead of re-resolving every ~1s tick, so a persistently-unresolvable container can't re-spawn the Node CLI per poll. Success caches permanently and clears the miss. This makes the doc-comment's "never on the steady-state path" claim actually hold. ✅
- **Test narrative** — the `255` comment now correctly attributes the bug to socket-cache *timing* rather than `/home` sort order alone. ✅

The core approach (resolve the authoritative `remoteUser` from the `devcontainer.metadata` label so both transports agree on the user) is sound, and `remoteUserFromMetadataJSON`'s last-non-empty / `containerUser`-fallback merge matches the devcontainer CLI's scalar merge semantics. Nice test matrix.

A few **minor, non-blocking** residual observations:

1. **Self-heal latency went from ~1s to up to 15s on a transient first miss.** Before, an empty probe was retried every poll; now the first `""` result locks the container out of re-resolution for `userResolveRetryInterval`. For the common case method 1 (`docker inspect`) succeeds on the first poll so nothing is recorded — but if a container's *first* resolution hits a transient failure across all three methods, its sessions stay invisible for up to 15s before the next attempt. Acceptable tradeoff; worth a line in the `containerUser` doc-comment so the user-visible delay isn't a surprise.

2. **Method 3 (`containerUserHeuristic`) isn't gated through `looksLikeUsername`.** Paths 1 and 2 are validated; the heuristic's `stat -c %U` output is returned as-is. In practice it can only yield `""` or a clean token (a name, or a bare numeric uid for an unmapped owner), so this isn't a live bug — but gating it too would make the "never cache a bogus `-u` target" invariant uniform across all three methods rather than relying on the heuristic's output happening to be clean.

3. **`userMissAt` has no eviction for removed containers.** It self-cleans on a successful resolve (`delete`), but an entry for a container removed while still in a miss state lingers, same as the long-standing `userCache` growth. Negligible in practice (keyed by container id, few/long-lived), just flagging the second unbounded map.

The deferred items (singleflight for the unlocked resolve window; clone metadata inheritance via `docker commit`) are reasonably deferred — the miss-throttle bounds the heavy path and clones do carry the source's `devcontainer.metadata` into the committed image, so they resolve via method 1. No blockers from me.

internal/backend/devcontainer/devcontainer_backend.go
```diff
@@ -22,59 +22,141 @@ type DevcontainerBackend struct {
...
+// userResolveRetryInterval so a stuck container can't re-spawn the resolution
+// subprocesses every tick. Caching "" permanently would, now that the backend
+// is reused across poll ticks, pin every later exec to the wrong user forever.
 func (devcontainerBackend *DevcontainerBackend) containerUser(containerID string) string {
 	devcontainerBackend.userCacheMu.Lock()
 	if cached, ok := devcontainerBackend.userCache[containerID]; ok {
 		devcontainerBackend.userCacheMu.Unlock()
 		return cached
 	}
+	if missAt, ok := devcontainerBackend.userMissAt[containerID]; ok && time.Since(missAt) < userResolveRetryInterval {
```
BenjaminBenetti, line 63: 🔍 **CodeReview Agent**

The miss-throttle correctly bounds the heavy resolution path, but it also changes self-heal behavior: previously an empty probe was retried on the next ~1s poll, whereas now the first `""` result suppresses re-resolution for `userResolveRetryInterval` (15s).

For the steady state this is invisible (method 1's `docker inspect` resolves real fleet containers on the first poll, so no miss is recorded). The edge case: a container whose *very first* resolution transiently fails across all three methods will have its tmux sessions stay invisible for up to 15s before the next attempt — versus ~1s before. That's a reasonable tradeoff to avoid the per-tick Node spawn, but the `containerUser` doc-comment frames the throttle purely as a cost guard; a note that it also bounds self-heal latency would set the right expectation for future readers.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & launch fleet TUI | ✅ Pass | Static binary builds; TUI renders armada with no errors |
| Provision two-user instance (`remoteUser: vscode`) | ✅ Pass | `fleet up` ok; postCreate added `app` user (uid 4001) + tmux |
| `devcontainer.metadata` carries `remoteUser` | ✅ Pass | Label resolves `vscode` (last-wins over feature defaults) |
| **Create session while `app` holds the ONLY tmux socket** | ✅ Pass | `mysession` rendered in the poll-fed list — the core regression guard |
| Foreign `app` session excluded | ✅ Pass | `appsess` (in `/tmp/tmux-4001`) never leaks into fleet's list |
| Open shell into session | ✅ Pass | Attaches as `vscode` to `alpha~mysession`, commands run |
| Session persists across TUI restart | ✅ Pass | Cold-cache re-resolve still picks `vscode`; session still listed |
| Multiple sessions / repeatability | ✅ Pass | `mysession` + `second` both listed correctly |
| Delete session | ✅ Pass | `second` removed from TUI and vscode's tmux server |
| Inverse `remoteUser: root` case | ✅ Pass | `rootsess` in `/tmp/tmux-0` (the dir the old heuristic skipped) renders |

**Overall: ✅ PASS — 10 of 10 scenarios passing.**

---

## Environment

All execution was isolated inside a throwaway **privileged Docker-in-Docker** container (`qa-212`, `docker:dind`) with the `devcontainer` CLI (0.87.0), node 24, tmux, and a `CGO_ENABLED=0` build of the PR-branch `fleet` binary. Fleet provisioned real devcontainers via its own inner dockerd; the TUI was driven over tmux inside that container. The container (and all nested state) was torn down at the end. Head tested: `081665b` (matches the trigger payload).

## What the change is supposed to do (from the PR / ticket)

In a fleet, created tmux sessions must actually appear in the TUI and be usable. The root cause was that fleet **creates/attaches** sessions via `devcontainer exec` (runs as the devcontainer **`remoteUser`**) but **lists/captures** them via `docker exec -u $(containerUser)`, where `containerUser()` *guessed* the user from `/tmp/tmux-*/` socket ownership. When a non-fleet user holds a tmux socket, the guess is wrong, the two transports land on different per-uid tmux sockets, and created sessions never render. The fix makes `containerUser()` authoritative — resolve the real `remoteUser` (from the `devcontainer.metadata` label, then `devcontainer exec id -un`, then the legacy heuristic).

## Detailed findings

### 1. Core regression — two-user instance (the exact ticket scenario) ✅
Reproduced the live-confirmed setup: a `remoteUser: vscode` devcontainer whose postCreate also adds an `app` user (uid 4001, `/home/app` — which sorts **before** `/home/vscode`).

Steps:
1. `fleet up alpha` → instance container running; `devcontainer.metadata` label contains `"remoteUser":"vscode"`.
2. **Poisoned the old heuristic first:** `docker exec -u app … tmux new-session -d -s appsess` → created `/tmp/tmux-4001` (owned by `app`) as the **only** tmux socket in the container.
3. In the TUI: expand `alpha` → `a` → name `mysession` → Enter.

Observed: status bar shows **"Session created"** and **`○ mysession` renders in the list**. Underlying state:
```
/tmp/tmux-1000/ owner=vscode   → alpha~mysession   (fleet created here, as the remoteUser)
/tmp/tmux-4001/ owner=app      → appsess           (foreign — correctly NOT listed)
```
This is decisive: the only pre-existing socket belonged to `app`, so the **old** `containerUser()` would have cached `app`, run `docker exec -u app tmux ls`, and shown nothing for fleet's session. Instead the session resolved against `vscode` and rendered, while `appsess` was correctly excluded — i.e. the authoritative resolution is active and working.

### 2. Session is usable end-to-end ✅
`Enter` on `mysession` opened a shell split. Inside: prompt `vscode ➜ /workspaces/itest-fleet`, `whoami` → `vscode`, `tmux display-message` → `ATTACHED_TO=alpha~mysession`. Create *and* attach agree on the same user/server.

### 3. Persistence across TUI restart ✅
Killed the TUI client; `alpha~mysession` survived in the instance. Relaunched `fleet` fresh (cold user cache) — `mysession` was still listed and `appsess` still excluded, confirming the re-resolution path also picks `vscode`.

### 4. Multiple sessions + delete ✅
Added `second` → both `mysession` and `second` listed (vscode's server held both; `app` still only `appsess`). Deleted `second` via `d` → `y`; it disappeared from the TUI and from vscode's tmux server.

### 5. Inverse `remoteUser: root` case ✅
Provisioned a second instance `beta` with `remoteUser: root`. Metadata last-wins → `root`. Created `rootsess`; it rendered in the TUI and lives in **`/tmp/tmux-0`** — exactly the socket dir the old `[1-9]*` heuristic explicitly skipped. The authoritative resolution handles it correctly.

### Stability
No errors, panics, or "permission denied" output appeared at any point across both instances and ~5 session create/attach/delete operations.

**Verdict: ✅ qa-passed** — the change resolves the reported bug (sessions now visible/usable when a foreign user holds a tmux socket, and for root-remoteUser containers) with no regressions observed in the exercised flows.

PR #214, opened 2026-06-25
fix(tui): stop a session group swallowing a sibling whose ID is a prefix

BenjaminBenetti:
## Problem

Creating two sessions where one name is a string prefix of the other — e.g. `dog` and `dog-2` — made opening `dog` show the panes of **both** `dog` and `dog-2`. From the bug report's `ps` output, a single outer-tmux window held panes attached to `…~dog`, `…~dog~fd96`, `…~dog-2`, and `…~dog-2~1177` — two groups mixed into one.

## Root cause

A group's tmux sessions are named `<instance>~<groupID>` (root) and `<instance>~<groupID>~<hex>` (panes). Group operations filtered membership with a raw `strings.HasPrefix(name, "<instance>~<groupID>")`. Because `alpha~dog` is a string prefix of `alpha~dog-2`, the group `dog` matched `dog-2`'s sessions. `parseGroupID` already splits on the group separator and draws the boundary correctly — the prefix filters just didn't use it. (The `groupIDFor` comment already flagged this exact hazard.)

## Fix

Add `sessionInGroup(sanitizedInstance, sessionName, groupID)` (built on `parseGroupID`) and route every group-membership filter through it:

- **`restoreSessionNames`** — the visible "open/restore shows both groups' panes" bug.
- **`deleteGroupSessionsCmd` / `renameGroupCmd`** — these were *data-loss / corruption*: deleting `dog` also killed `dog-2`; renaming `dog` also renamed `dog-2`.
- **`sessionStillExists`** — a dead group stayed "alive" because a sibling's name contained `~<groupID>`. The rewrite also fixes the inverse (a bare ad-hoc session whose pseudo-group ID has no `~` now matches via an exact-name fallthrough).
- **`migrateRenamedSession` / `migrateSavedGroup`** reprefix guards in `app.go`.

No change to session creation, `groupIDFor`, preset scripts, or the #158 multi-TUI snapshot logic — only the membership predicate is corrected, so the recent session fixes (#160, #165, #211, #212) are preserved.

## Tests

- **`internal/tui/group_prefix_test.go`** — unit coverage for the `dog`/`dog-2` boundary across `sessionInGroup`, `restoreSessionNames`, `sessionStillExists`, `deleteGroupSessionsCmd`, `renameGroupCmd`.
- **`integration/tests/265_tui_session_group_prefix.sh`** — end-to-end: create `dog` + `dog-2`, open `dog`, assert exactly one split pane (TUI + dog) and that no pane is titled with `dog-2`. Fails on the old code (3 panes), passes on the fix.

Existing `restoreSessionNames` tests (incl. the #158 cases) were updated to pass the group ID rather than the prebuilt prefix and still pass.

### DIFF
--- a/internal/tui/groups.go
+++ b/internal/tui/groups.go
@@ -129,9 +129,11 @@ func NewGroupPaneSessionName(sanitizedInstance, groupID string) string {
 // apply, but a re-apply (or a leftover from a prior failed apply) falls back to
 // a random id so the root never collides forever on `tmux new-session` — the
 // reported "stuck, a restart doesn't help" bug, since the session lives in the
-// container. It deliberately does NOT auto-suffix (dev-2): that would prefix-
-// match a sibling group ("dev") in the prefix-based group ops (open/rename/
-// delete) — readable names for re-applies need boundary-aware matching first.
+// container. It falls back to a random id rather than an auto-suffixed readable
+// name (dev-2): a random id is the minimal collision-free choice and keeps this
+// path simple. (The group ops now match on the group boundary — sessionInGroup,
+// not a raw string prefix — so an auto-suffixed sibling would no longer be
+// swallowed; readable auto-suffixing is therefore a safe possible future change.)
 func groupIDFor(sanitizedInstance, desired string, existing []tmuxSession) string {
 	for _, s := range existing {
 		if gid, ok := parseGroupID(sanitizedInstance, s.Name); ok && gid == desired {
--- a/internal/tui/groups.go
+++ b/internal/tui/groups.go
@@ -222,6 +222,23 @@ func isGroupedSession(sanitizedInstance, sessionName string) bool {
 	return ok
 }
 
+// sessionInGroup reports whether sessionName belongs to group groupID of the
+// given instance: its root (<inst>~<gid>) or one of its panes
+// (<inst>~<gid>~<suffix>).
+//
+// This is the boundary-aware membership test every group operation
+// (open/restore/rename/delete) must use. The naive alternative —
+// strings.HasPrefix(name, <inst>~<gid>) — also matches a sibling group whose
+// ID merely has this group's ID as a string prefix: group "dog" swallows
+// "dog-2" and "dog-2~ff00", which is exactly the reported bug where opening
+// "dog" shows the panes of both "dog" and "dog-2". parseGroupID splits on the
+// group separator, so it draws the group boundary correctly where a raw string
+// prefix does not.
+func sessionInGroup(sanitizedInstance, sessionName, groupID string) bool {
+	gid, ok := parseGroupID(sanitizedInstance, sessionName)
+	return ok && gid == groupID
+}
+
 // splitBindGroupID returns the group ID that is safe to pass to
 // `fleet shell --group` when rebinding the outer tmux split keys after
 // attaching to sessionName. It mirrors the isGroupedSession guard the
--- a/integration/tests/265_tui_session_group_prefix.sh
+++ b/integration/tests/265_tui_session_group_prefix.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+# Description: Opening a session group does not pull in a sibling group whose ID is a string prefix (dog vs dog-2)
+set -euo pipefail
+
+source "$(dirname "$0")/../common.sh"
+itest_cleanup() { tui_kill; }
+itest_begin
+
+# Repro for the "dog shows panes from both dog and dog-2" bug. The TUI names a
+# group's sessions <instance>~<groupID> (root) and <instance>~<groupID>~<hex>
+# (panes). Group operations used to filter membership with a raw
+# strings.HasPrefix(name, "<instance>~<groupID>"), so the prefix "alpha~dog"
+# also matched the sibling group "alpha~dog-2" (and its panes). Opening "dog"
+# then split a pane for every "dog" AND "dog-2" session — the reported symptom.
+#
+# This test creates "dog" and "dog-2", opens "dog", and asserts exactly one
+# shell pane appears (2 panes total incl. the TUI). On the buggy code "dog-2"
+# leaks in as a second shell pane (3 total) and the assertion fails.
+setup_test
+fleet_up alpha
+
+info "spawning TUI"
+tui_spawn
+
+info "waiting for TUI to hydrate before any keypress"
+tui_wait_for "alpha" 15
+tui_wait_for "○ idle" 60
+
+info "moving cursor to alpha instance row and expanding its session list"
+tui_send j
+sleep 0.5
+tui_send Space
+tui_wait_for "+ new session" 15
+
+# --- create session "dog" ---
+info "creating session 'dog'"
+tui_send a
+tui_wait_for "session-name (or empty for auto)" 5
+tui_send_text "dog"
+sleep 0.3
+tui_send Enter
+tui_wait_for "Session created" 15
+tui_wait_for "○ dog" 15
+
+# --- create sibling session "dog-2" (cursor is still on the instance row) ---
+info "creating sibling session 'dog-2'"
+tui_send a
+tui_wait_for "session-name (or empty for auto)" 5
+tui_send_text "dog-2"
+sleep 0.3
+tui_send Enter
+tui_wait_for "Session created" 15
+# Wait for the dog-2 row AND for the server's ~1s poll to surface it in the
+# runtime — opening "dog" reads that runtime session list, so both sessions
+# must be live there for the bug (if present) to manifest.
+tui_wait_for "○ dog-2" 15
+sleep 2
+
+# --- open the "dog" group ---
+# Rows under the expanded instance are sorted by group ID: "dog" then "dog-2".
+# From the instance row a single 'j' lands on "dog".
+info "opening the 'dog' group"
+tui_send j
+sleep 0.5
+tui_send Enter
+
+# The split opens asynchronously; wait for it to appear.
+info "waiting for the dog split pane to open"
+opened=0
+deadline=$(( $(date +%s) + $(_scale_timeout 20) ))
+while [ "$(date +%s)" -lt "${deadline}" ]; do
+  [ "$(tui_pane_count)" -ge 2 ] && { opened=1; break; }
+  sleep 0.25
+done
+[ "${opened}" -eq 1 ] || fail "the 'dog' group never opened a split pane"
+
+# restoreGroupCmd creates every pane synchronously, so the count settles fast.
+# Give a leaked dog-2 pane time to materialize before asserting the final count.
+sleep 3
+panes="$(tui_pane_count)"
+assert_equals 2 "${panes}" \
+  "opening 'dog' produced ${panes} panes; expected 2 (TUI + dog) — the 'dog-2' group leaked in via a prefix-matched group restore"
+
+# Belt-and-suspenders: a leaked pane is titled with the dog-2 session name.
+titles="$(tmux list-panes -t "${TUI_SESSION}" -F '#{pane_title}' 2>/dev/null || true)"
+assert_not_contains "${titles}" "dog-2" \
+  "the dog-2 session leaked into the dog group's split panes (titles: ${titles})"
+
+pass "opening a session group ignores a sibling group whose ID is a prefix"
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -444,12 +444,12 @@ func (m *model) migrateRenamedSession(msg sessionRenamedMsg) {
 				bindHostSplitKeys(msg.ref.Key(), msg.newGroupID)
 			}
 		}
-		if fleetPage.split.ref == msg.ref && strings.HasPrefix(fleetPage.split.session, oldPrefix) {
+		if fleetPage.split.ref == msg.ref && sessionInGroup(sanitized, fleetPage.split.session, msg.oldGroupID) {
 			fleetPage.split.session = newPrefix + fleetPage.split.session[len(oldPrefix):]
 		}
 		if last, ok := m.sessionStore.LastActive(msg.ref); ok && last.groupID == msg.oldGroupID {
 			last.groupID = msg.newGroupID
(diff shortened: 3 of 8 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the fix for the `dog` / `dog-2` group-prefix collision. This is a clean, well-scoped, and well-tested change — I could not find a correctness problem. Notes below.

**What I verified**

- Centralizing membership behind `sessionInGroup` (built on `parseGroupID`) is the right fix: the boundary is drawn at `groupSep`, so `dog` no longer swallows `dog-2`/`dog-2~ff00`.
- All live-list membership filters were rerouted: `restoreSessionNames` (open/restore), `deleteGroupSessionsCmd`, `renameGroupCmd`, `sessionStillExists`, and the `migrateRenamedSession` / `migrateSavedGroup` reprefix guards. I grepped the `internal/tui` package for any remaining raw `HasPrefix`/`Contains` group tests — the only other matches are `parseGroupID` itself and `SessionStore.Groups`, and the latter filters at the *instance* boundary (`sanitized + "~"`, separator-terminated), which is correct to leave alone.
- The reprefix arithmetic (`newPrefix + name[len(oldPrefix):]`) stays valid: `sessionInGroup` guarantees `name` is exactly `oldPrefix` or `oldPrefix + "~" + suffix`, so the slice is always well-formed.
- `sessionStillExists` rewrite is sound. Dropping the early `return false` and falling through to the exact-name check is purely additive: for a real group an exact-name hit would already have matched the grouped check, so the fall-through only newly helps the bare ad-hoc case — and that path's old `Contains(name, "~"+groupID)` actually *mis-pruned* live ad-hoc sessions, so this also fixes a latent "too tight" bug (covered by the new `ad-hoc bare session` case).
- The `restoreSessionNames` recreate-all-dead branch (`len(live)==0`) returns the snapshot names unfiltered, but that is safe: `derivePersistableSnapshot` already captures snapshots boundary-aware (`parseGroupID` + `parsedGID != groupID`), so a saved snapshot can't carry a sibling group's panes.
- `go build ./...` + `go vet ./internal/tui/` are clean; the new unit tests and the updated #158 `restoreSessionNames` tests pass; the integration helpers and `TUI_SESSION` all exist and `265_` does not collide with an existing test number.

**Minor / non-blocking**

1. `groupIDFor`'s doc comment (`internal/tui/groups.go:124–134`) still justifies "deliberately does NOT auto-suffix (dev-2)" because it "would prefix-match a sibling group ('dev') in the prefix-based group ops." With this PR those ops are no longer prefix-based, so that rationale is now stale/misleading. The behavior change is intentionally out of scope, but the comment is worth a follow-up touch-up (and it may be worth revisiting whether readable auto-suffixed names are now safe).
2. `derivePersistableSnapshot` (`splitpane.go`) hand-rolls `parseGroupID(...)` + `parsedGID != groupID`, which is exactly `!sessionInGroup(...)`. Small DRY opportunity to reuse the new helper.
3. Integration test `265` leans on cursor-position + lexical-sort assumptions (a single `j` from the instance row lands on `dog`) and fixed `sleep`s for async pane creation — a mild flakiness risk. The explicit pane-count assertion plus the `dog-2`-title belt-and-suspenders check make it a solid repro, so this is acceptable as-is.

Nice work — correct, minimal, and the unit + e2e coverage (including the "too loose"/"too tight" notes on `sessionStillExists`) is excellent.

internal/tui/groups.go
```diff
@@ -222,6 +222,23 @@ func isGroupedSession(sanitizedInstance, sessionName string) bool {
...
+//
+// This is the boundary-aware membership test every group operation
+// (open/restore/rename/delete) must use. The naive alternative —
+// strings.HasPrefix(name, <inst>~<gid>) — also matches a sibling group whose
+// ID merely has this group's ID as a string prefix: group "dog" swallows
+// "dog-2" and "dog-2~ff00", which is exactly the reported bug where opening
+// "dog" shows the panes of both "dog" and "dog-2". parseGroupID splits on the
+// group separator, so it draws the group boundary correctly where a raw string
+// prefix does not.
+func sessionInGroup(sanitizedInstance, sessionName, groupID string) bool {
```
BenjaminBenetti, line 239: 🔍 **CodeReview Agent**

Now that every group operation routes membership through `sessionInGroup`, the `groupIDFor` doc comment a few lines up (groups.go:124–134) is stale: it still says it "deliberately does NOT auto-suffix (dev-2): that would prefix-match a sibling group ('dev') in the prefix-based group ops … readable names for re-applies need boundary-aware matching first." Those ops are no longer prefix-based — boundary-aware matching now exists. Worth updating that comment in a follow-up, and possibly reconsidering whether readable auto-suffixed names (`dev-2`) are now safe to mint. Non-blocking.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| TUI launch & render | ✅ Pass | `alpha` instance shows `running / ○ idle`; no startup errors |
| Create sessions `dog` & `dog-2` | ✅ Pass | Both created; container holds `alpha~dog` + `alpha~dog-2` (the exact prefix-collision setup) |
| **Open `dog` (core fix)** | ✅ Pass | Exactly **2 panes** (TUI + `alpha~dog`); **no `dog-2` leak** |
| Delete `dog` spares `dog-2` | ✅ Pass | After deleting `dog`, only `alpha~dog-2` survives — no data loss |
| Rename `dog` → `cat` spares `dog-2` | ✅ Pass | Result is `alpha~cat` + `alpha~dog-2`; no collateral `alpha~cat-2` |
| Open `dog-2` (normal open / no regression) | ✅ Pass | Exactly 2 panes (TUI + `alpha~dog-2`) |
| Quit TUI | ✅ Pass | Clean exit with `q` |

**Overall: ✅ PASS — 7 of 7 scenarios passing.** The headline bug is fixed and the related data-loss paths (delete/rename) behave correctly.

---

### Environment
- Built `fleet` from the PR head (`8223507`) and exercised it **manually** like a user — driving the real TUI via `tmux send-keys` / `capture-pane`.
- All app execution was isolated inside a dedicated throwaway Docker container; a real devcontainer instance `alpha` (`mcr.microsoft.com/devcontainers/base:debian-12`) was spun up via `fleet up`.
- No unit/integration tests were run — this is behavior-level QA only.

### Detailed findings

**Setup / repro condition.** Created sessions `dog` then `dog-2` through the TUI ("a" → name → Enter). Inside the `alpha` container this produced exactly the two tmux sessions that trip the bug, where one name is a literal string prefix of the other:
```
alpha~dog:   1 windows
alpha~dog-2: 1 windows
```

**1. Open `dog` — the reported bug (restoreSessionNames).** Navigated to the `dog` row and pressed Enter. The split opened with exactly two panes and the second pane is the `dog` shell only:
```
pane 0 | title=[<tui-host>]  | cmd=fleet
pane 1 | title=[alpha~dog]   | cmd=sh
pane count: 2
```
No pane titled `alpha~dog-2` appeared. On the pre-fix code the raw `strings.HasPrefix(name, "alpha~dog")` would also match `alpha~dog-2`, producing a 3rd pane — the "opening dog shows panes of both groups" symptom. Confirmed fixed.

**2. Delete `dog` does not kill `dog-2` (deleteGroupSessionsCmd).** With both `dog` and `dog-2` present, deleted `dog` via `d` → `y`. TUI then listed only `○ dog-2`, and the container confirmed:
```
# after delete:
alpha~dog-2: 1 windows      # alpha~dog killed, dog-2 untouched
```
The old prefix match would have killed `alpha~dog-2` too (data loss). Confirmed fixed.

**3. Rename `dog` → `cat` does not rename `dog-2` (renameGroupCmd + app.go migrate guards).** Recreated `dog`, renamed it to `cat` via `r`. Result:
```
alpha~cat:   1 windows
alpha~dog-2: 1 windows
PASS: no alpha~cat-2 (dog-2 left untouched)
```
The old prefix match would have rewritten `alpha~dog-2` → `alpha~cat-2` (corruption). Confirmed fixed.

**4. Normal open still works (regression check).** Opened `dog-2` directly → exactly 2 panes (TUI + `alpha~dog-2`), no extra/leaked pane. The new `sessionInGroup` predicate did not break the ordinary single-group open path.

**5. Sanity.** TUI launches and renders the armada/instance/session tree without errors, session create/delete/rename dialogs render and respond, and `q` exits cleanly.

Verdict: **qa-passed**.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Second pass** — reviewing the new commit `131ea99` on top of the original fix. The prior note (stale `groupIDFor` doc comment) was addressed, and the follow-up commit cleanly reuses `sessionInGroup` in `derivePersistableSnapshot`/`snapshotMatchesRuntime`. Nice tightening.

Overall this is a well-scoped, correct, and well-tested fix. I verified the substance rather than just the diff:

**Correctness ✅**
- `parseGroupID` draws the group boundary with `strings.SplitN(rest, "~", 2)[0]`, so `sessionInGroup` genuinely distinguishes `dog` from `dog-2` — the core of the fix is sound.
- **Slice-safety holds.** Every `newPrefix + name[len(oldPrefix):]` (in `renameGroupCmd`, `migrateRenamedSession`, `migrateSavedGroup`) is now guarded by `sessionInGroup(..., oldGroupID)`, and at all three sites `oldPrefix == sanitized + groupSep + oldGroupID`. Since `sessionInGroup`→true implies the name starts with that exact prefix, there's no out-of-range slicing risk introduced by the predicate swap.
- The `splitpane.go` changes are behavior-preserving: `!sessionInGroup(...)` ≡ the old `!ok || gid != groupID`.
- `sessionStillExists`: dropping the early `return false` and falling through to the exact-name check is strictly *better* — the old `strings.Contains(name, "~"+groupID)` was both too loose (`alpha~dog-2` kept group `dog` alive) and too tight (a bare ad-hoc session, or one like `my~thing`, never matched). The new path handles both. Signature change has its only caller (`PruneStaleLastActive`) updated.

**Completeness ✅**
- I checked for other raw `HasPrefix`/`Contains` group-membership filters that might have been missed. The only remaining one — `SessionStore.Groups` (`sessionstore.go`) — is an *instance*-level filter (`prefix = sanitized + groupSep`, with the trailing `~`), so its boundary is already protected by the separator and it's correctly left unchanged. Every *group*-level filter now routes through `sessionInGroup`.

**Tests ✅**
- Unit + integration coverage map directly onto the regression (boundary table, restore, delete, rename, `sessionStillExists`, plus the e2e `265`). The `265` helpers (`tui_pane_count`, `assert_not_contains`, `_scale_timeout`) all match existing conventions.

**Non-blocking FYI**
- At review time the CI bucket that runs the new e2e test (`integration 200-300`, where `265` lives) is still `pending`. Unit tests and the other integration buckets are green. Worth a glance once that bucket finishes to confirm `265` passes on CI as it did locally.

No correctness or security concerns found. 👍

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Launch TUI & main list renders | ✅ Pass | `alpha` instance row renders `running ○ idle`, no startup errors |
| Create sessions `dog` and `dog-2` | ✅ Pass | Both rows appear under `alpha`; "Session created" confirmed |
| Open `dog` group — no `dog-2` leak (core fix) | ✅ Pass | Exactly 2 panes (TUI + one shell titled `alpha~dog`); zero `dog-2` panes |
| Open `dog-2` group separately (sanity) | ✅ Pass | Replaces split with one shell titled `alpha~dog-2`; `dog` did not leak in |

**Overall: ✅ PASS — 4 of 4 scenarios passing**

The reported bug — opening the `dog` session group pulling in the sibling `dog-2` group's panes because `alpha~dog` is a string prefix of `alpha~dog-2` — is fixed. Opening a group now restores **only** that group's panes.

---

### Environment
Built a static `fleet` binary from the PR head (`131ea99`) and ran **manual, behavior-level QA** inside an isolated privileged Docker-in-Docker container (own `dockerd` + `@devcontainers/cli` 0.87.0 + tmux 3.5a). The fixture fleet uses `mcr.microsoft.com/devcontainers/base:debian-12`. No automated unit/integration tests were run — the TUI was driven by hand via tmux (send-keys + capture-pane), exactly as a user would.

### Detailed findings

**1. Launch & main list — ✅ Pass**
`fleet up alpha` brought up the fixture devcontainer (container `4648d43f692f`, status `running`). Launched the TUI in tmux; the Armada list rendered `itest-fleet (1)` → `○ alpha  running  ○ idle` with no errors.

**2. Create `dog` and `dog-2` — ✅ Pass**
Expanded `alpha`, pressed `a`, typed `dog`, Enter → "Session created", `○ dog` row appeared. Repeated with `dog-2` → `○ dog-2` row appeared. Confirmed the instance now holds two distinct tmux sessions, which is precisely the prefix-collision case:
```
alpha~dog:   1 windows (attached)
alpha~dog-2: 1 windows (attached)
```
(`alpha~dog` is a string prefix of `alpha~dog-2` — the exact condition that triggered the bug.)

**3. Open `dog` group — core fix — ✅ Pass**
Moved the cursor to the `dog` row and pressed Enter. A single split pane opened. After waiting several seconds for any leaked pane to materialize:
```
panes = 2
0: title=[<tui>]        cmd=[fleet]
1: title=[alpha~dog]    cmd=[sh]   ← real shell: "vscode ➜ /workspaces/itest-fleet $"
```
Exactly **2** panes (TUI + the single `dog` shell). **No** pane titled `dog-2`. On the pre-fix code this would have produced 3 panes with a leaked `alpha~dog-2` pane.

**4. Open `dog-2` group separately — sanity — ✅ Pass**
Navigated to the `dog-2` row and pressed Enter. The TUI killed the prior `dog` split and opened `dog-2`'s pane:
```
panes = 2
0: title=[<tui>]         cmd=[fleet]
1: title=[alpha~dog-2]   cmd=[sh]
```
Exactly **2** panes; no standalone `~dog` pane leaked into the `dog-2` group. Group membership is isolated in both directions.

### Verdict
All tested user-facing flows behave as expected and meet the PR's acceptance criteria. Applying `qa-passed`.

PR #177, opened 2026-06-17
chore/tidy-cli-help — terse, unix-style CLI help text

BenjaminBenetti:
## Summary

`fleet --help` read as marketing copy, not a unix tool. This tightens the root description and subcommand summaries to be short and to the point, the way `git`/`docker`/`scp` write theirs: imperative, no trailing periods, no mechanism details or marketing parentheticals in the one-liners, and `instance` instead of `devcontainer instance`.

```
Before                                                  After
------------------------------------------------------  --------------------------------
fleet-man is a CLI/TUI tool for spawning and managing   Spawn and manage fleets of
fleets of devcontainers from a repo.                    devcontainers from a git repo.
                                                        Run without a command to open the TUI.

copy   Copy a file to or from this instance via your    copy   Copy files to, from, and
       fleet TUI (shorthand: fc)                               between instances
launch Open the in-instance Fleet Launch grid (or       launch Open an instance's links and apps
       list/launch its links and apps)
rebuild Rebuild an instance's container in place         rebuild Rebuild an instance's container
        (preserves the workspace)
list   List devcontainer instances                      list   List instances
start  Start an existing stopped devcontainer instance  start  Start a stopped instance
status Show fleet-wide status summary                   status Show fleet status
stop   Stop a devcontainer instance without removing it stop   Stop an instance without removing it
version Print the fleet binary version                  version Print the fleet version
```

The `fleet copy --help` long help was also trimmed (kept the endpoint rules + examples — `git`/`scp` man pages have examples — but cut the prose). No behavior change; help strings only.

> **Note:** stacked on #176 (`fix/fleet-copy-local-cwd-and-host-prefix`) because it also tidies `copy.go`'s help — base is that branch for a clean diff. Merge #176 first; this retargets to `main` automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #215, opened 2026-06-25
fix/pr-status-track-closed: keep the PR status indicator visible (purple) for closed PRs

### LINKED ISSUE #203: PR status indicator should track closed PRs
### Current State 
The PR status indicator goes away when a PR closes. This makes it hard to remember if the PR is done or not, because the indicator is also not there when the PR has never been opened.

### Desired State 
The PR indicator goes purple (just like the coloring on GitHub) but remains visible. So this way you know there is a closed PR attached to that instance.

BenjaminBenetti:
## Summary

Closes #203. The auto-tag PR indicator vanished the instant a PR closed or merged, making a finished instance indistinguishable from one that never had a PR. Now a branch whose only PR(s) are **closed or merged** keeps a persistent **purple `PR` badge** — matching GitHub's own closed/merged colour — so you can tell "this PR is done" from "no PR was ever opened". The badge stays clickable and opens the closed PR.

## What changed

- **Probe (`internal/server/pr_status.go`)** — when a branch has no open PR, the `gh` probe now falls back to the most recent closed/merged PR for that branch (picked by highest PR number, so it doesn't rely on `gh`'s list ordering). It only fetches `number,state,url,title` for closed PRs (no check detail needed).
- **Parser** — `parsePRProbeOutput` buckets decoded PRs into open vs closed/merged. **Open PRs always take priority**; the new `closedPRStatus()` builds the purple tag only when nothing is open. A branch that never had a PR still yields `nil` (no tag), exactly as before.
- **Proto** — new `PR_SIGNAL_PURPLE = 4` and `PrStatus.closed_count = 8` (field 8 was free; backward-compatible additions).
- **TUI (`internal/tui/auto_tag.go`, `styles.go`)** — `prSignalStyle` maps purple to a new `prPurpleStyle` (256-colour `99`); the render gate now keeps the tag when `closed_count > 0`. A closed-only status renders a bare purple `PR` (no review/checks segments). The existing row-gating and click-to-open navigation work unchanged because they key off the tag being non-empty and off `Prs`, which is populated with the closed ref.

Open-PR behaviour (green/yellow/red, review, checks, multi-PR) is untouched.

### DIFF
--- a/internal/server/pr_status.go
+++ b/internal/server/pr_status.go
@@ -75,7 +75,10 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
     # outside the window (and silently degrade to the purple tag) on a branch
     # that's been the head of many PRs over its life.
     list=$(gh pr list --state all --head "$br" --limit 100 --json number,state --jq '.[] | "\(.state) \(.number)"' 2>/dev/null)
-    open=$(printf '%s\n' "$list" | awk '$1 == "OPEN" { print $2 }')
+    # toupper() so the shell side is as case-defensive as the Go parser's
+    # strings.ToUpper: if gh ever stopped returning uppercase states, an open PR
+    # must still be classified open (not fall through to the purple closed tag).
+    open=$(printf '%s\n' "$list" | awk 'toupper($1) == "OPEN" { print $2 }')
     if [ -n "$open" ]; then
       for num in $open; do
         gh pr view "$num" \
@@ -86,7 +89,7 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
       # any) so the indicator persists in purple instead of disappearing. The
       # highest PR number is unambiguously the newest, so pick it without relying
       # on gh's list ordering.
-      num=$(printf '%s\n' "$list" | awk '$1 != "OPEN" && $2 > max { max = $2 } END { if (max) print max }')
+      num=$(printf '%s\n' "$list" | awk 'toupper($1) != "OPEN" && $2 > max { max = $2 } END { if (max) print max }')
       if [ -n "$num" ]; then
         gh pr view "$num" --json number,state,url,title 2>/dev/null
       fi
--- a/internal/server/pr_status.go
+++ b/internal/server/pr_status.go
@@ -70,8 +70,11 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
   [ "$br" = "HEAD" ] && continue
   ( cd "$dir" || exit 0
     # One list call for the branch (gh's embedded --jq, so no standalone jq
-    # dependency): "<STATE> <number>" per PR, e.g. "OPEN 12" / "MERGED 7".
-    list=$(gh pr list --state all --head "$br" --json number,state --jq '.[] | "\(.state) \(.number)"' 2>/dev/null)
+    # dependency): "<STATE> <number>" per PR, e.g. "OPEN 12" / "MERGED 7". The
+    # explicit --limit lifts gh's default 30-row cap so an open PR can't fall
+    # outside the window (and silently degrade to the purple tag) on a branch
+    # that's been the head of many PRs over its life.
+    list=$(gh pr list --state all --head "$br" --limit 100 --json number,state --jq '.[] | "\(.state) \(.number)"' 2>/dev/null)
     open=$(printf '%s\n' "$list" | awk '$1 == "OPEN" { print $2 }')
     if [ -n "$open" ]; then
       for num in $open; do
@@ -294,6 +297,12 @@ func parsePRProbeOutput(out string) *fleetgrpc.PrStatus {
 // visible — so a finished instance is distinguishable from one that never had a
 // PR — and carries the closed PRs' refs so the tag stays clickable. Returns nil
 // when there are no closed/merged PRs either (the branch never had one).
+//
+// Note the per-repo counting: the probe emits only the single most-recent
+// closed/merged PR PER repo, so ClosedCount is the number of repos with a
+// closed-but-no-open PR (rendered "PRxN"), NOT the total closed PRs in any one
+// repo's history. This deliberately differs from open_count, which sums every
+// open PR — a branch's closed-PR history would otherwise inflate the badge.
 func closedPRStatus(closed []ghPR) *fleetgrpc.PrStatus {
 	if len(closed) == 0 {
 		return nil
--- a/internal/server/pr_status.go
+++ b/internal/server/pr_status.go
@@ -48,11 +48,13 @@ const (
 // then emits the FULL detail of each via `gh pr view <n> --json ...` — one JSON
 // object per PR. gh pr view is used (not gh pr list's --json) because gh pr
 // list's statusCheckRollup is capped at the first 100 checks, while gh pr view
-// paginates the rollup and returns every check. The objects are concatenated on
-// stdout and read back with a streaming json.Decoder, so their exact
-// whitespace/formatting doesn't matter. It prints a sentinel and exits 0 when gh
-// is absent or not logged in, so the server can distinguish "degrade quietly"
-// from a transient exec failure.
+// paginates the rollup and returns every check. When the branch has NO open PR,
+// it instead emits the most recent closed/merged PR (number/state/url/title
+// only — no check detail needed) so the tag persists in purple rather than
+// vanishing. The objects are concatenated on stdout and read back with a
+// streaming json.Decoder, so their exact whitespace/formatting doesn't matter.
+// It prints a sentinel and exits 0 when gh is absent or not logged in, so the
+// server can distinguish "degrade quietly" from a transient exec failure.
 const prProbeScript = `
 command -v gh >/dev/null 2>&1 || { printf '%s\n' "FLEET_NO_GH"; exit 0; }
 gh auth status >/dev/null 2>&1 || { printf '%s\n' "FLEET_NO_AUTH"; exit 0; }
@@ -65,10 +67,20 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
   [ -z "$br" ] && continue
   [ "$br" = "HEAD" ] && continue
   ( cd "$dir" || exit 0
-    for num in $(gh pr list --state open --head "$br" --json number --jq '.[].number' 2>/dev/null); do
-      gh pr view "$num" \
-        --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,url,title 2>/dev/null
-    done
+    open=$(gh pr list --state open --head "$br" --json number --jq '.[].number' 2>/dev/null)
+    if [ -n "$open" ]; then
+      for num in $open; do
+        gh pr view "$num" \
+          --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,url,title 2>/dev/null
+      done
+    else
+      # No open PR on this branch — surface the most recent closed/merged one (if
+      # any) so the indicator persists in purple instead of disappearing. The
+      # highest PR number is unambiguously the newest, so pick it without relying
+      # on gh's list ordering.
+      num=$(gh pr list --state all --head "$br" --json number,state --jq 'map(select(.state != "OPEN")) | sort_by(.number) | last | .number // empty' 2>/dev/null)
+      [ -n "$num" ] && gh pr view "$num" --json number,state,url,title 2>/dev/null
+    fi
   )
 done
 `
@@ -134,8 +146,9 @@ func classifyCheck(c ghCheck) checkResult {
 	}
 }
 
-// aggregatePRStatus folds the open PRs across an instance's repos into the
-// PrStatus the TUI renders. Returns nil when there are no open PRs (no auto-tag).
+// aggregatePRStatus folds the open PRs across an instance's repos into the live
+// green/yellow/red PrStatus the TUI renders. Returns nil when there are no open
+// PRs — the caller then falls back to closedPRStatus for the purple closed tag.
 func aggregatePRStatus(prs []ghPR) *fleetgrpc.PrStatus {
 	if len(prs) == 0 {
 		return nil
@@ -224,10 +237,12 @@ func aggregatePRStatus(prs []ghPR) *fleetgrpc.PrStatus {
 }
 
 // parsePRProbeOutput turns the probe's stdout — one `gh pr view --json` object
-// per open PR, concatenated — into a PrStatus. A nil result means "no auto-tag"
-// (gh unavailable, or no open PRs) and clears any prior status. The gh-missing /
-// no-auth sentinels and the empty case all map to nil. A streaming json.Decoder
-// reads successive objects regardless of their interleaving whitespace; malformed
+// per PR, concatenated — into a PrStatus. Open PRs drive the live green/yellow/red
+// tag and take priority; only when there are none does a closed/merged PR drive
+// the persistent purple tag. A nil result means "no auto-tag" (gh unavailable, or
+// the branch never had a PR) and clears any prior status. The gh-missing / no-auth
+// sentinels and the empty case all map to nil. A streaming json.Decoder reads
+// successive objects regardless of their interleaving whitespace; malformed
 // trailing noise stops the scan without discarding what parsed cleanly.
 func parsePRProbeOutput(out string) *fleetgrpc.PrStatus {
 	// Match the sentinels by PREFIX, not Contains: the script emits one as the
@@ -238,18 +253,50 @@ func parsePRProbeOutput(out string) *fleetgrpc.PrStatus {
 	if strings.HasPrefix(trimmed, prNoGHSentinel) || strings.HasPrefix(trimmed, prNoAuthSentinel) {
 		return nil
 	}
-	var prs []ghPR
+	var open, closed []ghPR
 	dec := json.NewDecoder(strings.NewReader(out))
 	for {
 		var p ghPR
 		if err := dec.Decode(&p); err != nil {
 			break // io.EOF on clean exhaustion, or unexpected noise.
 		}
-		if strings.EqualFold(p.State, "OPEN") {
-			prs = append(prs, p)
+		switch strings.ToUpper(p.State) {
+		case "OPEN":
+			open = append(open, p)
+		case "CLOSED", "MERGED":
+			closed = append(closed, p)
 		}
 	}
-	return aggregatePRStatus(prs)
+	// Open PRs win: their live status is the actionable one. Fall back to the
+	// closed/merged tag only when nothing is open.
+	if s := aggregatePRStatus(open); s != nil {
+		return s
+	}
+	return closedPRStatus(closed)
+}
+
+// closedPRStatus builds the persistent purple "PR" tag shown when a branch has no
+// open PR but DID have one that's since closed or merged. It keeps the tag
+// visible — so a finished instance is distinguishable from one that never had a
+// PR — and carries the closed PRs' refs so the tag stays clickable. Returns nil
+// when there are no closed/merged PRs either (the branch never had one).
+func closedPRStatus(closed []ghPR) *fleetgrpc.PrStatus {
+	if len(closed) == 0 {
+		return nil
+	}
+	refs := make([]*fleetgrpc.PrRef, 0, len(closed))
+	for _, pr := range closed {
+		refs = append(refs, &fleetgrpc.PrRef{
+			Number: int32(pr.Number),
+			Url:    pr.URL,
+			Title:  pr.Title,
+		})
+	}
+	return &fleetgrpc.PrStatus{
+		ClosedCount: int32(len(closed)),
+		PrSignal:    fleetgrpc.PrSignal_PR_SIGNAL_PURPLE,
+		Prs:         refs,
+	}
 }
 
 // probePRStatus runs the gh probe inside the instance and parses the result.
--- a/internal/server/pr_status.go
+++ b/internal/server/pr_status.go
@@ -51,7 +51,9 @@ const (
 // paginates the rollup and returns every check. When the branch has NO open PR,
 // it instead emits the most recent closed/merged PR (number/state/url/title
 // only — no check detail needed) so the tag persists in purple rather than
-// vanishing. The objects are concatenated on stdout and read back with a
+// vanishing. A single `gh pr list --state all` per repo covers both cases (so a
+// branch parked without an open PR costs no extra round-trip vs. the old
+// open-only query). The objects are concatenated on stdout and read back with a
 // streaming json.Decoder, so their exact whitespace/formatting doesn't matter.
 // It prints a sentinel and exits 0 when gh is absent or not logged in, so the
 // server can distinguish "degrade quietly" from a transient exec failure.
@@ -67,7 +69,10 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
   [ -z "$br" ] && continue
   [ "$br" = "HEAD" ] && continue
   ( cd "$dir" || exit 0
-    open=$(gh pr list --state open --head "$br" --json number --jq '.[].number' 2>/dev/null)
+    # One list call for the branch (gh's embedded --jq, so no standalone jq
+    # dependency): "<STATE> <number>" per PR, e.g. "OPEN 12" / "MERGED 7".
+    list=$(gh pr list --state all --head "$br" --json number,state --jq '.[] | "\(.state) \(.number)"' 2>/dev/null)
+    open=$(printf '%s\n' "$list" | awk '$1 == "OPEN" { print $2 }')
     if [ -n "$open" ]; then
       for num in $open; do
         gh pr view "$num" \
@@ -78,11 +83,20 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
       # any) so the indicator persists in purple instead of disappearing. The
       # highest PR number is unambiguously the newest, so pick it without relying
       # on gh's list ordering.
-      num=$(gh pr list --state all --head "$br" --json number,state --jq 'map(select(.state != "OPEN")) | sort_by(.number) | last | .number // empty' 2>/dev/null)
-      [ -n "$num" ] && gh pr view "$num" --json number,state,url,title 2>/dev/null
+      num=$(printf '%s\n' "$list" | awk '$1 != "OPEN" && $2 > max { max = $2 } END { if (max) print max }')
+      if [ -n "$num" ]; then
+        gh pr view "$num" --json number,state,url,title 2>/dev/null
+      fi
     fi
   )
 done
+# Always exit 0 once the scan completes: an empty result (a repo with no PRs) is a
+# valid probe outcome that should CLEAR the tag, not a transient failure. Without
+# this, a no-PR repo visited last by find leaves the loop's exit status at 1, and
+# probePRStatus would treat the whole probe as failed (ok=false) — discarding any
+# PR JSON already emitted and freezing the prior tag. A non-zero exit (ok=false,
+# keep prior) is reserved for genuine exec/timeout errors, which abort earlier.
+exit 0
 `
 
 // ghPR mirrors the subset of `gh pr list --json` fields the auto-tag needs.
--- a/fleetgrpc/runtime.pb.go
+++ b/fleetgrpc/runtime.pb.go
@@ -589,11 +589,13 @@ type PrStatus struct {
 	// repo first, then subrepos). These are the OPEN PRs when open_count > 0,
 	// otherwise the closed/merged PRs backing the purple tag.
 	Prs []*PrRef `protobuf:"bytes,7,rep,name=prs,proto3" json:"prs,omitempty"`
-	// Closed or merged PRs counted across the repos, set only when NO PR is open
-	// (open_count 0). It keeps the tag visible in purple — matching GitHub's
-	// closed/merged colour — so a finished instance stays distinguishable from one
-	// that never had a PR. Open PRs always take priority; when both exist the open
-	// status is shown and this stays 0.
+	// Closed or merged PRs, set only when NO PR is open (open_count 0). It keeps
+	// the tag visible in purple — matching GitHub's closed/merged colour — so a
+	// finished instance stays distinguishable from one that never had a PR. This
+	// counts the most-recent closed/merged PR PER repo (so it's the number of repos
+	// with a closed-but-no-open PR), unlike open_count which sums every open PR.
+	// Open PRs always take priority; when both exist the open status is shown and
+	// this stays 0.
 	ClosedCount   int32 `protobuf:"varint,8,opt,name=closed_count,json=closedCount,proto3" json:"closed_count,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
--- a/fleetgrpc/runtime.pb.go
+++ b/fleetgrpc/runtime.pb.go
@@ -201,8 +201,10 @@ func (AgentActivity) EnumDescriptor() ([]byte, []int) {
 	return file_runtime_proto_rawDescGZIP(), []int{2}
 }
 
-// PrSignal is the three-state colour shared by the auto-tag indicators:
-// green = good, yellow = in-progress/neutral, red = needs attention.
+// PrSignal is the colour shared by the auto-tag indicators: green = good,
+// yellow = in-progress/neutral, red = needs attention, purple = the PR is closed
+// or merged (kept visible so a finished instance is distinguishable from one that
+// never had a PR), matching GitHub's own closed/merged colour.
 type PrSignal int32
 
 const (
@@ -210,6 +212,7 @@ const (
 	PrSignal_PR_SIGNAL_GREEN       PrSignal = 1
 	PrSignal_PR_SIGNAL_YELLOW      PrSignal = 2
 	PrSignal_PR_SIGNAL_RED         PrSignal = 3
+	PrSignal_PR_SIGNAL_PURPLE      PrSignal = 4
 )
 
 // Enum value maps for PrSignal.
@@ -219,12 +222,14 @@ var (
 		1: "PR_SIGNAL_GREEN",
 		2: "PR_SIGNAL_YELLOW",
 		3: "PR_SIGNAL_RED",
+		4: "PR_SIGNAL_PURPLE",
 	}
 	PrSignal_value = map[string]int32{
 		"PR_SIGNAL_UNSPECIFIED": 0,
 		"PR_SIGNAL_GREEN":       1,
 		"PR_SIGNAL_YELLOW":      2,
 		"PR_SIGNAL_RED":         3,
+		"PR_SIGNAL_PURPLE":      4,
 	}
 )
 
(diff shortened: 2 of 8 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary** — Reviewed the closed/merged "purple PR tag" change (proto + server probe/parser + TUI). I built the branch, ran `go vet` and the full test suite (all green), and confirmed `fleetgrpc/runtime.pb.go` regenerates **byte-identically** via `buf generate` — so the hand-touched generated proto is correct and there's no `rawDesc`/descriptor drift (which would otherwise panic at package init). Overall this is a clean, well-documented, well-tested change. A few minor, non-blocking observations:

**1. Extra `gh` API call per probe for branches without an open PR** — `internal/server/pr_status.go`
The new `else` branch adds a `gh pr list --state all` (plus a `gh pr view`) for every repo/subrepo on a branch with no open PR — including instances parked on `main` or any never-PR'd branch — on each 45s probe. Previously that case made a single `gh pr list --state open` call. It's gated by a connected TUI and the slow cadence so it stays well under rate limits, but with several subrepos it does multiply `gh` round-trips and probe latency. Worth being aware of; not a blocker.

**2. Multiple closed PRs collapse to a bare `PR` badge** — `internal/tui/auto_tag.go`
The label only consults `OpenCount`, so when `closed_count > 1` (e.g. two subrepos each with a closed PR on the branch) the badge still shows `PR`, not `PRx2`, yet the click-chooser then lists N entries. Open multi-PR shows `PRxN`, so there's a small open-vs-closed asymmetry. Looks intentional per the description ("bare purple PR"); just flagging it.

**3. Branch-name reuse can yield a false purple tag**
Because the closed fallback keys on branch name (`--head "$br"`), a reused/recreated branch that shares a name with an old merged PR shows purple even before any new PR exists. Inherent to branch-name keying and a rare edge — fine to accept.

**Nice touches**: keying on `--head "$br"` correctly avoids false purples on `main`/default branches (PRs are rarely opened *from* main); populating `Prs` with the closed ref keeps click-to-open and the inline-row gating working for free; and the `jq` `... | last | .number // empty` guard cleanly yields no output (rather than erroring) when there are no closed PRs. One thing worth a quick manual sanity check: this repo has `delete_branch_on_merge`, so the issue-#203 scenario is "merged PR + deleted branch" — that should still resolve because `gh --head` matches the PR's *recorded* head ref, but confirming on a real merged instance would be reassuring.

_No code was modified; this is a comment-only review._

internal/server/pr_status.go
```diff
@@ -65,10 +67,20 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
...
+      for num in $open; do
+        gh pr view "$num" \
+          --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,url,title 2>/dev/null
+      done
+    else
+      # No open PR on this branch — surface the most recent closed/merged one (if
+      # any) so the indicator persists in purple instead of disappearing. The
+      # highest PR number is unambiguously the newest, so pick it without relying
+      # on gh's list ordering.
+      num=$(gh pr list --state all --head "$br" --json number,state --jq 'map(select(.state != "OPEN")) | sort_by(.number) | last | .number // empty' 2>/dev/null)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Minor (efficiency): this `else` runs `gh pr list --state all` plus a follow-up `gh pr view` for every repo/subrepo with no open PR, on each 45s probe — including instances parked on `main` or branches that never had a PR. Previously the no-open-PR path made just the single `--state open` list call. It's well within rate limits given the cadence and TUI gating, but it multiplies `gh` round-trips (and probe latency) when an instance has several subrepos. Not blocking — just flagging the added cost.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./...` + `fleet` binary) | ✅ Pass | Compiles clean on go1.25, no errors |
| TUI launches & renders | ✅ Pass | Dependency-check page + Armada list render, no crash |
| Closed/merged PR → purple-tag data (single repo) | ✅ Pass | Probe emits `{#210, MERGED, url, title}` → feeds the purple `PR` badge |
| Pick most-recent closed PR by number | ✅ Pass | Branch with #190+#192 (both closed) → probe correctly picks **#192** |
| Open PR takes priority over closed fallback | ✅ Pass | Branch of this PR → full **OPEN #215** detail emitted, not the closed fallback |
| "Never had a PR" → no tag | ⚠️ Partial | Emits no JSON (correct outcome) **but the script now exits non-zero** — root cause of the regression below |
| Multi-repo / no-PR probe robustness | ❌ Fail | When the last repo `find` visits has no PR, the **whole probe exits 1 and valid PR JSON is discarded** → tag frozen |

**Overall: ❌ FAIL — 1 of 7 scenarios failing (multi-repo PR-status probe regression). The headline single-repo feature works, but the new no-PR exit path regresses existing behaviour.**

---

### How this was tested
Manual QA in a throwaway Docker container built from the repo's own devcontainer image (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`), repo mounted, `gh` authenticated inside. The TUI was driven via tmux. The live purple badge only surfaces for a *running instance with a gh-authed container* (needs the devcontainer CLI / nested Docker, unavailable in the QA sandbox), so the **server-side probe** — the heart of this change — was exercised directly by extracting the exact `prProbeScript` from `internal/server/pr_status.go` and running it against **real GitHub PRs in this repo**.

### ✅ What works
- **Build & launch** — `go build ./...` and the `fleet` binary build cleanly; the TUI starts and renders.
- **Closed/merged → purple data** — branch `feat/tui-daemon-log-stream` (PR #210, merged, no open PR) → probe emits `{"number":210,"state":"MERGED",...}`. The parser buckets `MERGED`/`CLOSED` into `closedPRStatus`, producing `PrStatus{ClosedCount:1, PrSignal_PURPLE}` → the persistent purple `PR` badge. ✔ matches issue #203.
- **Highest-number pick** — branch `bb-test` has two closed PRs (#190, #192); probe returns **#192**, confirming `sort_by(.number) | last`.
- **Open priority** — this PR's own branch → full open-PR detail (`state:"OPEN"`, checks rollup); the closed fallback is not taken.

### ❌ Regression: non-zero probe exit discards valid PR status (multi-repo / no-PR branch)
**Root cause.** The new `else` branch ends with:
```sh
num=$(gh pr list --state all --head "$br" --json number,state --jq '... | last | .number // empty' 2>/dev/null)
[ -n "$num" ] && gh pr view "$num" --json number,state,url,title 2>/dev/null
```
When a repo's current branch has **no open PR and no closed/merged PR** ("never had a PR"), `$num` is empty, so `[ -n "$num" ]` fails and this compound command's exit status is **1**. Since it's the last command in the `find … | while` loop body, when that no-PR repo is the **last** repo `find` traverses, the **entire script exits 1** — even though other repos already printed valid PR JSON to stdout.

`probePRStatus` treats *any* non-zero exit as a transient failure:
```go
out, err := runProbeWithTimeout(cmd, prProbeTimeout)
if err != nil { return nil, false }   // ok=false
```
and `prStatusPass` then **keeps the prior PrStatus and throws the captured stdout away** (`if !r.ok { continue }`). `ExecCommandQuiet` runs the script via `devcontainer exec … sh -c`, which propagates the inner exit code, so this fires in production.

This also contradicts the function's own documented contract: *"A successful probe returns ok=true even when the PrStatus is nil (gh absent / no PRs), which legitimately clears the auto-tag."* On `main` the no-open-PR path ended in a zero-iteration `for` loop → exit 0, so this is a **new** behaviour introduced by this PR.

**Impact**
- **Multi-repo workspaces** (the code explicitly supports "workspace repo PLUS nested subrepos"): if the last-traversed repo has no PR — common, since most branches have no PR — the instance's PR status update is dropped on every probe. The tag **freezes at its last value and never updates**, regressing *both* the new purple tag *and* the pre-existing open-PR green/yellow/red indicator.
- **Single-repo branch switch**: an instance that switches to a branch that never had a PR now exits 1 → `ok=false` → the old tag is **retained instead of cleared**.

**Reproduction** (exact `prProbeScript` from source, real PRs, run inside the QA container):
```
# main repo = no-PR branch (top level), nested subrepo = merged PR #210
find order:  ./sub/.git   ./.git        # the no-PR repo (.) is visited LAST
probe stdout: {"number":210,"state":"MERGED",...}   # valid JSON emitted
probe exit:   1                                     # ...but script exits non-zero
→ probePRStatus returns (nil,false); the #210 status is discarded, prior kept.
```
Single-repo control: branch `qa-never-had-pr-zzz123` → no output, **exit 1** (on `main` this path exited 0).
(When the no-PR repo happens *not* to be last, the script exits 0 and works — so the failure is layout-dependent but deterministic per workspace.)

**Suggested fix (any one):**
- Make the `else` branch always succeed, e.g. use an `if` so the no-match case exits 0:
  ```sh
  if [ -n "$num" ]; then gh pr view "$num" --json number,state,url,title 2>/dev/null; fi
  ```
  (or append `|| :` / a trailing `true`), and/or end the script with `exit 0`.
- Or have `probePRStatus` not treat a non-zero exit as failure when stdout parsed cleanly (reserve `ok=false` for genuine exec/timeout errors, matching the comment's intent).

No source was modified; this is behaviour-level QA only.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary:** Solid, well-scoped change. I checked out the head commit (`a811657`), and `go build ./...`, `go vet`, and `go test ./internal/server/... ./internal/tui/...` all pass. The design is clean — open PRs strictly take priority, the closed/merged fallback reuses the existing `Prs`/row-gating/click-to-open plumbing, and the new behavior is genuinely test-covered (parser bucketing, open-wins-over-closed, purple rendering, multi-PR `PRxN`, and an end-to-end list render).

The standout is the second commit's regression fix: the no-PR branch previously ended in `[ -n "$num" ] && gh pr view …`, so when the *last* repo `find` visited had no PR the whole probe exited non-zero and `probePRStatus` discarded already-emitted JSON, freezing the tag. Switching to `if` + a trailing `exit 0` (so a completed scan is always authoritative) is the right fix, and `TestPRProbeScript_AlwaysExitsZero` exercises the real script against a stubbed `gh` over a multi-repo no-PR layout — nice regression guard.

A few **non-blocking** observations:

1. **`gh pr list` default limit (see inline note).** The open path moved from `--state open` to `--state all`, so closed/merged PRs now share the implicit 30-row cap. Practically harmless for feature branches (gh returns newest-first, and the open PR is virtually always the newest), but an explicit `--limit` would make the "highest number = newest" assumption robust against a long-lived/reused branch.

2. **`closed_count` vs `open_count` counting semantics.** `open_count` is the total open PRs across all repos, whereas the closed path emits only the single highest-numbered closed PR *per repo* — so `PRx2` (closed) means "two repos each had a closed PR," not "two closed PRs on this branch." This is reasonable (you don't want `PRx15` from a repo's PR history) and matches the "most recent closed/merged one" comment, but the asymmetry is worth a one-line note in the proto/`closedPRStatus` doc so it's an obviously deliberate choice.

3. **Stale "open PRs" wording.** Now that the same plumbing carries closed refs, a few doc comments still say "open PRs" — `instancePRRefs` ("returns the open PRs"), `choosePRState` ("more than one open PR"), and `renderChoosePRDialog` ("one row per open PR"). Purely cosmetic; worth a quick sweep for consistency with the new behavior.

Nothing here blocks merge. Closed-branch `--head` matching (after `delete_branch_on_merge`) is the crux of the feature and appears to be what QA on #203 validated; just calling it out as the load-bearing assumption.

internal/server/pr_status.go
```diff
@@ -65,12 +69,34 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
   [ -z "$br" ] && continue
   [ "$br" = "HEAD" ] && continue
   ( cd "$dir" || exit 0
-    for num in $(gh pr list --state open --head "$br" --json number --jq '.[].number' 2>/dev/null); do
-      gh pr view "$num" \
-        --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,url,title 2>/dev/null
-    done
+    # One list call for the branch (gh's embedded --jq, so no standalone jq
+    # dependency): "<STATE> <number>" per PR, e.g. "OPEN 12" / "MERGED 7".
+    list=$(gh pr list --state all --head "$br" --json number,state --jq '.[] | "\(.state) \(.number)"' 2>/dev/null)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Switching from `--state open` to `--state all` means closed/merged PRs now share `gh pr list`'s implicit default limit (30 rows). The `else`-branch then picks the newest closed PR by `max($2)`, which is correct — but if a branch has been the head of >30 PRs over its life, an *open* PR could in theory fall outside the returned window and be missed, silently degrading to the purple tag.

In practice this is very unlikely (feature branches rarely accumulate that many PRs, and gh returns newest-first so the open PR is almost always present), so this is a low-priority hardening suggestion rather than a bug. Adding an explicit `--limit` (and/or `--state open` short-circuit before the `--state all` fallback) would make the "highest number = newest open" assumption robust regardless of branch history.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `go build ./...` + `fleet` binary + `go vet` | ✅ Pass | All compile clean; proto regen (`PR_SIGNAL_PURPLE`, `closed_count`) links; vet 0 findings |
| TUI launch & render | ✅ Pass | Dep-check + main "Armada" view render, no crash |
| Closed-only → highest-number pick (`bb-test` #190/#192) | ✅ Pass | Emits **#192** only (newest), `state:CLOSED`, minimal fields, exit 0 → purple |
| Merged-only fallback (`feat/tui-daemon-log-stream` #210) | ✅ Pass | Emits #210 `state:MERGED`, no check detail, exit 0 → purple |
| Open PR priority (`fix/pr-status-track-closed` #215) | ✅ Pass | Full open detail (rollup/review/mergeState), closed fallback **not** taken |
| Never-had-a-PR branch (single repo) | ✅ Pass | Empty stdout, **exit 0** → clears tag (was the regression; now fixed) |
| **Regression re-test** (prior QA #1 finding) | ✅ Pass | Buggy commit `279a521` exits **1**; fixed HEAD `a811657` exits **0** on same branch |
| Multi-repo, no-PR repo traversed **LAST** | ✅ Pass | Valid #210 JSON emitted **and** exit 0 (old script exited 1 & discarded it) |
| Mixed multi-repo (open #215 + merged #210) | ✅ Pass | Both emitted; parser keeps open #215, purple suppressed (open wins) |
| `gh` absent | ✅ Pass | `FLEET_NO_GH`, exit 0 → graceful degrade (nil → clears tag) |
| `gh` unauthenticated | ✅ Pass | `FLEET_NO_AUTH`, exit 0 → graceful degrade |

**Overall: ✅ PASS — 11 of 11 scenarios passing. The prior-QA multi-repo / no-PR exit-code regression is fixed, and the headline closed-PR purple feature works.**

---

### How this was tested
Manual, behavior-level QA in a throwaway Docker container built from the repo's own devcontainer image (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`), repo mounted, `gh` authenticated inside. The container was driven via `docker exec` (TUI via tmux). The change's core is the **server-side `gh` probe** in `internal/server/pr_status.go`; I exercised the **exact `prProbeScript`** — extracted verbatim from the source raw-string literal — against **real GitHub PRs in this repo**, capturing both stdout and exit code for every path. Build, `go vet`, the `fleet` binary, and a TUI launch/render were also verified.

> **Scope note (transparency):** the *live* purple `PR` badge in the TUI cannot be exercised end-to-end here — it requires a running instance with a gh-authed devcontainer (devcontainer CLI + nested Docker), which is unavailable in the QA sandbox (the TUI's own dependency check reports `devcontainer CLI — not found`). The probe (which computes the data the badge renders) was tested live; the data→render mapping is covered by inspection and the PR's own unit tests, and the build confirms the new proto constants link.

### ✅ Headline feature (issue #203)
- **Closed/merged → persistent purple data.** `bb-test` (closed #190/#192) → probe emits the **newest** closed PR `#192` (`sort_by number | last`), and `feat/tui-daemon-log-stream` (merged #210, no open) → emits `#210` `state:MERGED`. The parser buckets `CLOSED`/`MERGED` into `closedPRStatus` → `PrStatus{ClosedCount:1, PrSignal_PURPLE, Prs:[…]}` → a bare, clickable purple `PR` badge (256-colour `99`). ✔ matches the ticket: a finished instance stays distinguishable from one that never had a PR.
- **Open priority intact.** This PR's branch → full open-PR detail (`statusCheckRollup`/`reviewDecision`/`mergeStateStatus`); the closed fallback is not taken. In a mixed multi-repo (open #215 + merged #210) the open status wins and purple is suppressed — the documented "open always takes priority" contract.

### ✅ Regression from prior QA is fixed
The previous QA run failed this PR on a **non-zero probe exit discarding valid PR JSON** when a no-PR repo was traversed last (script exited 1 → `probePRStatus` returned `ok=false` → valid JSON dropped, prior tag frozen). The new commit `a811657` rewrites the `else` branch to `if [ -n "$num" ]; then … fi` and ends the script with an explicit `exit 0`. Verified directly:

```
# never-had-a-PR branch, single repo
OLD (279a521): exit 1     NEW (a811657): exit 0

# multi-repo, no-PR repo visited LAST (parent=merged #210, nested sub=no-PR)
find order:  ./.git  ./sub/.git        # no-PR repo is last
OLD (279a521): exit 1, #210 JSON emitted → would be discarded (ok=false), tag freezes
NEW (a811657): exit 0, #210 JSON emitted → flows through (ok=true), tag updates ✔
```

A genuinely empty result (no PR anywhere) now exits 0, so the auto-tag is **cleared** rather than frozen — matching `probePRStatus`'s documented contract (`ok=false` reserved for real exec/timeout errors).

### ✅ Graceful degrade
`gh` absent → `FLEET_NO_GH` (exit 0); `gh` unauthenticated → `FLEET_NO_AUTH` (exit 0). Both map (by prefix) to a nil `PrStatus` → the tag clears, preserving prior behaviour.

No source was modified; this is behaviour-level QA only.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the full diff for #215 (closed/merged PRs keep a persistent purple tag). This is a well-scoped, well-documented change and I verified the riskier parts locally — overall it looks solid. A couple of low-priority observations, no blockers.

**Verified ✓**
- **Proto is safe.** The hand-edited `fleetgrpc/runtime.pb.go` is **byte-for-byte identical** to a fresh `buf generate` (ran it against the committed file), so despite the `// DO NOT EDIT` header there's no generator drift. Additions are backward-compatible: `PR_SIGNAL_PURPLE = 4` and `PrStatus.closed_count = 8` (field 8 was previously unused). The raw descriptor length/varint edits are internally consistent and the message descriptor inits without panic.
- **Tests pass** for `internal/server`, `internal/tui`, and `fleetgrpc`. Nice coverage — `TestPRProbeScript_AlwaysExitsZero` is a good guard for the #203 exit-status regression, and the open-wins-over-closed precedence is exercised end-to-end.
- **`awk` dependency is consistent with conventions** — `internal/backend/tool_probe.go` already runs `awk` inside the container, and the new usage sticks to portable features (no interval-regex `{n}` that would trip mawk, per the caveat called out in `codex_script.go`).
- **All tag gate sites flow through `instanceAutoTag`** — the inline-PR-row gating in `page_fleet_rows.go` keys off it, so closed-only status correctly still gets an inline row.

**Minor observations (non-blocking)**

1. **Case-sensitivity asymmetry between the shell and Go state checks.** The probe's `awk '$1 == "OPEN"'` (pr_status.go:78) is an exact-case match, while the Go parser defensively normalizes with `strings.ToUpper(p.State)` (and the rest of the file uses `ToUpper`/`EqualFold` throughout). `gh`'s JSON state is stably uppercase, so this is theoretical — but if that ever changed, the failure mode is asymmetric and worse than before: an *open* PR whose state didn't match `"OPEN"` would fall through to the closed branch and render as a purple "closed" badge rather than its live status. Matching the Go side's defensiveness (e.g. `toupper($1) == "OPEN"`) would make the two paths agree. Inline note added at the line.

2. **A transient `gh pr view` failure can briefly flip a live tag to purple.** Per-repo `gh` errors are swallowed with `2>/dev/null` and the script always `exit 0` (correctly — that's the #203 fix). But consider repo A with an open PR + repo B with a closed PR: if `gh pr view` for A's open PR fails transiently, stdout carries only B's closed object, so the probe is treated as authoritative and the instance shows the purple "closed" tag instead of its real open status. It self-corrects on the next probe (~45s), and the old code had the same class of issue (it would clear the tag); the closed-fallback just makes the transient wrong-state more visible. Not worth holding the PR, but flagging since the new code slightly changes the failure presentation.

Both are edge cases; the core behavior, tests, and proto are in good shape.

internal/server/pr_status.go
```diff
@@ -65,12 +69,37 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
...
-      gh pr view "$num" \
-        --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,url,title 2>/dev/null
-    done
+    # One list call for the branch (gh's embedded --jq, so no standalone jq
+    # dependency): "<STATE> <number>" per PR, e.g. "OPEN 12" / "MERGED 7". The
+    # explicit --limit lifts gh's default 30-row cap so an open PR can't fall
+    # outside the window (and silently degrade to the purple tag) on a branch
+    # that's been the head of many PRs over its life.
+    list=$(gh pr list --state all --head "$br" --limit 100 --json number,state --jq '.[] | "\(.state) \(.number)"' 2>/dev/null)
+    open=$(printf '%s\n' "$list" | awk '$1 == "OPEN" { print $2 }')
```
BenjaminBenetti: 🔍 **CodeReview Agent**

**Case-sensitivity asymmetry (low priority).** This `awk` matches `OPEN` exactly, but the Go parser normalizes with `strings.ToUpper(p.State)` (and the rest of this file uses `ToUpper`/`EqualFold`). `gh` returns stably-uppercase states today, so this works — but the two paths disagree on robustness. If `gh`'s casing ever changed, an *open* PR here would fail the `== "OPEN"` test, fall into the `else` branch, and render as a purple **closed** badge instead of its live open status — a more misleading failure than simply vanishing. Consider `toupper($1) == "OPEN"` (and `toupper($1) != "OPEN"` in the closed selector below) to match the Go side.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behavior-level QA of #215 (closes #203) — keep the auto-tag PR indicator visible as a persistent **purple `PR`** badge once a branch's only PR(s) are **closed/merged**, instead of vanishing. Tested at head `51646bd` inside a throwaway `mcr.microsoft.com/devcontainers/go:2-1.25-bookworm` Docker container.

| Test | Result | Notes |
|------|--------|-------|
| Build gate (`go build ./...` + `fleet` binary) | ✅ Pass | Compiles clean; `fleet --help` runs |
| TUI launch & dashboard render | ✅ Pass | Dependency-check → Armada dashboard, no crash; quits cleanly on `q` |
| Closed/merged PR → purple-tag fallback (probe) | ✅ Pass | Branch of merged PR #214 now surfaces `{state:MERGED, number:214}` (was: nothing) |
| Open PR → live status keeps priority (probe) | ✅ Pass | Open PR #216 emits full open detail; no purple fallback |
| Branch that never had a PR → no tag | ✅ Pass | Empty probe output, exit 0 → tag stays cleared (no false badge) |
| Parse→render wiring (purple=99, bare badge, clickable, `PRxN`) | ✅ Pass | Verified against generated proto + render source |

**Overall: ✅ PASS — 6 of 6 scenarios passing.**

---

### How this was tested

The user-facing badge is driven by a server-side `gh` probe (`internal/server/pr_status.go`) whose output feeds `parsePRProbeOutput` → `PrStatus` → the TUI `instanceAutoTag` render. I exercised the real change end-to-end: built and launched the actual TUI binary, then drove the **exact `prProbeScript` from the PR** (extracted byte-for-byte from the source, 2414 bytes) against **real GitHub PR data** in this repo, using a fresh clone and local branches named after real PR head refs.

### Detailed findings

**1. Build gate — ✅ Pass**
`go build ./...` and `go build -o fleet ./cmd/fleet` both succeed; `fleet --help` lists subcommands. (Note: a `-buildvcs` "dubious ownership" error appears only because the repo is bind-mounted from another UID — an environment artifact, resolved with `git config --global --add safe.directory`; not a code issue.)

**2. TUI launch & dashboard render — ✅ Pass**
Launched `fleet` in a tmux session inside the container. It rendered the Dependency-Check screen, then on `enter` the **Armada [ local ]** dashboard ("No instances…"). No panic/error; `q` exits cleanly. (devcontainer CLI is absent in the container, so no live instances were created — see limitation below.)

**3. Closed/merged PR → purple fallback — ✅ Pass (the fix)**
On a branch named `fix/session-group-prefix-isolation` (head ref of **merged PR #214**), the probe emitted exactly:
```json
{"number":214,"state":"MERGED","title":"fix(tui): stop a session group swallowing…","url":"https://github.com/BenjaminBenetti/fleet-man/pull/214"}
```
This is the new behavior — pre-fix this branch produced no output (badge vanished). `parsePRProbeOutput` buckets this as closed → `closedPRStatus` → `PrStatus{ClosedCount:1, PrSignal:PURPLE, Prs:[#214]}`, which the render gate (`auto_tag.go:57`, now `ClosedCount>0`) keeps visible as a purple `PR`. Exit 0.

**4. Open PR keeps priority — ✅ Pass (no regression)**
On `fix/remote-indicator-all-connections` (head of **open PR #216**), the probe emitted the full open-PR detail (`state:OPEN`, `mergeStateStatus:UNSTABLE`, `statusCheckRollup` with the live CI runs, etc.) and did **not** fall back to any closed PR. Open PRs still drive the green/yellow/red live tag.

**5. Branch with no PR history — ✅ Pass (no false positive)**
On a synthetic `qa-nonexistent-branch-zzz`, the probe produced **empty** output and exited 0 → `PrStatus` nil → tag stays cleared. A branch that never had a PR is still distinguishable from a finished one.

**6. Parse→render & proto wiring — ✅ Pass**
Confirmed against the generated contract + render source:
- `PR_SIGNAL_PURPLE = 4` → `prSignalStyle` → `prPurpleStyle` (256-colour `99`, GitHub's closed/merged purple).
- `PrReviewState` zero value is `UNSPECIFIED (0)`, **not** `PENDING (3)` — so a closed-only status (review unset, no checks) renders a **bare** purple `PR` with no stray "Pending"/"Checks" segments, matching the ticket.
- `closed_count = 8` is a backward-compatible field addition (field 8 was free).
- The badge stays clickable: `Prs` is populated with the closed ref, so `instancePRRefs`/`openSelectedPR` open the closed PR's URL; `PRx N` counting uses `ClosedCount` when `OpenCount==0`.

### Limitation / scope note
Rendering a populated, **coloured** badge in the *live* dashboard requires a running devcontainer instance whose in-container branch has a closed PR with `gh` auth — a nested-devcontainer setup that isn't reproducible in this CI sandbox (no devcontainer CLI / docker-in-docker). I therefore validated the badge via its **determining input** (the real `gh` probe, run end-to-end against live GitHub data) plus the deterministic parse/render path against the actual proto and render source, rather than a fully reproduced live instance. Every layer that *could* be exercised here behaved exactly as the ticket specifies.

BenjaminBenetti:
Thanks for the thorough passes. On the two latest non-blocking notes:

1. **Case-sensitivity (`awk '$1 == "OPEN"'` vs Go `ToUpper`)** — fixed in 9a73973: both awk checks now use `toupper($1)`, matching the parser's defensiveness, plus a lowercase-state case added to `TestPRProbeScript_AlwaysExitsZero`.

2. **Transient `gh pr view` failure briefly flipping a multi-repo open+closed instance to purple** — accepting this one. It self-corrects on the next ~45s probe, and the pre-existing code had the same class of transient (it would clear the tag). A real fix means threading a 'an open PR's detail fetch failed' sentinel through the probe so the parser can treat it as transient rather than authoritative — meaningful added complexity for a brief, self-healing edge. Happy to do it as a follow-up if you'd prefer.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary** — Clean, well-scoped fix for #203. The proto additions are backward-compatible (`closed_count` = field 8, `PR_SIGNAL_PURPLE` = 4), and `runtime.pb.go` matches a fresh `buf generate` exactly (no hand-edit drift). The open-PR path is genuinely untouched, and the "open PRs always win" priority is enforced in both the shell probe and the Go parser. Builds, `go vet`, and the `internal/server` + `internal/tui` suites all pass locally.

Test coverage is a highlight: the end-to-end `prProbeScript` test driving a stubbed `gh` — including the lowercase-state defensiveness and the always-exit-0 regression guard — locks in exactly the behavior that's easy to break later.

A few non-blocking considerations (details inline):

1. **Transient `gh` failures now clear the tag** instead of preserving the last-known state. The unconditional `exit 0` is the right fix for the no-PR case, but it also removes the signal that previously let a failing `gh pr view` hold the prior tag. See the inline note on `exit 0`.
2. **`closed_count`'s "PRxN" counts repos, not PRs**, which differs from the open badge's meaning of N. Documented and intentional, but a minor UX asymmetry — inline note.
3. Minor / pre-existing: `--head "$br"` for the closed fallback can match another fork's same-named branch in a shared upstream and then pick the highest *global* PR number for that head. The open path already did this, so it isn't introduced here — flagging only for awareness.

Overall this looks correct and ready to merge; the items above are judgment calls, not defects.

internal/server/pr_status.go
```diff
@@ -65,12 +69,40 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
...
+    fi
   )
 done
+# Always exit 0 once the scan completes: an empty result (a repo with no PRs) is a
+# valid probe outcome that should CLEAR the tag, not a transient failure. Without
+# this, a no-PR repo visited last by find leaves the loop's exit status at 1, and
+# probePRStatus would treat the whole probe as failed (ok=false) — discarding any
+# PR JSON already emitted and freezing the prior tag. A non-zero exit (ok=false,
+# keep prior) is reserved for genuine exec/timeout errors, which abort earlier.
+exit 0
```
BenjaminBenetti, line 105: 🔍 **CodeReview Agent**

The unconditional `exit 0` correctly fixes the no-PR regression, but note one side effect: it removes the only thing that distinguished a *transient `gh` failure* from a legitimate "no PR" result.

`gh pr list` / `gh pr view` errors are swallowed (`2>/dev/null`) and the script now always exits 0, so a momentary rate-limit / network blip inside the container produces **empty stdout → `probePRStatus` returns `(nil, true)` → the caller clears a previously-healthy tag**. `runProbeWithTimeout` only yields `ok=false` (keep-prior) on a process-level timeout/exec failure, never on gh's own non-zero exits.

On `main`, the last-visited repo's failing `gh pr view` propagated a non-zero exit, which preserved the prior tag. So the net change is: a brief GitHub hiccup can now blank a green/purple badge for up to one `prStatusInterval` (~45s) until the next probe succeeds.

Likely acceptable given the slow cadence (and the old protection was partial — last-repo-only). But if preserving the prior tag through transient gh errors matters, you'd want to distinguish "gh ran but errored" from "gh ran, no PRs" (e.g. emit a sentinel on a non-zero gh exit and map it to `ok=false`) rather than relying on the script's exit status alone.

internal/server/pr_status.go
```diff
@@ -238,18 +273,56 @@ func parsePRProbeOutput(out string) *fleetgrpc.PrStatus {
...
+// visible — so a finished instance is distinguishable from one that never had a
+// PR — and carries the closed PRs' refs so the tag stays clickable. Returns nil
+// when there are no closed/merged PRs either (the branch never had one).
+//
+// Note the per-repo counting: the probe emits only the single most-recent
+// closed/merged PR PER repo, so ClosedCount is the number of repos with a
+// closed-but-no-open PR (rendered "PRxN"), NOT the total closed PRs in any one
+// repo's history. This deliberately differs from open_count, which sums every
+// open PR — a branch's closed-PR history would otherwise inflate the badge.
+func closedPRStatus(closed []ghPR) *fleetgrpc.PrStatus {
```
BenjaminBenetti, line 309: 🔍 **CodeReview Agent**

Worth confirming this is the intended UX: `ClosedCount` is the number of *repos* that have a closed/merged-but-no-open PR (the probe emits only the single newest closed PR per repo), whereas `OpenCount` sums *every* open PR across repos. So the rendered "PRxN" means different things on the two paths — e.g. a single repo with three closed PRs on the branch shows a bare `PR`, while three open PRs show `PRx3`.

It's clearly documented here and in the proto, so this is mainly a flag for awareness: a reader comparing a purple `PRx2` against a green `PRx2` might read "2 PRs" in both cases, when the purple one actually means "2 repos". Low priority — only raising in case you'd prefer the closed badge to always render a plain `PR` to sidestep the ambiguity.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./...` + `fleet` binary) | ✅ Pass | Compiles clean in QA image (Go 1.25.8); binary runs |
| TUI launch & main view render | ✅ Pass | Dependency check + "Armada [ local ]" view render, no panic; quits clean |
| Probe — open-PR branch (#215 OPEN) | ✅ Pass | Emits full `state:"OPEN"` JSON w/ check rollup; open PRs still take priority |
| Probe — merged-only branch (#210 MERGED) | ✅ Pass | Emits lean closed ref (`number/state/url/title` only) → purple-tag data |
| Probe — closed/rejected branch (#177 CLOSED) | ✅ Pass | Emits lean `state:"CLOSED"` ref correctly |
| Probe — multi-closed branch (`bb-test`: #190 & #192) | ✅ Pass | Picks **highest** number #192, not list-order-dependent |
| Probe — never-had-a-PR branch | ✅ Pass | Empty output → tag stays cleared (stays distinct from closed) |
| Probe — gh-missing / no-auth sentinels | ✅ Pass | `FLEET_NO_GH` / `FLEET_NO_AUTH` → graceful degrade to no tag |
| awk `toupper()` case-defensiveness (commit `9a73973`) | ✅ Pass | Lowercase `open` still classified OPEN; old code would have shown false purple |
| Proto additions (`PR_SIGNAL_PURPLE=4`, `closed_count`=field 8) | ✅ Pass | Generated code compiles; values & field number match description (backward-compatible) |
| Purple badge live render in running TUI | ⏭️ Skipped | Requires a live fleet instance (DinD devcontainer) whose branch has a closed PR — verified via the data path + code review instead (see below) |

**Overall: ✅ PASS — 10 of 10 executed scenarios pass; 1 scenario skipped (infra-bound, covered indirectly). No defects or regressions found.**

---

## How this was tested

All execution was isolated in a throwaway Docker container built from the repo's own QA image (`ghcr.io/benjaminbenetti/fleet-man/qa:latest`), with the PR head (`9a73973`) checked out and `gh` authenticated. No automated unit/integration suites were run — this is behaviour-level QA driving the real shipped artifacts. The ticket (#203) asks that a closed PR's indicator **persist in purple** (matching GitHub) instead of vanishing, while a branch that *never* had a PR shows **no** indicator.

## Detailed findings

**Build & launch.** `go build ./...` and `go build -o /tmp/fleet ./cmd/fleet` both succeed. Launching the binary in a headless tmux renders the dependency-check screen, then the main `Armada [ local ]` fleet view, and quits cleanly — no panic / nil-pointer in the changed rendering path.

**Probe behaviour (the core change).** The exact `prProbeScript` was extracted verbatim from `internal/server/pr_status.go` and run against **real** GitHub PRs by checking out local branches whose names match real PR head refs:

- **Open (#215)** → emits the full `gh pr view` object with `"state":"OPEN"`, `statusCheckRollup`, `mergeStateStatus`, etc. → parser buckets as open → live green/yellow/red tag (unchanged behaviour, open takes priority). ✅
- **Merged-only (#210)** → `{"number":210,"state":"MERGED","title":...,"url":...}` — exactly the lean payload `closedPRStatus` needs for the persistent purple `PR` badge. ✅
- **Closed/rejected (#177)** → same lean shape with `"state":"CLOSED"`. ✅
- **Multi-closed (`bb-test` has #190 + #192, both CLOSED)** → probe emits **#192** (highest number), confirming it doesn't rely on `gh`'s list ordering, as the description claims. ✅
- **Never had a PR** (synthetic branch) → empty stdout → `nil` PrStatus → no tag. This is the key distinction the ticket asked for: "closed PR" (purple) vs "never had one" (nothing). ✅
- **Degrade paths** → `FLEET_NO_GH` when `gh` is absent and `FLEET_NO_AUTH` when unauthenticated, both prefix-matched to `nil` (no tag). ✅

**`toupper()` fix (latest commit `9a73973`).** Drove the actual awk one-liners with a lowercase `"open 12"` state line. New code → `open=[12] closed=[]` (correctly OPEN). The pre-fix `$1=="OPEN"` form → `open=[] closed=[12]`, i.e. it would have *missed* the open PR and wrongly rendered a purple closed tag. Uppercase `OPEN/MERGED/CLOSED` still classify correctly. The fix does what it says. ✅

**Proto.** Generated `runtime.pb.go` contains `PrSignal_PR_SIGNAL_PURPLE = 4` and `ClosedCount` on field `8` (a previously-free field), matching the `.proto`. The whole tree compiles, so the TUI's references to `PR_SIGNAL_PURPLE`, `GetClosedCount()` and `prPurpleStyle` (256-colour `99`) are wired up.

### Note on the skipped scenario
Rendering the **live** purple badge in the TUI requires a running fleet instance (a real devcontainer, via Docker-in-Docker) whose workspace branch has a closed-but-no-open PR — a heavy end-to-end setup whose failure modes are mostly unrelated to this change. Instead, the **data-production path** (the probe) was verified end-to-end against real GitHub for every case, and the **render path** was verified by inspection: `instanceAutoTag` derives its count from `ClosedCount`, the row gate keeps the tag when `closed_count > 0`, a closed-only status renders a bare purple `PR` (no review/checks segments), and the refs stay populated so click-to-open still works. Both halves are sound; only the final ANSI paint wasn't exercised on a live instance.

**Conclusion:** the change meets the acceptance criteria from #203 — closed/merged PRs keep a persistent (purple) indicator, branches that never had a PR show none, and open-PR behaviour is untouched. Recommend **QA pass**.

PR #216, opened 2026-06-25
fix/remote-indicator-all-connections: show remote indicator for gRPC and webhook

### LINKED ISSUE #199: Fleet TUI - remote connection indicator bug
The remote connection indicator (that looks like a wifi signal) is only linked to remote MCP. So it does not show up when GRPC or webhook are enabled.

### Desire State 
The remote connection indicator should display in the TUI if any of the 3 remote connections are enabled MCP, GRPC, Webhook.

BenjaminBenetti:
## Summary

The remote-connection "wifi" indicator (`·))` next to the fleet logo) was only shown when **remote MCP** (`RemoteMcpSettings.Enabled`) was on. It stayed hidden when only **remote gRPC** (`FleetEnabled`) or the **automation webhook** (`WebhookEnabled`) was enabled — even though all three ride the same outbound gateway tunnel.

This changes `remoteIndicator` to display whenever **any** of the three remote surfaces is enabled. The connection colour is unchanged (green when the shared tunnel is `CONNECTED`, red while connecting/errored), since a single tunnel carries all three.

Adds `TestRemoteIndicator` covering each surface individually, the all-off and nil-config hidden cases, and the green/red connection states.

Fixes #199


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary:** Clean, well-scoped fix for #199. The broadened condition (`!rs.Enabled && !rs.FleetEnabled && !rs.WebhookEnabled`) correctly mirrors the tunnel manager's own gating in `internal/server/remote/manager.go` — `desiredState.on()` brings the outbound tunnel up while **any** of `mcp || grpc || webhook` is set. Since all three ride the one shared tunnel, a single CONNECTED/red signal is the right model. `go test ./internal/tui/...` and `go vet` both pass on the head commit.

A few non-blocking observations:

**1. Indicator vs. tunnel `on()` differ by the `GatewayURL` check (minor UX).** `desiredState.on()` is `(mcp || grpc || webhook) && gatewayURL != ""`, but `remoteIndicator` only checks the three booleans. So when a surface is toggled on but **no gateway URL is configured**, the manager treats it as a no-op and never dials — yet the indicator renders persistently **red** ("connecting/error"). This is pre-existing behavior (the old `Enabled`-only path had the same gap), but the change makes it reachable from two more toggles. Optional: also gate on `rs.GatewayURL != ""` so the red glyph only shows when the tunnel is genuinely attempting to connect. See the inline note.

**2. Optional DRY.** The "any remote surface enabled" predicate now lives inline here and conceptually duplicates `desiredState.on()`. A small `RemoteMcpSettings` helper (e.g. `AnyRemoteEnabled()`) would centralize it and keep the indicator in sync if a fourth surface is added later. Low priority.

**Nice:** the test renders `errorStyle`/`statusRunningStyle` dynamically rather than hardcoding ANSI, so it stays robust to style tweaks, and it covers each surface individually plus the nil-config / all-off / green / red states — the gRPC and webhook green cases would correctly fail against the old MCP-only code.

internal/tui/render.go
```diff
@@ -161,12 +161,18 @@ func agentToolLabelProto(tool fleetgrpc.AgentTool) string {
...
+// webhook (WebhookEnabled). It is green once the tunnel is CONNECTED and red
+// while enabled but not yet up (connecting / error). When all three are
+// disabled it returns "" so the header renders unchanged.
 func remoteIndicator(m *model) string {
-	if m.config == nil || !m.config.RemoteMcpSettings.Enabled {
+	if m.config == nil {
+		return ""
+	}
+	rs := m.config.RemoteMcpSettings
+	if !rs.Enabled && !rs.FleetEnabled && !rs.WebhookEnabled {
```
BenjaminBenetti: 🔍 **CodeReview Agent**

The tunnel manager only dials when a gateway URL is set: `desiredState.on()` is `(mcp || grpc || webhook) && gatewayURL != ""`. This condition omits the `GatewayURL != ""` part, so a surface enabled with an empty/unconfigured `GatewayURL` will render this glyph **red forever** even though the manager never attempts a connection (it's a documented no-op). Pre-existing for the `Enabled` path, but now reachable from `FleetEnabled`/`WebhookEnabled` too.

Optional: `if !rs.Enabled && !rs.FleetEnabled && !rs.WebhookEnabled || rs.GatewayURL == ""` to keep the indicator's "should be lit" notion aligned with the tunnel's `on()`.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & launch TUI | ✅ Pass | Builds & TUI renders the fleet view cleanly (see build note below) |
| All remote off (no `config.json`) | ✅ Pass | Header logo renders with **no** `·))` indicator |
| `remote_mcp_settings: {}` (all flags false) | ✅ Pass | Indicator hidden, as expected |
| MCP only (`enabled`) | ✅ Pass | `·))` indicator shown |
| **gRPC only (`fleet_enabled`)** | ✅ Pass | `·))` shown — **was hidden before this fix (#199)** |
| **Webhook only (`webhook_enabled`)** | ✅ Pass | `·))` shown — **was hidden before this fix (#199)** |
| All three enabled | ✅ Pass | `·))` shown |
| Indicator colour (no live tunnel) | ✅ Pass | Renders red (`errorStyle`/256-colour 196) = "enabled but not connected" |
| Dynamic toggle **ON** via Settings | ✅ Pass | Starting from a daemon with everything off, enabling **Webhook** made `·))` appear live |
| Dynamic toggle **OFF** via Settings | ✅ Pass | Disabling it again removed `·))` live |
| Settings page renders / navigable | ✅ Pass | Full settings page (incl. Fleet MCP toggles) renders, no crash |

**Overall: ✅ PASS — 11 of 11 scenarios passing.** The fix fully resolves #199.

---

### How it was tested
All execution was isolated in a throwaway Docker container (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`) with the PR branch (`fix/remote-indicator-all-connections`, commit `3005a42`) mounted. The `fleet` TUI was built and driven inside the container via `tmux` (140×40), capturing the rendered pane after each step. Each scenario used an isolated `$HOME` so its auto-spawned `fleetd` read its own `~/.fleet/config.json`. The TUI reads config from the daemon over gRPC, so each `config.json` exercises the real load path.

### Detailed findings

**1. Static config scenarios (indicator presence)** — for each I wrote a `config.json`, launched the TUI, and inspected the logo's first line:

- `off` (no config) → `  __ _         _` — **no** indicator. ✅
- `{"remote_mcp_settings":{}}` → no indicator. ✅
- `{"remote_mcp_settings":{"enabled":true}}` → `  __ _         _    ·))`. ✅
- `{"remote_mcp_settings":{"fleet_enabled":true}}` → `·))` shown. ✅ *(the regression in #199)*
- `{"remote_mcp_settings":{"webhook_enabled":true}}` → `·))` shown. ✅ *(the regression in #199)*
- all three `true` → `·))` shown. ✅

**2. Colour** — captured ANSI shows the glyph rendered with `\e[38;5;196m` (red, `errorStyle`), the correct "enabled but tunnel not yet CONNECTED" state given no real gateway is configured. `statusRunningStyle` (green, 42) would require a live CONNECTED tunnel, which isn't reachable in an offline QA box; the green path is covered by the added unit test.

**3. End-to-end user flow (the real #199 repro)** — booted a daemon with **all remote surfaces off** (indicator correctly absent), opened **Settings → Fleet MCP**, toggled **Enable Webhook → on** (status line: `Webhook set to on`), returned to the fleet view → `·))` **appeared**. Toggled it back off (`Webhook set to off`) → indicator **disappeared**. Confirms the fix works dynamically, not just at startup.

**4. Settings UI sanity** — the Settings page renders fully and the three toggles map 1:1 to the config fields verified above: **Enable Remote MCP** (`enabled`), **Enable Remote Fleet** (`fleet_enabled`), **Enable Webhook** (`webhook_enabled`). Navigation and the armada-switch dialog also work; nothing obviously regressed.

> **Build note (environment, not a defect):** `go build ./cmd/fleet` initially failed with `error obtaining VCS status: exit status 128` because the repo is bind-mounted into the container under a different uid (git "dubious ownership"). Building with `-buildvcs=false` succeeded. This is a side effect of the QA mount setup, not an issue with the PR — normal devcontainer/CI builds are unaffected.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary:** Tight, well-scoped fix for #199. I checked out the head, applied the diff, and ran the affected packages locally — `go vet ./internal/state ./internal/tui` is clean and both new suites (`TestTunnelDesired`, `TestRemoteIndicator`) pass.

**Correctness — verified ✅**
- `RemoteMcpSettings.TunnelDesired()` faithfully mirrors `desiredState.on()` in `internal/server/remote/manager.go`: `(mcp || grpc || webhook) && gatewayURL != ""`. The `strings.TrimSpace` on the URL also matches `Manager.Reconcile`, which trims before storing into `desired` — so the indicator lights exactly when the tunnel will actually dial. Requiring the gateway URL is the right call.
- Worth flagging: this is also a (correct) behavior change for the **MCP-only** case. Previously `Enabled` alone lit the indicator even with no gateway URL; now it stays dark. That matches reality (the tunnel is a documented no-op without a URL), so 👍 — just calling it out in case any QA/screenshot expectation assumed the old behavior.

**Maintainability (minor, non-blocking):** `TunnelDesired()` and `desiredState.on()` are now two independent copies of the same rule living in different packages. The doc comment + a test on each side mitigate drift, but nothing structural ties them together — adding a 4th remote surface would require editing both in lockstep. See the inline note for a suggestion.

**Test coverage (optional nits):** Coverage of each surface and the hidden/nil cases is solid. Two small gaps if you want belt-and-suspenders:
- a case with multiple surfaces enabled simultaneously (e.g. `FleetEnabled && WebhookEnabled`);
- the explicit `REMOTE_MCP_CONN_ERROR` state — currently only `CONNECTING` and `nil` exercise the red branch; `ERROR` falls through the same default branch but isn't asserted.

Clear comments and good tests overall. No blocking concerns.

internal/state/remote_mcp_settings.go
```diff
@@ -39,3 +41,14 @@ type RemoteMcpSettings struct {
...
 }
+
+// TunnelDesired reports whether the outbound gateway tunnel should be up: any of
+// the three remote surfaces (MCP / gRPC / webhook) is enabled AND a gateway URL
+// is configured. It mirrors the manager's desiredState.on()
+// (internal/server/remote/manager.go) so the TUI's remote-connection indicator
+// lights exactly when the tunnel will actually dial — without a gateway URL the
+// tunnel is a documented no-op, so the indicator stays dark rather than showing
+// a misleading "connecting" state forever.
+func (s RemoteMcpSettings) TunnelDesired() bool {
```
BenjaminBenetti, line 53: 🔍 **CodeReview Agent**

This predicate duplicates `desiredState.on()` in `internal/server/remote/manager.go` (`(mcp || grpc || webhook) && gatewayURL != ""`, with the same URL trim done in `Manager.Reconcile`). The mirror is correct today and both sides are pinned by tests, but they can silently drift — e.g. adding a 4th remote surface, or changing the "what counts as configured" rule, requires editing both places. Consider giving the manager a single source of truth (e.g. deriving its `desiredState` from `RemoteMcpSettings`, or having `on()` delegate to this method) so the TUI indicator and the actual dial can't disagree. Minor / non-blocking.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| App builds & TUI launches | ✅ Pass | `go build ./cmd/fleet` clean; TUI renders logo/menu, daemon auto-spawns, `q` quits cleanly |
| A — Default config (all remote surfaces off) | ✅ Pass | No `·))` indicator — header unchanged, as expected |
| B — Remote **MCP** only + gateway URL | ✅ Pass | Indicator **shown** — pre-existing behavior preserved |
| C — Remote **gRPC** (`fleet_enabled`) only + gateway URL | ✅ Pass | Indicator **shown** — this is the fix for #199 (was hidden before) |
| D — **Webhook** (`webhook_enabled`) only + gateway URL | ✅ Pass | Indicator **shown** — fix also covers the webhook surface |
| E — gRPC enabled but **no gateway URL** | ✅ Pass | Indicator **hidden** — no misleading "connecting" glyph |
| F — Gateway URL set but **all surfaces off** | ✅ Pass | Indicator **hidden** |
| G — Surface enabled + **blank/whitespace** gateway URL | ✅ Pass | Indicator **hidden** — `TrimSpace` gate works |
| Indicator colour when tunnel not up | ✅ Pass | Glyph renders **bold red (256-colour 196)** while connecting/errored against an unreachable gateway |
| Settings page UI sanity | ✅ Pass | Fleet MCP section shows the 3 independent toggles + Gateway URL; seeded config reflected; no UI regressions |

**Overall: ✅ PASS — 10 of 10 scenarios passing. The remote indicator now appears for gRPC and webhook (not just MCP) and stays hidden without a configured gateway URL, matching the ticket (#199).**

---

### Method
Built and ran the real `fleet` TUI **inside a throwaway Docker container** (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`), driven via `tmux` (`send-keys` / `capture-pane`). No automated unit/integration tests were run — each scenario seeds `~/.fleet/config.json`, launches the TUI (which auto-spawns the daemon that loads the config and serves it to the client over gRPC), and inspects the rendered header next to the fleet logo. The indicator glyph is `·))`.

This exercises the changed code paths end-to-end: `RemoteMcpSettings.TunnelDesired()` (visibility gate) and `remoteIndicator()` (header rendering).

### Detailed findings

**Build / launch.** `go build -o fleet ./cmd/fleet` succeeded. Launching `fleet` inside tmux renders the gradient logo, the Armada bar, and the menu; the auto-spawned daemon comes up (`~/.fleet/fleet.sock`), `fleet list` returns an empty fleet list, and `q` exits cleanly. No crashes or panics observed in any run.

**Core fix (C, D) — the indicator now appears for gRPC and webhook.** With only `fleet_enabled` (gRPC) **or** only `webhook_enabled` set, plus a `gateway_url`, the `·))` glyph appears off the top-right of the logo — exactly the behavior #199 asked for. Captured header (scenario C):
```
  __ _         _    ·))
 / _| |___ ___| |_
|  _| / -_) -_)  _|
|_| |_\___\___|\___|
```
Scenario B (MCP only) confirms the original MCP-triggered behavior is unchanged.

**Gating (E, F, G) — hidden when the tunnel would be a no-op.** When a surface is enabled but no gateway URL is configured (E), when a URL is set but every surface is off (F), or when the URL is only whitespace (G), the indicator stays hidden — the header renders identically to the all-off baseline (A). This matches the second commit's intent: don't show a permanently-"connecting" indicator when the tunnel will never dial.

**Colour.** Against an (intentionally) unreachable `https://gateway.example.com`, the tunnel never reaches CONNECTED, so the glyph renders **bold red** — raw capture shows `ESC[1m ESC[38;5;196m` (errorStyle, colour 196) on the glyph. This is the "red while desired but not yet up (connecting / error)" state. The green/CONNECTED colour path was not directly exercised because it requires a live, reachable fleet gateway; note that this PR does not change the colour logic (only the visibility gate), and the not-connected → red path verifies the styling branch.

**Settings UI sanity.** Opening the settings page with gRPC+webhook seeded shows `Enable Remote MCP [ off ]`, `Enable Remote Fleet [ on ]`, `Enable Webhook [ on ]`, and `Gateway URL https://gateway.example.com` — the three independent toggles and the shared gateway URL all render correctly, with no broader UI regression. (The `Public Webhook URL error … produced zero addresses` line is the expected consequence of the unreachable test gateway, not a defect.)

The container was torn down after testing.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the diff for #216 (fix the remote "wifi" indicator so it lights for gRPC/webhook, not just MCP — fixes #199). This is a clean, well-scoped, well-tested fix. I built it, ran `go vet`, and ran the affected packages' tests locally — all green. A few notes, only one of which is actionable.

### What's good
- **Single source of truth.** Centralizing the "tunnel is wanted" predicate in `RemoteMcpSettings.TunnelDesired()` and routing both the dial gate (`desiredState.on()`) and the TUI indicator through it is the right call — it removes exactly the drift the original bug came from.
- **The manager refactor is behavior-preserving.** `on()` now delegates to `TunnelDesired()`, which adds `strings.TrimSpace(GatewayURL)`. That's safe: `Reconcile` (manager.go:172) already trims `gatewayURL` before storing it in `desired`, so `on()` only ever sees a pre-trimmed value — the extra trim is a harmless safety net on that path and a genuine improvement on the TUI path (config may carry whitespace). The existing `on()` tests (manager_test.go:112–115) still pass.
- **The added `TrimSpace` also keeps the indicator honest** — without a gateway URL the tunnel is a documented no-op, so the glyph correctly stays dark instead of showing a perpetual red "connecting."
- **Test coverage is thorough**: each surface individually, multi-surface, all-off, nil-config, no-gateway-URL, and the green/connected vs. red connecting/error/nil-status branches. Green=`statusRunningStyle` (color 42) and red=`errorStyle` (color 196) match the comment.

### Scope check — the fix is complete for #199
I looked for sibling gates with the same `.Enabled`-only bug class and found none that need changing:
- The per-feature Public URL rows (`page_settings.go`) are each gated on their own toggle (MCP / `FleetEnabled` / `WebhookEnabled`) — correct.
- The Bearer Token row's `Enabled || FleetEnabled` (deliberately excluding webhook) is **intentional and correct**: the webhook receiver is documented as unauthenticated (the unguessable public URL is the capability), so it doesn't pair with the bearer token.

### Actionable
- **`internal/tui/render_test.go` is not gofmt-clean.** `gofmt -d` produces a diff — the struct keys in the table-test literals aren't column-aligned. CI won't catch this (the golangci-lint step only enables `depguard`, and there's no `gofmt`/`go test -run fmt` gate), but it's inconsistent with the rest of the tree. A quick `gofmt -w internal/tui/render_test.go` fixes it. Inline comment below.

### Optional / non-blocking
- The shared status type and field are still MCP-named (`remoteMcpStatus`, `RemoteMcpStatus`, `RemoteMcpConn_REMOTE_MCP_CONN_*`) even though, as this PR makes user-visible, the one tunnel now also carries gRPC and webhook. Pre-existing and out of scope here, but a future rename to something tunnel-centric would reduce confusion now that the multi-surface nature is surfaced in the UI.

Nice work — the only thing I'd actually change before merge is the gofmt nit.

internal/tui/render_test.go
```diff
@@ -4,9 +4,84 @@ import (
...
+	cases := []struct {
+		name   string
+		cfg    *state.Config
+		status *fleetgrpc.RemoteMcpStatus
+		want   string
+	}{
+		{name: "nil config hides indicator", cfg: nil, want: ""},
+		{name: "all remote surfaces off hides indicator", cfg: &state.Config{}, want: ""},
+		{
+			name: "enabled but no gateway url hides indicator",
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Nit (non-blocking, won't fail CI): this file isn't gofmt-clean. `gofmt -d internal/tui/render_test.go` wants the struct keys in these table-test literals column-aligned (e.g. `name:` → `name:  `, `cfg:` → `cfg:   ` so they line up with `status:`). A single `gofmt -w internal/tui/render_test.go` fixes the whole table. The rest of the changed files are already formatted.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & launch real TUI (Docker) | ✅ Pass | `go build ./cmd/fleet` clean; TUI renders, daemon auto-spawns |
| All remote surfaces OFF | ✅ Pass | No `·))` indicator (as expected) |
| Remote MCP only + gateway URL | ✅ Pass | Indicator shown — no regression of original behavior |
| Remote Fleet (gRPC) only + gateway URL | ✅ Pass | **Indicator now shown** — core fix for #199 |
| Webhook only + gateway URL | ✅ Pass | **Indicator now shown** — core fix for #199 |
| All three surfaces + gateway URL | ✅ Pass | Indicator shown |
| Surface enabled but NO gateway URL | ✅ Pass | Indicator hidden (gateway-URL gate) |
| Blank/whitespace gateway URL | ✅ Pass | Treated as unset → indicator hidden |
| Gateway URL set but all surfaces OFF | ✅ Pass | Indicator hidden |
| Live UI toggle (Enable Remote Fleet) | ✅ Pass | Toggling ON in Settings makes the indicator appear immediately on the main view |
| Indicator colour, gateway unreachable | ✅ Pass | Renders **red** (`color 196`, bold) while not-yet-connected, per spec |
| Settings page navigation/render | ✅ Pass | Remote MCP / Remote Fleet / Webhook / Gateway URL all render correctly |
| Clean quit (`q`) | ✅ Pass | TUI exits cleanly; daemon persists by design |
| Green/CONNECTED colour | ⏭️ Skipped | Needs a live gateway tunnel (not feasible in sandbox); colour branch is unchanged by this PR and covered by `TestRemoteIndicator` |

**Overall: ✅ PASS — 13 of 13 executed scenarios passing (1 skipped, not feasible in sandbox)**

---

### How this was tested
All execution was isolated in a throwaway Docker container (`golang:1.25-bookworm`, repo + module cache mounted). The `fleet` binary was built inside the container and the **real TUI** was driven via tmux (`send-keys` / `capture-pane`). For each scenario `~/.fleet/config.json` was seeded with the relevant `remote_mcp_settings`, the daemon was restarted so it re-read config fresh, and the rendered header line was inspected (plain + ANSI). The container was torn down afterward.

The wifi-style glyph appears appended to the first line of the `fleet` ASCII logo, e.g.:
```
  __ _         _    ·))
 / _| |___ ___| |_
|  _| / -_) -_)  _|
|_| |_\___\___|\___|
```

### Detailed findings

**Core fix (issue #199) — indicator now lights for gRPC and webhook.**
- `{"fleet_enabled":true,"gateway_url":"https://gateway.example.com"}` → `·))` shown (previously hidden — this was the bug).
- `{"webhook_enabled":true,"gateway_url":"…"}` → `·))` shown (previously hidden).
- `{"enabled":true,"gateway_url":"…"}` (MCP, the only surface that used to work) → still shown — no regression.
- `{"enabled":true,"fleet_enabled":true,"webhook_enabled":true,"gateway_url":"…"}` → shown.

**Gateway-URL gate (commit `96e4f0d`).** With a surface enabled but no usable gateway URL the indicator stays dark, so it never shows a permanent misleading "connecting" state:
- `{"fleet_enabled":true}` (no URL) → hidden.
- `{"enabled":true,"gateway_url":"   "}` (blank URL) → hidden.
- `{"gateway_url":"…"}` (URL only, all surfaces off) → hidden.

**Live end-to-end flow.** Started with `gateway_url` set + all surfaces OFF (no indicator). Opened **Settings**, toggled **Enable Remote Fleet → [ on ]**, returned to the main view — the `·))` indicator appeared immediately. This exercises the full path: settings toggle → `UpdateConfig` → daemon `Reconcile` (`desiredState.on()` now delegating to `RemoteMcpSettings.TunnelDesired()`) → re-render via `remoteIndicator`.

**Colour.** In every "shown" case the glyph rendered **red** (`errorStyle`, color `196`, bold), which is correct — the test gateway is unreachable so the shared tunnel is connecting/errored, never `CONNECTED`. The green (`statusRunningStyle`) branch requires a real connected tunnel and could not be reproduced in the sandbox; it is unchanged by this PR and is covered by the added `TestRemoteIndicator` unit test.

No defects or regressions observed. The change matches the ticket's desired state: the remote-connection indicator displays whenever any of the three remote surfaces (MCP / gRPC / webhook) is enabled and a gateway URL is configured.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Baseline — all remote surfaces off | ✅ Pass | No `·))` glyph; header renders unchanged |
| gRPC (`fleet_enabled`) + gateway URL | ✅ Pass | `·))` now shows — **core fix for #199** |
| Webhook (`webhook_enabled`) + gateway URL | ✅ Pass | `·))` now shows — **core fix for #199** |
| Remote MCP (`enabled`) + gateway URL | ✅ Pass | `·))` shows — existing behavior preserved |
| Multiple surfaces (gRPC + webhook) + gateway | ✅ Pass | `·))` shows |
| gRPC enabled, **no** gateway URL | ✅ Pass | Indicator hidden — no misleading "connecting" |
| Gateway URL set, all surfaces off | ✅ Pass | Indicator hidden |
| Whitespace-only gateway URL (`"   "`) | ✅ Pass | Indicator hidden (TrimSpace gate works) |
| Glyph color (not-connected) | ✅ Pass | Renders red via ANSI `38;5;196` = `errorStyle`, as expected |
| Glyph color (CONNECTED → green) | ⏭️ Skipped | Needs a live gateway tunnel, not reproducible in QA env. Connect-state→color logic is unchanged by this PR |
| TUI sanity sweep | ✅ Pass | Launch, daemon spawn, fleet list, Armada dialog, Settings page all render & respond |

**Overall: ✅ PASS — 9 of 9 testable scenarios pass; 1 skipped (green/connected state requires a live gateway).**

---

### Environment
- Checked out PR head `3a8e4d1` (matches the `synchronize` event SHA).
- All app execution isolated in a throwaway Docker container (`debian:bookworm-slim`); the `fleet` TUI was driven inside a container-side tmux session and the screen captured with `tmux capture-pane`. The auto-spawned `fleet server` daemon came up cleanly and served config to the TUI.
- Each scenario used a fresh `HOME` so the daemon read a fresh `~/.fleet/config.json`.

### What the ticket asks (#199)
The remote-connection "wifi" glyph (`·))` next to the fleet logo) previously appeared **only** when remote MCP (`enabled`) was on, and stayed hidden when only remote gRPC (`fleet_enabled`) or the automation webhook (`webhook_enabled`) was enabled — even though all three ride the same outbound tunnel. The glyph should appear whenever **any** of the three is enabled (and a gateway URL is configured).

### Detailed findings

**Core fix verified.** With `fleet_enabled: true` + a gateway URL, the header's first logo line rendered:
```
  __ _         _    ·))
 / _| |___ ___| |_
|  _| / -_) -_)  _|
|_| |_\___\___|\___|
```
Identical result for `webhook_enabled: true` and for `enabled: true` (MCP), and for combinations — so all three surfaces now light the indicator. Before this PR, the gRPC-only and webhook-only cases left the header bare.

**Color check.** Inspecting the raw ANSI of the glyph in the gRPC scenario showed `ESC[1mESC[38;5;196m·))` — color **196 (red)** = `errorStyle`, the correct "desired but not yet connected" state (no real gateway was reachable in the QA env). The green/`statusRunningStyle` (color 42) CONNECTED branch could not be exercised without a live gateway tunnel, so it's marked Skipped; note that branch is untouched by this PR (only the *visibility* predicate changed).

**Gateway-URL gate verified.** The PR also gates the glyph on a configured gateway URL (`TunnelDesired`). Confirmed:
- gRPC enabled with empty `gateway_url` → glyph **hidden** (avoids a permanent misleading "connecting" red when the tunnel is a documented no-op).
- A whitespace-only `gateway_url` (`"   "`) → glyph **hidden** (TrimSpace).
- Gateway URL set but all three surfaces off → glyph **hidden**.

This is a deliberate, sensible behavior change: an MCP-enabled config with no gateway URL no longer shows the indicator, matching when the daemon actually dials.

**Daemon/manager alignment.** The shared `RemoteMcpSettings.TunnelDesired()` now backs both the TUI indicator and the daemon's `desiredState.on()`. The indicator going red whenever a surface + gateway is configured is consistent with the daemon attempting to bring the shared tunnel up; the tunnel's actual dial behavior is unchanged (the refactor only adds gateway-URL whitespace trimming).

**No regressions observed.** The fleet list page, the Armada switch dialog, and the Settings page all render and respond to input normally with the change applied.

### Reproduction (core fix)
1. `~/.fleet/config.json` → `{"remote_mcp_settings":{"fleet_enabled":true,"gateway_url":"https://gw.example.com"}}`
2. Launch `fleet` → the `·))` glyph appears (red) at the top-right of the logo.
3. Repeat with `webhook_enabled:true` (instead of `fleet_enabled`) → glyph appears.
4. Remove `gateway_url` → glyph disappears.

internal/state/remote_mcp_settings.go
```diff
@@ -39,3 +41,14 @@ type RemoteMcpSettings struct {
 	// Enabled/FleetEnabled — all three ride the SAME outbound tunnel.
 	WebhookEnabled bool `json:"webhook_enabled,omitempty"`
 }
+
+// TunnelDesired reports whether the outbound gateway tunnel should be up: any of
+// the three remote surfaces (MCP / gRPC / webhook) is enabled AND a gateway URL
+// is configured. It mirrors the manager's desiredState.on()
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Doc nit: this says `TunnelDesired` "mirrors the manager's `desiredState.on()`", but after this change `on()` *delegates to* `TunnelDesired()` — so this method is the source of truth and `on()` is the mirror. Consider rewording (e.g. "`desiredState.on()` delegates to this") to keep the direction accurate for the next reader.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Reviewed #216 — show remote indicator for gRPC and webhook**

Solid, well-scoped bugfix. I checked out the head (`9224ad9`) and confirmed `go build ./...`, `go vet`, and the affected-package tests all pass.

**What's good**
- **Single source of truth.** Routing both the dial decision (`desiredState.on()`) and the TUI indicator (`remoteIndicator`) through `RemoteMcpSettings.TunnelDesired()` means the two can't drift — exactly the failure mode behind #199.
- **No behavior change on the manager path.** `Reconcile` already stores `strings.TrimSpace(gatewayURL)` (`manager.go:172`), so the new `TrimSpace` inside `TunnelDesired()` is a no-op for the dial decision and only adds robustness on the TUI path (where the config value may be untrimmed). It tightens behavior without risking a dial/indicator mismatch.
- **No import cycle.** `internal/state` doesn't import `internal/server/remote`, so the new `state` import in the manager is safe.
- **Thorough tests** — each surface individually, all-off, nil-config, no-gateway-URL, blank-URL, plus green/connecting/error/nil-status. The expected styles (`statusRunningStyle` / `errorStyle`) match the real render path.
- I also checked the adjacent settings-page gates: each row already keys on its own surface, and the Bearer Token row correctly excludes webhook (the webhook receiver is unauthenticated by design). So the wifi indicator was the one genuinely-miswired surface — this completes the fix rather than papering over part of it.

**Minor (non-blocking)**
1. **Doc direction is now backwards.** `TunnelDesired`'s comment says it "mirrors the manager's `desiredState.on()`", but `on()` now *delegates to* `TunnelDesired()` — this method is the source of truth, not the mirror. Worth rewording (inline note left).
2. **Small new coupling.** `desiredState.on()` builds a throwaway `RemoteMcpSettings{...}` just to reach the predicate, adding a `server/remote → state` dependency. It reads fine and is self-documenting; a free `state.TunnelDesired(mcp, grpc, webhook bool, url string)` would avoid the struct literal, but the method form is reasonable. Your call.

Nothing here blocks merge. 👍

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Verdict: looks good — no blocking issues.** Clean, well-scoped fix for #199.

### What I verified
- **Correctness** — `remoteIndicator` now gates on `RemoteMcpSettings.TunnelDesired()` (any of MCP / gRPC / webhook enabled **and** a gateway URL set) instead of only `Enabled`, which is exactly the bug in #199. ✅
- **No drift** — Routing both `desiredState.on()` and the TUI indicator through the single `TunnelDesired()` predicate is the right call (DRY); the dial decision and the indicator can no longer disagree.
- **Behavior parity for the manager** — `Reconcile` already trims the gateway URL (`manager.go:172`) and `Config.Normalize` trims it on load (`config.go:51`), so the new `strings.TrimSpace` inside `TunnelDesired` is purely defensive — `on()` keeps its previous semantics. No regression.
- **No import cycle** — `internal/server/remote` importing `internal/state` is fine; `internal/state` does not import back into `server/remote`, and the build confirms it.
- **Completeness** — Swept every `RemoteMcpSettings.{Enabled,FleetEnabled,WebhookEnabled}` reference: the indicator was the only shared "is the tunnel up?" gate; the rest are legitimately per-surface (toggles, reconcile args, per-surface status rows). Nothing else needed the same change.
- **Build & tests** — `go build ./...` plus `go test` for `internal/state`, `internal/tui`, and `internal/server/remote` all pass on the head commit. Test coverage is thorough: each surface individually, all-off, nil-config, blank-URL, and the green/red/no-status connection states.

### Optional (non-blocking) observations
1. **Shared status, MCP-flavoured name.** The colour is still driven by `m.remoteMcpStatus`, whose type/field name predates this change but now represents the *shared* tunnel carrying all three surfaces. The behaviour is correct (one tunnel → one status), but the `Mcp` naming reads as MCP-specific. Worth a future rename/comment for clarity — out of scope here.
2. **Redundant trim, intentionally.** As noted above, `TunnelDesired`'s `TrimSpace` duplicates trimming already done at both call sites. That's reasonable belt-and-suspenders for a now-public predicate and it's pinned by the `"blank url counts as unset"` test — flagging only so it's a deliberate choice, not an accident.

Nice work on the focused fix and the table-driven tests mirroring the manager's predicate.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & launch TUI | ✅ Pass | Builds clean; daemon auto-spawns; dashboard renders, no stderr errors |
| MCP only + gateway URL → indicator shows | ✅ Pass | `·))` shown red (baseline / pre-existing behavior intact) |
| **gRPC (`fleet_enabled`) only + gateway URL → shows** | ✅ Pass | **Core fix (#199): indicator now shown red — was hidden before** |
| **Webhook (`webhook_enabled`) only + gateway URL → shows** | ✅ Pass | **Core fix (#199): indicator now shown red — was hidden before** |
| Multiple surfaces (gRPC + webhook) + gateway → shows | ✅ Pass | `·))` shown red |
| No config / all surfaces off → hidden | ✅ Pass | No glyph rendered, header unchanged |
| Surface on, **no gateway URL** → hidden | ✅ Pass | No-op tunnel correctly stays dark (no misleading "connecting") |
| Surface on, **blank/whitespace gateway URL** → hidden | ✅ Pass | `"   "` treated as unset (trimmed) — stays hidden |
| Gateway URL set, **all surfaces off** → hidden | ✅ Pass | URL alone does not light the indicator |
| Live Settings toggle: Remote Fleet ON → indicator appears | ✅ Pass | Toggled in Settings UI; dashboard lit up live (red) on return |
| Live Settings toggle: Remote Fleet OFF → indicator hides | ✅ Pass | Reactive both directions |
| CONNECTED → green (color 42) state | ⏭️ Skipped | Needs a real reachable gateway; not exercisable in sandbox (path unchanged by this PR — only the gating predicate changed) |

**Overall: ✅ PASS — 11 of 11 executed scenarios pass; 1 skipped (green/CONNECTED state requires a live gateway).**

---

### How this was tested

All execution was isolated in a throwaway Docker container (`golang:1.25-bookworm`, workspace mounted) — never in the agent's own session. The real `fleet` binary was built (`go build ./cmd/fleet`) and driven through `tmux` inside the container. Each case launched the actual TUI, which auto-spawned the `fleetd` daemon; the daemon loaded `~/.fleet/config.json` and pushed config to the TUI over the Watch stream, so the indicator was rendered by the real end-to-end pipeline — not a unit test.

The indicator is the small "wifi" glyph **`·))`** that floats off the top-right of the `fleet` ASCII logo on the dashboard. Color was verified from the raw ANSI: **`38;5;196m` = bold red** (`errorStyle`, "desired but not yet connected") in every "shown" case — consistent with there being no real gateway to connect to.

### Detailed findings

**Bug fix verification (#199) — gRPC and webhook now light the indicator.**
- Seeded `config.json` with only `fleet_enabled: true` + a `gateway_url`. Launched the TUI → the logo's first line rendered ``  __ _         _    ·))`` with the glyph in red. Before this PR the indicator was gated solely on `RemoteMcpSettings.Enabled`, so a gRPC-only config showed nothing.
- Repeated with only `webhook_enabled: true` + a `gateway_url` → indicator shown (red). Same previously-broken path, now fixed.
- Combined `fleet_enabled` + `webhook_enabled` + `gateway_url` → shown.
- The pre-existing MCP path (`enabled: true` + `gateway_url`) still shows — no regression.

**Negative / guard cases — indicator stays dark when the tunnel would be a no-op.**
- No config file (all three surfaces off) → hidden.
- A surface enabled but **no** `gateway_url` → hidden (the tunnel never dials, so no misleading red "connecting" forever).
- A surface enabled with a **whitespace-only** `gateway_url` (`"   "`) → hidden (trimmed to empty).
- `gateway_url` set but every surface off → hidden.

**Live user flow (end-to-end, reactive).**
1. Started with `gateway_url` set and all surfaces off → dashboard indicator hidden.
2. Opened **Settings → Fleet MCP**, moved the cursor to **Enable Remote Fleet** and pressed space → row flipped to `[ on ]`.
3. Pressed `esc` back to the dashboard → indicator appeared live (red).
4. Re-entered Settings, toggled it back to `[ off ]`, returned → indicator disappeared again.

This confirms the live `SetConfig → daemon → Watch push → indicator` round-trip for the newly-covered gRPC surface, in both directions.

**Note on the green state:** the CONNECTED → green (`statusRunningStyle`, color 42) branch could not be exercised because it requires a real reachable fleet gateway, which isn't available in the QA sandbox. This PR does not change that branch — it only changes the predicate that decides whether the glyph renders at all (`RemoteMcpSettings.TunnelDesired()`), which is what was tested above. The red↔green selection logic is unchanged from `main`.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build & launch (real TUI + auto-spawned daemon) | ✅ Pass | Builds clean; fleet-list page renders, daemon connects (`Armada [ local ]`) |
| All-off baseline → no indicator | ✅ Pass | No `·))` glyph on the logo (regression guard) |
| Remote **MCP** only → indicator shows | ✅ Pass | Red `·))` appears (the one case that already worked) |
| Remote **gRPC** (`fleet_enabled`) only → indicator shows | ✅ Pass | Red `·))` now appears — **the fix** (was hidden before) |
| **Webhook** (`webhook_enabled`) only → indicator shows | ✅ Pass | Red `·))` now appears — **the fix** (was hidden before) |
| Enabled but **no gateway URL** → no indicator | ✅ Pass | Stays dark — no misleading "connecting forever" state |
| All three enabled + gateway → indicator shows | ✅ Pass | Red `·))`, no interaction/regression |

**Overall: ✅ PASS — 7 of 7 scenarios pass. The gRPC/webhook remote-indicator fix (#199) behaves exactly as specified.**

---

### Test method

All execution was isolated in a throwaway Docker container (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`) with the PR branch checked out and mounted. The **real** `fleet` TUI was built and driven inside the container's own tmux (never the host session); the TUI auto-spawned its `fleet` daemon, which read `~/.fleet/config.json`. Each scenario seeds a different `remote_mcp_settings` block, relaunches the TUI, and inspects the rendered VT frame stream for the wifi-style `·))` glyph that floats off the top of the fleet logo.

Color was force-enabled (`CLICOLOR_FORCE=1`) so lipgloss emits SGR codes in the headless pane. In every "shows" case the glyph rendered **bold red** (`\e[1;91m·))\e[0m` = `errorStyle`), i.e. the **desired-but-not-connected** state — correct, since no real gateway is reachable in the test container.

### Detailed findings

**1. All-off baseline (`config.json` absent → defaults).**
Logo renders as `  __ _  …  _` with nothing appended. `grep '·))'` → 0 frames. Indicator correctly hidden. ✅

**2. Remote MCP only** — `{"remote_mcp_settings":{"enabled":true,"gateway_url":"https://gateway.example.com"}}`.
Logo first line becomes `  __ _ … _    ·))`, glyph present across all painted frames, color bold-red. Confirms the full config → daemon → TUI pipeline and the previously-working case. ✅

**3. Remote gRPC only (THE FIX)** — `{"remote_mcp_settings":{"fleet_enabled":true,"gateway_url":"…"}}`.
Indicator present (red `·))`). Before this PR, `remoteIndicator` gated on `RemoteMcpSettings.Enabled` only, so this case rendered **no** indicator. Now it lights via `TunnelDesired()`. ✅

**4. Webhook only (THE FIX)** — `{"remote_mcp_settings":{"webhook_enabled":true,"gateway_url":"…"}}`.
Indicator present (red `·))`). Same previously-hidden case, now fixed. ✅

**5. Enabled but no gateway URL** — `{"remote_mcp_settings":{"fleet_enabled":true,"webhook_enabled":true,"gateway_url":""}}`.
TUI comes up and renders the fleet-list (`No instances` / `settings` visible, process still running) but the indicator stays **dark** (0 frames with `·))`). This is the exact controlled counterpart to #3/#4 — identical setup, only `gateway_url` differs — validating the `&& gatewayURL != ""` clause that prevents a permanent misleading "connecting" glyph for an unconfigured tunnel. ✅

**6. All three enabled + gateway.**
Indicator present (red `·))`); no rendering glitch from combining surfaces. ✅

### Note on the green (CONNECTED) state
The indicator's **green** branch requires a live, CONNECTED gateway tunnel, which isn't stood up in the headless QA container, so only the red/not-connected color was exercised here. That color branch is unchanged by this PR (the diff only changes the *presence* predicate from `Enabled` → `TunnelDesired()`), and the green path is covered by the added `TestRemoteIndicator` unit test. No concern.

PR #220, opened 2026-06-27
fix/pr-status-skip-default-branch: exclude the default branch from PR status indicator (#217)

### LINKED ISSUE #217: Fleet TUI - PR status indicator bug
The PR status indicator will incorrectly consider PRs involving the repos default branch. This was fine before the PR closed state was added last release. Now that the closed state indicator is there, if the default branch was ever PRed to any other branch then the RP status indicator will show closed, when it should not. 

# Desired State 
The "default" branch (often main) should not be considered for PR status indicator checks. Be sure to also do the "default" branch determination on any sub repos that may exist.

This change will prevent a "false" PR status indicator from displaying.

BenjaminBenetti:
## Summary

Fixes #217. The PR status auto-tag was showing a bogus purple "closed PR" indicator on instances merely parked on a repo's default branch (usually `main`).

The probe (`internal/server/pr_status.go`) lists PRs by `--head <current-branch>` for the workspace repo and every nested subrepo. When the checked-out branch is the repo's default branch, any PR that ever used that branch as its *HEAD* into another branch matches. This was harmless until PR #215 added the persistent purple closed/merged tag — now those long-closed PRs surface a false indicator on an instance just sitting on the trunk.

The fix skips a repo when its checked-out branch is that repo's own default branch:

- **Per-repo** determination, so nested subrepos are judged against their *own* default (as the ticket calls out).
- Prefers the local `origin/HEAD` ref (set by `git clone`, **no API round-trip**), falling back to `gh repo view --json defaultBranchRef` only when `origin/HEAD` isn't configured.
- On any failure the skip is a no-op — degrading to the prior behaviour rather than risk hiding a real PR.

The default branch is the trunk; it has no in-flight PR to track, so it's excluded entirely (open and closed).

A new `TestPRProbeScript_SkipsDefaultBranch` exercises the real probe script against a stubbed `gh` + real git repos, covering the on-default (skipped), on-feature (kept), and mixed per-repo subrepo cases.

### DIFF
--- a/internal/server/pr_status.go
+++ b/internal/server/pr_status.go
@@ -44,7 +44,9 @@ const (
 
 // prProbeScript runs inside the container with the workspace folder as its cwd
 // (the backend's ExecCommand resolves that). For the workspace repo and every
-// nested subrepo it finds the open PRs on the current branch (gh pr list) and
+// nested subrepo it first skips the repo when its checked-out branch is that
+// repo's own default branch (the trunk has no in-flight PR to track; issue #217),
+// then finds the open PRs on the current branch (gh pr list) and
 // then emits the FULL detail of each via `gh pr view <n> --json ...` — one JSON
 // object per PR. gh pr view is used (not gh pr list's --json) because gh pr
 // list's statusCheckRollup is capped at the first 100 checks, while gh pr view
@@ -69,6 +71,18 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
   [ -z "$br" ] && continue
   [ "$br" = "HEAD" ] && continue
   ( cd "$dir" || exit 0
+    # Skip the repo's DEFAULT branch (usually main): it is the trunk, not in-flight
+    # work, yet it may have been the HEAD of PRs into other branches over its life.
+    # Surfacing those long-closed PRs put a bogus purple "closed PR" tag on an
+    # instance merely parked on main (issue #217). Resolved PER-REPO so a nested
+    # subrepo is judged against its OWN default branch. Prefer the local origin/HEAD
+    # ref (set by git clone, no API round-trip); fall back to gh only when it isn't
+    # configured. On any failure "default" is empty and we DON'T skip — degrading to
+    # the prior (over-eager) behaviour rather than risk hiding a real PR.
+    default=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null)
+    default=${default#origin/}
+    [ -z "$default" ] && default=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name' 2>/dev/null)
+    [ -n "$default" ] && [ "$br" = "$default" ] && exit 0
     # One list call for the branch (gh's embedded --jq, so no standalone jq
     # dependency): "<STATE> <number>" per PR, e.g. "OPEN 12" / "MERGED 7". The
     # explicit --limit lifts gh's default 30-row cap so an open PR can't fall
--- a/internal/server/pr_status_test.go
+++ b/internal/server/pr_status_test.go
@@ -62,6 +62,18 @@ esac
 exit 0
 `
 
+// setSymbolicRef writes a symbolic ref in dir (e.g. refs/remotes/origin/HEAD ->
+// refs/remotes/origin/main, the marker `git clone` writes). git stores it as-is,
+// so no real remote or fetch is needed to emulate a freshly-cloned repo.
+func setSymbolicRef(t *testing.T, dir, name, target string) {
+	t.Helper()
+	c := exec.Command("git", "symbolic-ref", name, target)
+	c.Dir = dir
+	if out, err := c.CombinedOutput(); err != nil {
+		t.Fatalf("git symbolic-ref %s %s: %v\n%s", name, target, err, out)
+	}
+}
+
 func gitInitOnBranch(t *testing.T, dir, branch string) {
 	t.Helper()
 	if err := os.MkdirAll(dir, 0o755); err != nil {
@@ -247,6 +259,85 @@ func TestPRProbeScript_SkipsDefaultBranch(t *testing.T) {
 		if got.GetClosedCount() != 1 {
 			t.Errorf("ClosedCount = %d, want 1 (subrepo on its default branch skipped); out=%q",
 				got.GetClosedCount(), out)
+		}
+	})
+}
+
+// TestPRProbeScript_SkipsDefaultBranch_LocalOriginHead covers the production-
+// dominant arm of the issue-#217 skip: a cloned repo has refs/remotes/origin/HEAD
+// set, so the probe resolves the default branch from that LOCAL ref without a gh
+// round-trip. The gh stub here fails `repo view`, so the fallback can't supply a
+// default — any skip must come from the local ref alone. This guards the
+// `git symbolic-ref --short` output shape plus the `${default#origin/}` strip,
+// which the no-remote tests above never exercise (they fall through to gh).
+func TestPRProbeScript_SkipsDefaultBranch_LocalOriginHead(t *testing.T) {
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git not available")
+	}
+	// gh stub whose `repo view` FAILS: the default-branch fallback is unavailable,
+	// so a skip proves the local origin/HEAD ref drove it. `pr list` still reports a
+	// merged PR, so a repo that is NOT skipped surfaces a (wrong) purple tag.
+	const ghNoRepoView = `#!/bin/sh
+case "$1" in
+  auth) exit 0 ;;
+  repo) exit 1 ;;
+  pr)
+    case "$2" in
+      list) cat "$GH_FAKE_PRLIST" 2>/dev/null ; exit 0 ;;
+      view) printf '{"number":%s,"state":"MERGED","url":"https://example.test/%s","title":"x"}\n' "$3" "$3" ; exit 0 ;;
+    esac ;;
+esac
+exit 0
+`
+	bin := t.TempDir()
+	if err := os.WriteFile(filepath.Join(bin, "gh"), []byte(ghNoRepoView), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	closedList := filepath.Join(t.TempDir(), "closed")
+	if err := os.WriteFile(closedList, []byte("MERGED 210\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	// run builds a repo on `main` and runs the real probe; when withOriginHead is
+	// set it first writes origin/HEAD -> origin/main, emulating a fresh clone.
+	run := func(t *testing.T, withOriginHead bool) string {
+		t.Helper()
+		ws := t.TempDir()
+		gitInitOnBranch(t, ws, "main")
+		if withOriginHead {
+			setSymbolicRef(t, ws, "refs/remotes/origin/HEAD", "refs/remotes/origin/main")
+		}
+		cmd := exec.Command("sh", "-c", prProbeScript)
+		cmd.Dir = ws
+		cmd.Env = append(os.Environ(),
+			"PATH="+bin+string(os.PathListSeparator)+os.Getenv("PATH"),
+			"GH_FAKE_PRLIST="+closedList)
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			t.Fatalf("probe script exited non-zero: %v\noutput: %q", err, out)
+		}
+		return string(out)
+	}
+
+	t.Run("local origin/HEAD alone drives the skip", func(t *testing.T) {
+		out := run(t, true)
+		if got := parsePRProbeOutput(out); got != nil {
+			t.Errorf("default branch resolved via local origin/HEAD parsed to %+v, want nil — "+
+				"the local ref must drive the skip with no gh fallback; out=%q", got, out)
+		}
+	})
+
+	t.Run("negative control: no origin/HEAD and no gh fallback => not skipped", func(t *testing.T) {
+		// Same repo on main, but without the local ref: the lookup misses AND the gh
+		// fallback fails, so the default branch is NOT recognised and its closed PR
+		// surfaces. This proves the skip above came from the local ref, not elsewhere.
+		out := run(t, false)
+		got := parsePRProbeOutput(out)
+		if got == nil {
+			t.Fatalf("with no default-branch signal the repo should not be skipped, want a purple tag; out=%q", out)
+		}
+		if got.GetPrSignal() != fleetgrpc.PrSignal_PR_SIGNAL_PURPLE {
+			t.Errorf("got %+v, want PURPLE (repo not skipped)", got)
 		}
 	})
 }
--- a/internal/server/pr_status_test.go
+++ b/internal/server/pr_status_test.go
@@ -41,9 +41,14 @@ func TestRunProbeWithTimeout_KillsOnTimeout(t *testing.T) {
 // the open path requests statusCheckRollup, so the stub reports state OPEN; the
 // closed fallback requests only number,state,url,title, so it reports MERGED.
 // That lets a test prove which branch the script's awk classification took.
+// `gh repo view --json defaultBranchRef --jq ...` reports $GH_FAKE_DEFAULT_BRANCH
+// (default "main"), the default-branch determination's gh fallback — the test
+// repos have no origin remote, so the script's local origin/HEAD lookup misses and
+// this fallback is what the issue-#217 skip is exercised through.
 const stubGHScript = `#!/bin/sh
 case "$1" in
   auth) exit 0 ;;
+  repo) printf '%s\n' "${GH_FAKE_DEFAULT_BRANCH:-main}" ; exit 0 ;;
   pr)
     case "$2" in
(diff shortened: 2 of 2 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the default-branch skip for the PR-status probe (`internal/server/pr_status.go` + tests). The fix is correct and well-scoped: I verified that `git symbolic-ref --short refs/remotes/origin/HEAD` yields `origin/main` and `${default#origin/}` strips it to `main`; the per-repo `cd` makes both the git ref lookup and the `gh` fallback resolve against the right repo; `exit 0` correctly skips only the current repo's subshell (not the whole script); and an empty `default` degrades to the prior behaviour. `go vet`, the full `internal/server` suite, and the new `TestPRProbeScript_SkipsDefaultBranch` all pass locally.

A few **non-blocking** observations:

1. **The primary `origin/HEAD` path is untested.** The new test builds repos with `git init` (no `origin` remote), so the local `origin/HEAD` lookup always misses and only the `gh repo view` fallback is exercised. The production-dominant path — cloned repos *do* have `origin/HEAD` — and specifically the `--short` + `${default#origin/}` strip is the piece most likely to break silently, yet it has no coverage. Consider a case that does `git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/<default>` and asserts the skip happens (ideally with a `gh` stub that fails, proving the local ref alone drives it). See the inline note.

2. **An open PR whose head *is* the default branch is now hidden.** The skip excludes the default branch entirely (open *and* closed), as the description intends. Worth a conscious confirmation that this is acceptable for any flow where the trunk is itself the head of a live PR (fork→upstream, release-train branches). For standard trunk-based development this is clearly fine.

3. **Extra `gh repo view` round-trip when `origin/HEAD` is unset.** In the fallback case this runs once per such repo on every ~45s probe, inside the 20s-bounded pass. Rare for freshly-cloned repos but possible for some worktree/clone layouts; acceptable as-is, though caching the resolved default per repo path would eliminate it should it ever surface as latency/quota pressure.

Nice touches: matching the existing case-defensive shell style, and the per-repo subrepo test case is a good guard against a global (rather than per-repo) skip.

internal/server/pr_status.go
```diff
@@ -69,6 +71,18 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
...
   ( cd "$dir" || exit 0
+    # Skip the repo's DEFAULT branch (usually main): it is the trunk, not in-flight
+    # work, yet it may have been the HEAD of PRs into other branches over its life.
+    # Surfacing those long-closed PRs put a bogus purple "closed PR" tag on an
+    # instance merely parked on main (issue #217). Resolved PER-REPO so a nested
+    # subrepo is judged against its OWN default branch. Prefer the local origin/HEAD
+    # ref (set by git clone, no API round-trip); fall back to gh only when it isn't
+    # configured. On any failure "default" is empty and we DON'T skip — degrading to
+    # the prior (over-eager) behaviour rather than risk hiding a real PR.
+    default=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null)
```
BenjaminBenetti, line 82: 🔍 **CodeReview Agent**

This is the production-dominant path — freshly cloned repos have `origin/HEAD` set — but the new test never exercises it: its repos are created with `git init` (no `origin` remote), so this lookup always returns empty and only the `gh repo view` fallback on the next line runs. The `--short` output format plus the `${default#origin/}` strip is the most likely thing to break silently here.

Suggest adding a test case that sets the ref explicitly, e.g. `git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/<default>`, and asserts the repo is skipped — bonus points for pointing the `gh` stub at a failing script so the assertion proves the local ref alone drove the skip (no API round-trip).

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./...`) + `go vet` | ✅ Pass | Compiles clean with the change; vet clean |
| Parked on default branch (local `origin/HEAD`) | ✅ Pass | The bogus purple tag is gone; **zero** `gh repo view` API call (used local ref) |
| Feature branch surfaces closed PR | ✅ Pass | Purple tag still shown — no over-skip |
| Feature branch surfaces open PR | ✅ Pass | Live (open) tag still shown |
| Per-repo: subrepo judged on its **own** default | ✅ Pass | Exactly one repo emits in both directions; both-on-default ⇒ 0 |
| Non-`main` default branch honored | ✅ Pass | `develop`/`master` defaults skipped; on `main` *kept* when default is `master` |
| Detached HEAD | ✅ Pass | Guarded by existing `HEAD` check — no crash, no tag |
| `gh repo view` API fallback (no local `origin/HEAD`) | ✅ Pass | Skips on default via API; keeps on feature |
| Graceful degrade — default undeterminable | ✅ Pass | Does **not** skip; a real open/closed PR is never hidden |
| Sentinels: gh missing / not authenticated | ✅ Pass | `FLEET_NO_GH` / `FLEET_NO_AUTH` still emitted |
| Skipped-default exit code (issue #203 guard) | ✅ Pass | Script still exits 0; tag clears, not treated as a transient failure |

**Overall: ✅ PASS — 11 of 11 scenarios pass. Meets the issue #217 acceptance criteria.**

---

### Methodology
This PR changes the embedded `prProbeScript` shell that the server runs **inside each container** (`b.ExecCommandQuiet(workspaceDir, ["sh","-c",prProbeScript])`). I exercised the **verbatim** probe script (extracted from the PR head commit `08bc271`) the same way the server invokes it — `sh -c`, cwd = the workspace folder — entirely inside a throwaway Docker container (`golang:1.25`, git 2.47). Fixtures were **real git repos**; `gh` was a stub on `PATH` that mirrors the real `gh auth status` / `gh pr list` / `gh pr view` / `gh repo view` outputs and **logs every call** so I could verify the "no API round-trip" claim. The container was torn down afterward.

> Note: this is behavior-level QA of the actual code path, not a re-run of the package's Go unit tests.

### Detailed findings

**1. Core fix — parked on the default branch (issue #217).** Workspace repo on `main` with a *merged* PR in its history and `origin/HEAD → main` set locally. Probe output was **empty ⇒ no tag** (previously this surfaced a bogus purple "closed PR" indicator). The call log contained only `gh auth status` — **no `gh repo view`** — confirming both the preferred local-ref path *and* that the `origin/main → main` prefix strip (`${default#origin/}`) matched correctly (otherwise it would not have skipped).

**2. No over-skip on real work.** Same repo moved to `feature/login`: a merged PR correctly surfaces the **purple** tag, and an open PR surfaces the **live (OPEN)** tag. The skip only fires on the default branch.

**3. Per-repo judgement (explicit ask in #217).**
- top-level `feature` (kept) + nested subrepo on its own default `main` (skipped) ⇒ **1** PR object emitted.
- top-level `main` (skipped) + nested subrepo on `feature` (kept) ⇒ **1** PR object emitted — proving the default-branch skip is applied **per-repo, not globally**.
- both repos on their own default ⇒ **0** emitted (no tag).

**4. Non-`main` defaults & detached HEAD.** A repo whose default is `develop`, on `develop` ⇒ skipped. A repo whose default is `master`, currently on `main` ⇒ **kept** (purple) — confirms the default is each repo's *own* default, not a hardcoded "main". Detached HEAD (`branch == HEAD`) ⇒ skipped by the pre-existing guard, no crash.

**5. `gh repo view` API fallback.** With **no** local `origin/HEAD` configured, on `main`, the probe falls back to `gh repo view --json defaultBranchRef` (confirmed present in the call log) ⇒ skipped; on a feature branch ⇒ kept.

**6. Graceful degrade (safety property).** With no local `origin/HEAD` **and** `gh repo view` failing (default undeterminable), on `main`: the repo is **not** skipped — a merged PR still shows purple and, critically, a real **open** PR still surfaces as a live tag. The fix never hides an in-flight PR when it can't determine the default.

**7. Unchanged paths / regression sanity.** `gh` absent ⇒ `FLEET_NO_GH`; `gh` present but unauthenticated ⇒ `FLEET_NO_AUTH`; empty workspace ⇒ empty output, exit 0. A skipped default-branch repo being the *only/last* repo still leaves the whole script's exit status at **0** (the subshell `exit 0` only exits the subshell; the trailing `exit 0` governs the script) — so the issue #203 guard (a completed probe is authoritative, not a transient failure) is preserved.

No defects or regressions observed.

BenjaminBenetti:
Addressed the code review:

1. **Untested local `origin/HEAD` path** — fixed in 1bd2544. Added `TestPRProbeScript_SkipsDefaultBranch_LocalOriginHead`: a repo with `origin/HEAD` set and a `gh` stub whose `repo view` **fails**, so any skip can only be driven by the local ref + the `${default#origin/}` strip. A negative control (no `origin/HEAD` + failing `gh`) proves the same repo is *not* skipped without the local signal.
2. **Open PR whose head is the default branch is hidden** — intentional and matches #217 ("the default branch should not be considered for PR status indicator checks"). The trunk isn't tracked for either open or closed state; standard trunk-based development is unaffected.
3. **Extra `gh repo view` round-trip when `origin/HEAD` is unset** — acceptable as-is. The local-first resolution means the common (freshly-cloned) case pays no API cost; the fallback only runs for layouts missing `origin/HEAD`. Left a caching optimization out to keep the change minimal.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the default-branch skip added to the PR-status probe (`internal/server/pr_status.go`). This is a clean, well-scoped fix for #217 with genuinely thorough tests. No blocking issues found.

**Correctness — verified**
- I reproduced the shell behaviour locally: `git symbolic-ref --short refs/remotes/origin/HEAD` emits `origin/main`, and `${default#origin/}` correctly strips it to `main` (also handles slash-named defaults, e.g. `origin/release/next` → `release/next`, which still matches `git rev-parse --abbrev-ref HEAD`).
- A missing `origin/HEAD` returns empty (rc 128) → the `gh repo view` fallback engages as intended; a hard fallback failure leaves `default` empty → the repo is **not** skipped, preserving the prior (safe) behaviour. The fail-open design matches the PR description.
- Per-repo resolution inside the `( cd "$dir" … )` subshell is correct, and the early `exit 0` skips only that subshell, so sibling repos and the loop's final `exit 0` are unaffected. No quoting/injection concerns — `$br`/`$default` are quoted in the comparison.

**Non-blocking considerations**

1. **gh fallback cost for repos without `origin/HEAD`.** For any repo where the local `origin/HEAD` ref isn't set (subrepos added by means other than `git clone`, or clones where `set-head` never ran), every probe cycle now pays one extra `gh repo view` API call per such repo. Bounded and acceptable, but it's no longer "no round-trip" for those repos — worth keeping in mind if the probe runs on a tight interval across many subrepos.

2. **Stale `origin/HEAD` silently wins over the gh fallback.** The fallback only triggers when `origin/HEAD` is *unset*, not when it's *stale*. If a repo's remote default was renamed (e.g. `master` → `main`) but the local `origin/HEAD` still points at the old name, the stale local value is used and `gh` is never consulted — which could skip the wrong branch or fail to skip the real trunk. Very rare and self-heals via `git remote set-head origin -a`; just noting the precedence.

**Tests** — strong. Exercising the real `prProbeScript` against a stubbed `gh` + real git repos is the right call, and the two suites cover both arms well: the no-remote suite drives the `gh` fallback, while `…_LocalOriginHead` isolates the local-ref path (gh `repo view` deliberately fails) with a clean negative control. The per-repo subrepo case nicely proves the skip is scoped per repo, not global. Nice work on the comment density too — it matches the file's existing convention.

See the inline note for one behavioural trade-off on the skip line.

internal/server/pr_status.go
```diff
@@ -69,6 +71,18 @@ find . -maxdepth 5 -name node_modules -prune -o -name .git -prune -print 2>/dev/
...
+    default=${default#origin/}
+    [ -z "$default" ] && default=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name' 2>/dev/null)
+    [ -n "$default" ] && [ "$br" = "$default" ] && exit 0
```
BenjaminBenetti, line 85: 🔍 **CodeReview Agent**

Behavioural trade-off worth confirming is intended: this skips the repo's default branch **entirely**, including any genuinely *open* PR whose HEAD is the default branch. That happens in a few real workflows — a release PR `main → release`, or a fork's `main → upstream:main`. In those cases an instance parked on `main` does have an in-flight PR, and its open/green indicator now disappears too (not just the stale closed/purple one #217 is about).

This is almost certainly the right call for the dominant "parked on trunk" case, and is consistent with the issue. Flagging only so the open-PR-from-default scenario is a conscious accepted loss rather than an oversight.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| App builds (`go build ./...`) | ✅ Pass | Clean compile incl. changed `pr_status.go` |
| `fleet` binary launches (`--help`) | ✅ Pass | CLI/TUI starts, command tree renders |
| A — parked on default branch `main` (cloned, `origin/HEAD` set) | ✅ Pass | **Before:** purple tag (bug). **After:** no tag — issue #217 fixed |
| B — feature branch, only a closed/merged PR | ✅ Pass | Purple "closed PR" tag still surfaces (no over-skip) |
| C — feature branch with an open PR | ✅ Pass | Live status path unaffected; emits open PR detail |
| D — default branch, no local `origin/HEAD` | ✅ Pass | Skip falls back to `gh repo view`; repo skipped |
| E — per-repo: top-level feature + subrepo on `main` | ✅ Pass | Exactly 1 PR emitted; subrepo on its own default skipped |
| F — negative control: default unknown (`origin/HEAD` miss + `gh` fails) | ✅ Pass | Does **not** skip → degrades to prior behaviour, real PR not hidden |
| G — per-repo default is `master`, branch is `main` | ✅ Pass | `main` treated as a feature branch here → emits (per-repo, not hardcoded) |

**Overall: ✅ PASS — 9 of 9 scenarios passing**

---

## How this was tested

This PR changes `prProbeScript` in `internal/server/pr_status.go` — a shell script that runs **inside each fleet instance container** to compute the per-instance PR status auto-tag. Rather than run the unit tests, I QA'd the **actual shipped script** the way it runs in production: extracted it verbatim from the checked-out source and executed it under `sh` (dash) inside a throwaway Docker container against **real git repos** + a `gh` stub emulating the `auth`/`repo view`/`pr list`/`pr view` calls it makes. I also extracted the **old script from `main`** to demonstrate the bug genuinely existed and is now fixed.

I interpreted the script's stdout exactly as Go's `parsePRProbeOutput` does:
- empty stdout → **no auto-tag** (cleared)
- an `OPEN` PR object → **live** green/yellow/red tag
- only `MERGED`/`CLOSED` objects → persistent **purple** "closed PR" tag

Environment: `golang:1.25-bookworm` container, workspace mounted at `/work`, app built and launched inside the container (never on the host).

## Detailed findings

### Build & launch sanity
`go build ./...` compiled cleanly (exit 0), and `go build -o /tmp/fleet ./cmd/fleet` + `fleet --help` launched and printed the full command tree. The app builds and runs.

### Scenario A — the issue #217 bug, before/after (core fix)
Workspace: a single repo checked out on `main` with `refs/remotes/origin/HEAD → origin/main` (exactly what `git clone` leaves), and a merged PR #210 in `gh`'s history whose head was once `main`.

- **Before (old script from `main`):** emitted the merged PR JSON →
  `… "state":"MERGED" … ` → instance gets a **bogus purple "closed PR" tag** while merely parked on the trunk. This reproduces issue #217.
- **After (this PR):** stdout **empty** → **no auto-tag**. The default-branch repo is skipped. ✅

This is the production-dominant path (default resolved from the local `origin/HEAD` ref, **no `gh` round-trip**).

### Scenario B — regression guard (feature branch, closed PR)
On `feature` with only a merged PR → the persistent **purple** tag still appears. The fix does not over-skip closed-PR indicators on real work branches. ✅

### Scenario C — open PR unaffected
On `feature` with an open PR #216 → the script emits the full open-PR detail (`statusCheckRollup`, etc.) → **live** status tag. The normal path is untouched. ✅

### Scenario D — `gh` fallback for the default branch
On `main` with **no** local `origin/HEAD` → the script falls back to `gh repo view --json defaultBranchRef`, learns the default is `main`, and skips. stdout empty → no tag. ✅

### Scenario E — per-repo determination
Top-level on `feature` + a nested subrepo on `main` → **exactly one** PR object emitted (the top-level's); the subrepo on its own default branch is skipped. Proves the default-branch test is applied **per repo**, not globally. ✅

### Scenario F — graceful degrade (negative control)
On `main` with **no** `origin/HEAD` **and** `gh repo view` failing → the default branch can't be determined, so the script does **not** skip and the PR surfaces. This confirms the "on any failure, don't skip" guard: it degrades to the prior (over-eager) behaviour rather than risk **hiding a real PR**. ✅

### Scenario G — default branch ≠ `main`
Repo whose default is `master` but checked out on `main` (`origin/HEAD → origin/master`, `gh` reports `master`) → `main` is correctly treated as a feature branch here and the PR is emitted. The skip keys off each repo's actual default branch, not a hardcoded `"main"`. ✅

## Verdict
The fix resolves issue #217 (no bogus purple tag on the default branch) without regressing open-PR live status or the closed-PR purple tag on real branches, applies the determination per-repo, and degrades safely when the default branch can't be resolved. **QA passed.**

PR #222, opened 2026-06-28
feat(flog): log previously-untraced actions to fleet.log

BenjaminBenetti:
## Summary

Audited fleetd for **actions** that ran without leaving any semantic trail in `~/.fleet/fleet.log`, then added a dedicated `flog` record for each. Loops, pollers, probes and samplers are deliberately left silent (per the audit's no-flood guidance).

24 gaps closed across 20 files. Each is a discrete, event- or user-triggered action — state mutations, lifecycle transitions, external side-effects, copies, automation fires, cache wiring, and rejected inbound events.

## Records added

| Area | Action(s) | Record |
|------|-----------|--------|
| MCP | `fleet_exec`, `fleet_session_spawn`, `fleet_session_exec` | `exec` / `session spawned` / `session exec` (+ `… failed`) |
| Data plane | `CopyFile` (out), `CopyInto` (in: file + dir), `Logs` | `file copied out/in`, `logs opened` |
| Control socket | `browser.open`, `file.copy` dispatch | `browser open requested`, `file copy requested` |
| Lifecycle | start/stop backend failure (CLI in-process path) | `instance status change failed` |
| Automation | agent launch + event-file write (success) | `automation: agent launched` / `event file written` |
| Webhook | 400 non-json, 404 unknown-name, 413 oversized | `webhook: rejected` (reason=…) |
| Server | startup failures (mkdir/socket/listen/chmod) + serve crash | `fleet server start failed` / `crashed` |
| Misc | Watch connect/disconnect, PrepareBrowser, DeleteGroupLayout, SetLastSeenVersion | `watch subscribed/unsubscribed`, `browser prepared`, `group layout deleted`, `last seen version set` |
| Cache wiring | instance buildx / apt / image-cache configured | `instance … configured` |

## Notable structural changes

- **`ConfigureInstanceBuildx` / `ConfigureInstanceApt`** now return `(configured bool, err error)` (mirroring imagecache's existing pattern) so only *real* wiring is logged, not the silent docker/apt-absent skips. Caller signatures (`configureBuildkit`/`setupDebCache`) are unchanged; unit tests updated.
- **`WriteWarn`** now mirrors non-fatal warning banners into `fleet.log`; a new **`WriteWarnQuiet`** is used by the runtime poller's two reinstall call sites so a persistently-failing per-tick warning can't flood the log.

## Verification

`go build ./...`, `go vet`, `gofmt`, `golangci-lint` (0 issues), and tests for the changed packages (buildkit, debcache, create, instanceops, state, server, server/remote) all pass.

### DIFF
--- a/internal/server/copy.go
+++ b/internal/server/copy.go
@@ -3,7 +3,9 @@ package server
 import (
 	"archive/tar"
 	"bytes"
+	"context"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -44,6 +46,39 @@ import (
 // default message cap while keeping per-frame overhead negligible.
 const copyChunkSize = 64 * 1024
 
+// isClientGone reports whether err is an ordinary client disconnect / cancel
+// (TUI closed, Ctrl-C mid-copy) rather than a real transfer failure. Such errors
+// surface from stream.Send and from a cancelled request context; the repo logs
+// these quietly (cf. probeFailureIsAlarming in schedule.go) instead of as
+// alarming ERRORs.
+func isClientGone(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, context.Canceled) {
+		return true
+	}
+	switch status.Code(err) {
+	case codes.Canceled, codes.Unavailable:
+		return true
+	}
+	return false
+}
+
+// logCopyOutcome records a copy RPC result once on completion: success at info,
+// a benign client cancellation at info ("… canceled", not an alarming ERROR),
+// and a genuine failure at error. dir is "out" (CopyFile) or "in" (CopyInto).
+func logCopyOutcome(dir, fleetName, instance, p string, start time.Time, err error) {
+	switch {
+	case err == nil:
+		flog.Info("file copied "+dir, "fleet", fleetName, "instance", instance, "path", p, "ms", flog.MillisSince(start))
+	case isClientGone(err):
+		flog.Info("file copy "+dir+" canceled", "fleet", fleetName, "instance", instance, "path", p, "err", err)
+	default:
+		flog.Error("file copy "+dir+" failed", "fleet", fleetName, "instance", instance, "path", p, "err", err)
+	}
+}
+
 // CopyFile streams the file at req.path out of the instance: first a meta
 // chunk (name/mode/size), then data chunks until EOF. Only regular files are
 // supported. A relative path resolves against the backend exec working
@@ -55,13 +90,7 @@ func (s *service) CopyFile(req *fleetgrpc.CopyFileRequest, stream grpc.ServerStr
 		return err
 	}
 	filePath := req.GetPath()
-	defer func() {
-		if err != nil {
-			flog.Error("file copy out failed", "fleet", req.GetFleet(), "instance", req.GetInstance(), "path", filePath, "err", err)
-		} else {
-			flog.Info("file copied out", "fleet", req.GetFleet(), "instance", req.GetInstance(), "path", filePath, "ms", flog.MillisSince(start))
-		}
-	}()
+	defer func() { logCopyOutcome("out", req.GetFleet(), req.GetInstance(), filePath, start, err) }()
 	base := path.Base(filePath)
 	if filePath == "" || base == "/" || base == "." || base == ".." {
 		return status.Errorf(codes.InvalidArgument, "invalid file path %q", filePath)
@@ -294,11 +323,7 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 		// Directory copy-in returns here, before the single-file defer below, so
 		// log it on its own. The destination dir is the client-requested target.
 		err = s.copyDirInto(inst, open, stream)
-		if err != nil {
-			flog.Error("file copy in failed", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", open.GetDest(), "err", err)
-		} else {
-			flog.Info("file copied in", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", open.GetDest(), "ms", flog.MillisSince(start))
-		}
+		logCopyOutcome("in", open.GetFleet(), open.GetInstance(), open.GetDest(), start, err)
 		return err
 	}
 	// Single-file copy-in: register the outcome log now so validation, path
@@ -309,13 +334,7 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 	// so a successful record names the file actually written — the rename form
 	// (dest is a full file path) resolves to a single path, not dest/name.
 	destPath := path.Join(open.GetDest(), open.GetName())
-	defer func() {
-		if err != nil {
-			flog.Error("file copy in failed", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "err", err)
-		} else {
-			flog.Info("file copied in", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "ms", flog.MillisSince(start))
-		}
-	}()
+	defer func() { logCopyOutcome("in", open.GetFleet(), open.GetInstance(), destPath, start, err) }()
 	if err := validateCopyName(open.GetName()); err != nil {
 		return err
 	}
--- a/internal/server/copy.go
+++ b/internal/server/copy.go
@@ -11,10 +11,12 @@ import (
 	"path"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/BenjaminBenetti/fleet-man/fleetgrpc"
 	"github.com/BenjaminBenetti/fleet-man/internal/backendutil"
 	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
+	"github.com/BenjaminBenetti/fleet-man/internal/flog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -46,12 +48,20 @@ const copyChunkSize = 64 * 1024
 // chunk (name/mode/size), then data chunks until EOF. Only regular files are
 // supported. A relative path resolves against the backend exec working
 // directory (the workspace folder).
-func (s *service) CopyFile(req *fleetgrpc.CopyFileRequest, stream grpc.ServerStreamingServer[fleetgrpc.CopyFileChunk]) error {
+func (s *service) CopyFile(req *fleetgrpc.CopyFileRequest, stream grpc.ServerStreamingServer[fleetgrpc.CopyFileChunk]) (err error) {
+	start := time.Now()
 	inst, err := resolveServerInstance(req.GetFleet(), req.GetInstance())
 	if err != nil {
 		return err
 	}
 	filePath := req.GetPath()
+	defer func() {
+		if err != nil {
+			flog.Error("file copy out failed", "fleet", req.GetFleet(), "instance", req.GetInstance(), "path", filePath, "err", err)
+		} else {
+			flog.Info("file copied out", "fleet", req.GetFleet(), "instance", req.GetInstance(), "path", filePath, "ms", flog.MillisSince(start))
+		}
+	}()
 	base := path.Base(filePath)
 	if filePath == "" || base == "/" || base == "." || base == ".." {
 		return status.Errorf(codes.InvalidArgument, "invalid file path %q", filePath)
@@ -263,7 +273,8 @@ var copyIntoExecCommand = func(inst *fleet.Instance, argv []string) *exec.Cmd {
 // stream (the declared size is a hard cap) fails the RPC without ever clobbering
 // an existing destination, the same atomicity the download direction has. Only
 // single regular files; the mode is preserved.
-func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoChunk, fleetgrpc.CopyIntoReply]) error {
+func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoChunk, fleetgrpc.CopyIntoReply]) (err error) {
+	start := time.Now()
 	first, err := stream.Recv()
 	if err != nil {
 		return status.Errorf(codes.InvalidArgument, "copy into: no open header: %v", err)
@@ -280,7 +291,15 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 		return status.Errorf(codes.FailedPrecondition, "instance %s/%s is not running", open.GetFleet(), open.GetInstance())
 	}
 	if open.GetIsDir() {
-		return s.copyDirInto(inst, open, stream)
+		// Directory copy-in returns here, before the single-file defer below, so
+		// log it on its own. The destination dir is the client-requested target.
+		err = s.copyDirInto(inst, open, stream)
+		if err != nil {
+			flog.Error("file copy in failed", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", open.GetDest(), "err", err)
+		} else {
+			flog.Info("file copied in", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", open.GetDest(), "ms", flog.MillisSince(start))
+		}
+		return err
 	}
 	if err := validateCopyName(open.GetName()); err != nil {
 		return err
@@ -293,6 +312,14 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 	if err != nil {
 		return err
 	}
+	destPath := path.Join(finalDir, finalName)
+	defer func() {
+		if err != nil {
+			flog.Error("file copy in failed", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "err", err)
+		} else {
+			flog.Info("file copied in", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "ms", flog.MillisSince(start))
+		}
+	}()
 
 	mode := os.FileMode(open.GetMode()).Perm()
 	if mode == 0 {
--- a/internal/server/copy.go
+++ b/internal/server/copy.go
@@ -301,6 +301,18 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 		}
 		return err
 	}
+	// Single-file copy-in: register the outcome log now so validation, path
+	// resolution (incl. a rejected path-traversal name), and transfer failures
+	// are all traced — symmetric with the directory branch above. The logged
+	// destination is the client-requested target.
+	destPath := path.Join(open.GetDest(), open.GetName())
+	defer func() {
+		if err != nil {
+			flog.Error("file copy in failed", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "err", err)
+		} else {
+			flog.Info("file copied in", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "ms", flog.MillisSince(start))
+		}
+	}()
 	if err := validateCopyName(open.GetName()); err != nil {
 		return err
 	}
@@ -312,14 +324,6 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 	if err != nil {
 		return err
 	}
-	destPath := path.Join(finalDir, finalName)
-	defer func() {
-		if err != nil {
-			flog.Error("file copy in failed", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "err", err)
-		} else {
-			flog.Info("file copied in", "fleet", open.GetFleet(), "instance", open.GetInstance(), "path", destPath, "ms", flog.MillisSince(start))
-		}
-	}()
 
 	mode := os.FileMode(open.GetMode()).Perm()
 	if mode == 0 {
--- a/internal/server/copy.go
+++ b/internal/server/copy.go
@@ -303,8 +303,11 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 	}
 	// Single-file copy-in: register the outcome log now so validation, path
 	// resolution (incl. a rejected path-traversal name), and transfer failures
-	// are all traced — symmetric with the directory branch above. The logged
-	// destination is the client-requested target.
+	// are all traced — symmetric with the directory branch above. destPath starts
+	// as the client-requested target (the best identifier we have if we fail
+	// before resolving) and is replaced with the resolved write path once known,
+	// so a successful record names the file actually written — the rename form
+	// (dest is a full file path) resolves to a single path, not dest/name.
 	destPath := path.Join(open.GetDest(), open.GetName())
 	defer func() {
 		if err != nil {
@@ -324,6 +327,7 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 	if err != nil {
 		return err
 	}
+	destPath = path.Join(finalDir, finalName)
 
 	mode := os.FileMode(open.GetMode()).Perm()
 	if mode == 0 {
--- a/internal/server/mcp_tools.go
+++ b/internal/server/mcp_tools.go
@@ -15,6 +15,7 @@ import (
 	"github.com/BenjaminBenetti/fleet-man/internal/backendutil"
 	"github.com/BenjaminBenetti/fleet-man/internal/dotfiles"
 	"github.com/BenjaminBenetti/fleet-man/internal/fleet"
+	"github.com/BenjaminBenetti/fleet-man/internal/flog"
 	"github.com/BenjaminBenetti/fleet-man/internal/tui"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"google.golang.org/grpc/status"
@@ -628,6 +629,7 @@ func (s *service) mcpExec(ctx context.Context, _ *mcp.CallToolRequest, in FleetE
 	cmd := backendutil.NewForInstance(inst, false).ExecCommand(inst.WorkspaceDir, in.Command).Cmd
 	cmd.Stdout = &outBuf
 	cmd.Stderr = &errBuf
+	began := time.Now()
 	runErr := runCmd(cctx, cmd)
 	exitCode := 0
 	if runErr != nil {
@@ -637,9 +639,11 @@ func (s *service) mcpExec(ctx context.Context, _ *mcp.CallToolRequest, in FleetE
 		if errors.As(runErr, &ee) {
 			exitCode = ee.ExitCode()
 		} else {
+			flog.Error("exec", "fleet", in.Fleet, "instance", in.Instance, "cmd", strings.Join(in.Command, " "), "ms", flog.MillisSince(began), "err", runErr)
 			return nil, FleetExecOutput{}, fmt.Errorf("exec: %w", runErr)
 		}
 	}
+	flog.Info("exec", "fleet", in.Fleet, "instance", in.Instance, "cmd", strings.Join(in.Command, " "), "code", exitCode, "ms", flog.MillisSince(began))
 	return nil, FleetExecOutput{Stdout: outBuf.String(), Stderr: errBuf.String(), ExitCode: exitCode}, nil
 }
 
@@ -670,8 +674,10 @@ func (s *service) mcpSessionSpawn(ctx context.Context, _ *mcp.CallToolRequest, i
 	defer cancel()
 	snippet := dotfiles.TmuxEnsureInstalled + fmt.Sprintf(`tmux new-session -d -s %s`, dotfiles.ShQuote(sessionName))
 	if out, err := runContainerShell(cctx, inst, snippet); err != nil {
+		flog.Error("session spawn failed", "fleet", in.Fleet, "instance", in.Instance, "session", sessionName, "err", err)
 		return nil, FleetSessionMessageOutput{}, fmt.Errorf("spawn session %q: %w: %s", in.Session, err, strings.TrimSpace(out))
 	}
+	flog.Info("session spawned", "fleet", in.Fleet, "instance", in.Instance, "session", sessionName)
 	return nil, FleetSessionMessageOutput{Message: fmt.Sprintf("created tmux session %q in %s/%s", sessionName, in.Fleet, in.Instance)}, nil
 }
 
@@ -697,8 +703,10 @@ func (s *service) mcpSessionExec(ctx context.Context, _ *mcp.CallToolRequest, in
 	defer cancel()
 	snippet := fmt.Sprintf(`tmux send-keys -t %s %s Enter`, dotfiles.ShQuote(sessionName), dotfiles.ShQuote(in.Command))
 	if out, err := runContainerShell(cctx, inst, snippet); err != nil {
+		flog.Error("session exec failed", "fleet", in.Fleet, "instance", in.Instance, "session", sessionName, "err", err)
 		return nil, FleetSessionMessageOutput{}, fmt.Errorf("exec in session %q: %w: %s", in.Session, err, strings.TrimSpace(out))
 	}
+	flog.Info("session exec", "fleet", in.Fleet, "instance", in.Instance, "session", sessionName, "cmd", in.Command)
 	return nil, FleetSessionMessageOutput{Message: fmt.Sprintf("sent command to session %q; read the session to see output", sessionName)}, nil
 }
 
--- a/internal/server/webhook.go
+++ b/internal/server/webhook.go
@@ -70,7 +70,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	if !nameFound {
 		// No trigger anywhere carries this name. Same 404 the gateway gives an
 		// unknown id, so a probe can't enumerate configured webhook names.
-		flog.Warn("webhook: rejected", "reason", "unknown-name", "name", name, "status", 404)
+		flog.Warn("webhook: rejected", "reason", "unknown-name", "name", clipName(name), "status", 404)
 		http.Error(w, "no webhook trigger with that name", http.StatusNotFound)
 		return
 	}
@@ -99,7 +99,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	// treats as json-path is a misconfiguration worth surfacing loudly, not
 	// silently half-firing whichever same-named regex triggers happen to match.
 	if jsonPathActive && !bodyIsJSON(body) {
-		flog.Warn("webhook: rejected", "reason", "non-json", "name", name, "status", 400)
+		flog.Warn("webhook: rejected", "reason", "non-json", "name", clipName(name), "status", 400)
 		http.Error(w, "json-path webhook filter requires a JSON body; set the webhook content type to application/json", http.StatusBadRequest)
 		return
 	}
@@ -191,6 +191,18 @@ func (s *service) enqueueWebhookFires(ctx context.Context, fires []triggerFire) 
 	case <-timer.C:
 		return false
 	}
+}
+
+// clipName bounds an externally-supplied webhook name before it is written to
+// the append-only event log, so a prober hitting this unauthenticated endpoint
+// with unknown / oversized names can't emit arbitrarily long log lines. slog
+// already escapes the value, so only the length needs bounding.
+func clipName(name string) string {
+	const max = 128
+	if len(name) > max {
+		return name[:max] + "...(clipped)"
+	}
+	return name
 }
 
 // firstPathSegment returns the first path segment (the webhook name), trimming
--- a/internal/server/webhook.go
+++ b/internal/server/webhook.go
@@ -48,13 +48,14 @@ func (s *service) webhookHandler() http.Handler {
 func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	name := firstPathSegment(r.URL.Path)
 	if name == "" {
+		flog.Warn("webhook: rejected", "reason", "empty-name", "status", 400)
 		http.Error(w, "webhook name required", http.StatusBadRequest)
 		return
 	}
 
 	body, err := io.ReadAll(http.MaxBytesReader(w, r.Body, maxWebhookBodySize))
 	if err != nil {
-		flog.Warn("webhook: rejected", "reason", "body-too-large", "status", 413)
+		flog.Warn("webhook: rejected", "reason", "body-too-large", "name", clipName(name), "status", 413)
 		http.Error(w, "request body too large or unreadable", http.StatusRequestEntityTooLarge)
 		return
 	}
@@ -198,11 +199,16 @@ func (s *service) enqueueWebhookFires(ctx context.Context, fires []triggerFire) 
 // with unknown / oversized names can't emit arbitrarily long log lines. slog
 // already escapes the value, so only the length needs bounding.
 func clipName(name string) string {
-	const max = 128
-	if len(name) > max {
-		return name[:max] + "...(clipped)"
+	const maxRunes = 128
+	if len(name) <= maxRunes { // bytes <= maxRunes implies runes <= maxRunes
+		return name
 	}
-	return name
+	runes := []rune(name)
+	if len(runes) <= maxRunes {
+		return name
+	}
+	// Cut on a rune boundary so a multi-byte rune is never split.
+	return string(runes[:maxRunes]) + "...(clipped)"
 }
 
 // firstPathSegment returns the first path segment (the webhook name), trimming
--- a/internal/server/webhook.go
+++ b/internal/server/webhook.go
@@ -54,6 +54,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 
 	body, err := io.ReadAll(http.MaxBytesReader(w, r.Body, maxWebhookBodySize))
 	if err != nil {
+		flog.Warn("webhook: rejected", "reason", "body-too-large", "status", 413)
 		http.Error(w, "request body too large or unreadable", http.StatusRequestEntityTooLarge)
 		return
 	}
@@ -69,6 +70,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	if !nameFound {
 		// No trigger anywhere carries this name. Same 404 the gateway gives an
 		// unknown id, so a probe can't enumerate configured webhook names.
+		flog.Warn("webhook: rejected", "reason", "unknown-name", "name", name, "status", 404)
 		http.Error(w, "no webhook trigger with that name", http.StatusNotFound)
 		return
 	}
@@ -97,6 +99,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	// treats as json-path is a misconfiguration worth surfacing loudly, not
 	// silently half-firing whichever same-named regex triggers happen to match.
 	if jsonPathActive && !bodyIsJSON(body) {
+		flog.Warn("webhook: rejected", "reason", "non-json", "name", name, "status", 400)
 		http.Error(w, "json-path webhook filter requires a JSON body; set the webhook content type to application/json", http.StatusBadRequest)
 		return
 	}
--- a/internal/server/webhook.go
+++ b/internal/server/webhook.go
@@ -55,7 +55,9 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 
 	body, err := io.ReadAll(http.MaxBytesReader(w, r.Body, maxWebhookBodySize))
 	if err != nil {
-		flog.Warn("webhook: rejected", "reason", "body-too-large", "name", clipName(name), "status", 413)
+		// Not logged: this fires on the unauthenticated path for ANY name, so a
+		// prober could otherwise grow fleet.log one line per request. The sender
+		// still gets the 413 (surfaced in its delivery dashboard).
 		http.Error(w, "request body too large or unreadable", http.StatusRequestEntityTooLarge)
 		return
 	}
@@ -70,8 +72,9 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	fires, nameFound, jsonPathActive := collectWebhookFires(st, name, body)
 	if !nameFound {
 		// No trigger anywhere carries this name. Same 404 the gateway gives an
-		// unknown id, so a probe can't enumerate configured webhook names.
-		flog.Warn("webhook: rejected", "reason", "unknown-name", "name", clipName(name), "status", 404)
+		// unknown id, so a probe can't enumerate configured webhook names. Not
+		// logged: an unknown name is attacker-controllable probe noise on this
+		// unauthenticated path, so recording it would let a prober grow fleet.log.
 		http.Error(w, "no webhook trigger with that name", http.StatusNotFound)
 		return
 	}
--- a/internal/buildkit/instance.go
+++ b/internal/buildkit/instance.go
@@ -24,30 +24,36 @@ const (
 
 // ConfigureInstanceBuildx points an instance's docker buildx at the fleet's
 // shared buildkit server. It first probes for BOTH docker and the buildx plugin
-// inside the instance; if either is missing it returns nil WITHOUT touching the
-// instance — the documented "do nothing, no error" behaviour for images that
-// lack docker tooling. When both are present it (re)creates a "remote" builder
-// bound to the bind-mounted socket and selects it as the default.
+// inside the instance; if either is missing it returns (false, nil) WITHOUT
+// touching the instance — the documented "do nothing, no error" behaviour for
+// images that lack docker tooling. When both are present it (re)creates a
+// "remote" builder bound to the bind-mounted socket and selects it as the
+// default.
+//
+// The returned bool reports whether buildx was actually wired (both tools
+// present): a probe error yields (false, err); the silent-skip branch yields
+// (false, nil); a configure-script failure yields (true, err); success yields
+// (true, nil).
 //
 // The configure step is idempotent (remove-then-create), so it is safe to call
 // on every create, clone, and start. Callers should treat a returned error as
 // non-fatal (warn and continue) — a build-cache wiring failure must not block
 // an otherwise-usable instance.
-func ConfigureInstanceBuildx(b instanceExecer, workspaceDir string) error {
+func ConfigureInstanceBuildx(b instanceExecer, workspaceDir string) (configured bool, err error) {
 	out, err := b.ExecCommand(workspaceDir, []string{"sh", "-c", probeScript()}).CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("probe docker/buildx: %w (%s)", err, strings.TrimSpace(string(out)))
+		return false, fmt.Errorf("probe docker/buildx: %w (%s)", err, strings.TrimSpace(string(out)))
 	}
 	if !buildxPresent(string(out)) {
 		// docker or buildx absent — silently skip per the feature contract.
-		return nil
+		return false, nil
 	}
 
 	out, err = b.ExecCommand(workspaceDir, []string{"sh", "-c", configureScript()}).CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("configure buildx: %w (%s)", err, strings.TrimSpace(string(out)))
+		return true, fmt.Errorf("configure buildx: %w (%s)", err, strings.TrimSpace(string(out)))
 	}
-	return nil
+	return true, nil
 }
 
 // probeScript reports whether docker AND the buildx plugin are usable inside the
--- a/internal/buildkit/instance_test.go
+++ b/internal/buildkit/instance_test.go
@@ -66,8 +66,12 @@ func TestProbeAndConfigureScripts(t *testing.T) {
 
 func TestConfigureInstanceBuildxSkipsWithoutDocker(t *testing.T) {
 	fe := &fakeExecer{probeOut: probeMarkerAbsent}
-	if err := ConfigureInstanceBuildx(fe, "/ws"); err != nil {
+	configured, err := ConfigureInstanceBuildx(fe, "/ws")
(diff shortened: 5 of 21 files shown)


internal/server/mcp_tools.go
```diff
@@ -637,9 +639,11 @@ func (s *service) mcpExec(ctx context.Context, _ *mcp.CallToolRequest, in FleetE
 		if errors.As(runErr, &ee) {
 			exitCode = ee.ExitCode()
 		} else {
+			flog.Error("exec", "fleet", in.Fleet, "instance", in.Instance, "cmd", strings.Join(in.Command, " "), "ms", flog.MillisSince(began), "err", runErr)
 			return nil, FleetExecOutput{}, fmt.Errorf("exec: %w", runErr)
 		}
 	}
+	flog.Info("exec", "fleet", in.Fleet, "instance", in.Instance, "cmd", strings.Join(in.Command, " "), "code", exitCode, "ms", flog.MillisSince(began))
```
BenjaminBenetti, line 646: 🔍 **CodeReview Agent**

**Secrets-in-logs surface.** Both the success (`flog.Info("exec", … "cmd", …)`) and error paths log the *full* command line, and `session exec` (line ~709) logs `in.Command` verbatim. These are agent-driven tool calls whose argv can carry tokens/credentials (e.g. `curl -H "Authorization: Bearer …"`, `export TOKEN=…`).

`fleet.log` is opened `O_APPEND` with no rotation or size cap (`flog.initLogger`), so anything written here is retained in plaintext indefinitely. These actions were previously untraced, so this is a genuine behavior change. Consider whether to log the command at all, truncate/redact it, or gate full-command logging behind a debug level — at minimum it's worth a deliberate decision rather than a default.

internal/server/webhook.go
```diff
@@ -69,6 +70,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	if !nameFound {
 		// No trigger anywhere carries this name. Same 404 the gateway gives an
 		// unknown id, so a probe can't enumerate configured webhook names.
+		flog.Warn("webhook: rejected", "reason", "unknown-name", "name", name, "status", 404)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

**This rejection log is externally drivable.** Per the file header, this endpoint has no auth — the capability is the public session URL — and the `name` here is fully attacker-controlled. Combined with `fleet.log` being append-only with no rotation (`flog.initLogger`), anyone holding the URL can drive unbounded `Warn` lines (including arbitrary `name` values) by probing unknown names / oversized bodies / non-JSON.

This is partly pre-existing (the success path at line ~110/127 already logs every delivery), and it's capability-gated, so severity is low. But it's notable that the same PR adds `WriteWarnQuiet` specifically to stop *internal* per-tick flooding while adding the most *externally* floodable log in the codebase. Worth confirming this is intended; if so, consider sampling/rate-limiting the reject path or dropping the unbounded `name` attribute.

internal/server/copy.go
```diff
@@ -293,6 +312,14 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
 	if err != nil {
 		return err
 	}
+	destPath := path.Join(finalDir, finalName)
+	defer func() {
```
BenjaminBenetti: 🔍 **CodeReview Agent**

**Single-file copy-in validation failures aren't traced.** This `defer` is registered *after* `validateCopyName` (304), the negative-size check (307), and `resolveCopyIntoDest` (311). A failure in any of those returns before the defer is in scope, so no `file copy in failed` line is emitted — whereas the directory branch (296–301) logs all of its errors, and every error *after* this point is logged.

A rejected name (the path-traversal / bad-name guard) is arguably exactly the kind of inbound event you'd want in the log. If the omission is intentional (only log "real" copy attempts), fine — but the asymmetry between the dir and single-file paths is worth a second look.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Review of #222 — `feat(flog): log previously-untraced actions to fleet.log`**

Solid, well-scoped change. The audit-then-instrument approach is consistent, the `(configured bool, err error)` refactor cleanly mirrors imagecache so only *real* wiring is logged, and the `WriteWarn` / `WriteWarnQuiet` split is the right call. I confirmed:

- `time` is already imported in `mcp_tools.go`; `start`/`result` exist in `lifecycle.go` — no compile breaks.
- `WriteWarnQuiet` covers **both** per-tick `WriteWarn` sites in `statsActivityPass`; the remaining `WriteWarn` callers are all one-shot create/transition paths, so the no-flood guarantee holds.
- The named-return `(err error)` + `defer` pattern in `CopyFile`/`CopyInto` resolves correctly (shadowing `if err :=` blocks still propagate via `return`).
- The `(true, err)` configure-failure case is never mislogged — callers return on `err` before the `if configured` log.

A few non-blocking observations (details inline):

1. **Secrets in logs** — `exec` / `session exec` persist full command lines to `~/.fleet/fleet.log`, which is append-only with **no rotation/size cap** (`flog.initLogger`). Worth a conscious decision. *(inline on `mcp_tools.go`)*
2. **Webhook rejection logging is externally drivable** — the 404 path logs an attacker-controlled `name`; given no log rotation and the PR's own no-flood discipline, this is the one unbounded external write the change adds. *(inline on `webhook.go`)*
3. **`CopyInto` single-file validation failures bypass the failure log** — the `defer` sits below `validateCopyName`/size/`resolveCopyIntoDest`, so a rejected name isn't traced, unlike the dir branch. *(inline on `copy.go`)*

Minor nits (no change required):

- **New attribute keys aren't reflected in the convention list.** `flog.go` documents a canonical key set and asks callers to "keep keys consistent … so the log stays greppable". This PR adds `cmd`, `path`, `url`, `bytes`, `follow`, `trigger`, `group`, `version`, `runtime`, `reason`, `name`, `status`, `step`, `code`, `workspace` — none added to that list. Nice reuse of `from`/`to`/`ms`/`err` though. Consider extending the documented set.
- `schedule.go` calls `w.event.payload()` twice (the write arg and `"bytes", len(...)`); negligible, but a local `pl := w.event.payload()` would avoid the re-format.
- `Logs` logs `logs opened` with no symmetric close, whereas `Watch` logs both subscribe/unsubscribe — minor asymmetry, fine to leave.

Nothing here blocks merge.

BenjaminBenetti:
Thanks for the review — addressed the actionable points in deeff3c:

1. **copy.go** — `CopyInto`'s outcome `defer` now registers *before* `validateCopyName` / size / `resolveCopyIntoDest`, so single-file validation failures (including a rejected path-traversal name) are traced, symmetric with the directory branch.
2. **webhook.go** — the externally-supplied `name` is now `clipName`-bounded (128 chars) before the 404/400 rejection logs, so a prober can't emit arbitrarily long lines. (Note: `fleet.log` *does* get daily rotation w/ 100-day retention via the logrotate path, so total volume is bounded — but clipping the line size is still the right hardening.)
3. **schedule.go** — `payload()` computed once.
4. **flog.go** — documented the new attribute keys (`cmd`, `path`, `url`, `bytes`, `follow`, `trigger`, `group`, `version`, `runtime`, `reason`, `name`, `status`, `step`, `code`, `workspace`, `agent`).

On **secrets-in-`exec`-logs**: this is intentional — logging the command was an explicit ask so agent-driven container execution is traceable. Left as-is, but flagging it as the conscious decision you called out; happy to redact/omit the `cmd` attribute (or gate it behind a setting) if the secrets-in-argv exposure outweighs the audit value.

The `logs opened` (no symmetric close) asymmetry I left as-is — a `--follow` close is lower-signal than the open.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `go build ./...` + `go vet` | ✅ Pass | Signature changes (`ConfigureInstanceBuildx`/`ConfigureInstanceApt` → `(bool, error)`), `WriteWarn`/`WriteWarnQuiet`/`writeWarnFile` split all compile; 0 vet issues; `gofmt -l` clean on changed files |
| Server startup → `fleet.log` plumbing | ✅ Pass | Daemon auto-spawns on first client; structured records land in `~/.fleet/fleet.log` |
| Instance create (devcontainer) | ✅ Pass | `qa-fixture/qa1` came up in ~15s; created/logged correctly |
| MCP `fleet_exec` → `exec` record | ✅ Pass | `msg=exec … cmd="echo qa-mcp-exec-marker" code=0 ms=222` |
| MCP `fleet_session_spawn` → `session spawned` | ✅ Pass | `session=qa1~qasess` |
| MCP `fleet_session_exec` → `session exec` | ✅ Pass | `session=qa1~qasess cmd="echo qa-session-exec-marker"` |
| `fleet copy` out → `file copied out` | ✅ Pass | `path=/etc/hostname ms=401`; copied content correct |
| `fleet copy` in → `file copied in` | ✅ Pass | `path=/tmp/upload.txt`; file verified present in container |
| `fleet logs` → `logs opened` | ✅ Pass | `follow=false` |
| TUI Watch → `watch subscribed`/`unsubscribed` | ✅ Pass | `runtime=true` on both; TUI rendered fleet/instance + exited cleanly on `q` |
| Forced backend failure → `instance status change failed` | ✅ Pass | `from=running to=stopped err="exit status 1"`; logged from CLI in-process path (`pid` ≠ daemon pid) |
| stop / start success (regression) | ✅ Pass | State transitions intact; postStart hook ran |
| Cache-`configured` records on a non-addon fleet | ✅ Pass | Correctly **silent** — validates the new `(configured bool, …)` return only logs real wiring |
| Webhook / automation / control-socket / server-start-failure records | ⏭️ Skipped | Need a gateway tunnel / trigger+agent fixture / fault injection not set up in this env; each is a simple `flog` call on an existing branch |

**Overall: ✅ PASS — 13 of 13 exercised scenarios behaved as expected; 1 group of edge-path records skipped (environment, not a defect).**

---

### Environment
Isolated **privileged Docker container** (`qa-222`, image `mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`) with the PR branch checked out, the shared docker daemon mounted, and a matched `HOME` path so devcontainer bind-mounts resolve. All app execution / TUI / tmux ran **inside** that container. Built `fleet` from `af57d7e`, brought up a real devcontainer instance against the repo's own `integration/fixture` (`mcr.microsoft.com/devcontainers/base:debian-12`), and drove the live daemon.

### Detailed findings

**1. Build / vet / fmt** — `go build ./...` and `go vet ./...` both exit 0; `gofmt -l` on the 20 changed `.go` files is empty. The structural changes (the two `Configure*` functions now returning `configured bool`, and the `WriteWarn` → `writeWarnFile`/`WriteWarn`/`WriteWarnQuiet` split) compile with all callers updated.

**2. Headline feature — records actually land in `fleet.log`.** Every action below was driven through the real running app, then confirmed by grepping `~/.fleet/fleet.log`:

```
msg="file copied out"           fleet=qa-fixture instance=qa1 path=/etc/hostname ms=401
msg="file copied in"            fleet=qa-fixture instance=qa1 path=/tmp/upload.txt ms=884
msg="logs opened"               fleet=qa-fixture instance=qa1 follow=false
msg=exec                        fleet=qa-fixture instance=qa1 cmd="echo qa-mcp-exec-marker" code=0 ms=222
msg="session spawned"           fleet=qa-fixture instance=qa1 session=qa1~qasess
msg="session exec"              fleet=qa-fixture instance=qa1 session=qa1~qasess cmd="echo qa-session-exec-marker"
msg="watch subscribed"          runtime=true
msg="watch unsubscribed"        runtime=true
msg="instance status change failed" fleet=qa-fixture instance=qa1 from=running to=stopped err="exit status 1" ms=9
```

- The MCP records (`exec`, `session spawned`, `session exec`) were exercised via a real MCP client connecting to the daemon's Streamable-HTTP endpoint with the bearer token from `~/.fleet/mcp.env` — note the CLI `exec`/`spawn-session`/`exec-in-session` commands run client-side and intentionally do **not** hit these handlers, so the MCP transport is the correct surface for them.
- Keys are consistent and greppable (`fleet`/`instance`/`session`/`path`/`ms`/`err`), matching the conventions documented in `internal/flog`.
- The `instance status change failed` record was produced by force-removing the container out from under fleet and then issuing `fleet stop`; it logged from a **different pid** than the daemon, confirming the CLI in-process lifecycle path the PR specifically targets.

**3. No regressions observed.** TUI launched, rendered the fleet/instance with live stats, and quit cleanly (no leaked tmux session). `fleet stop`/`fleet start` success paths work and preserve state. The only `level=ERROR` line in the whole log is the one fault I injected deliberately — no panics, no stray errors during create/exec/copy/logs/watch.

**4. Positive check on the structural change.** During a normal create against a fleet without buildx/apt/image-cache addons enabled, **no** `instance … configured` records were emitted — exactly the intended effect of switching `ConfigureInstanceBuildx`/`ConfigureInstanceApt` to return `(configured bool, err error)` so only real wiring is logged rather than the silent docker/apt-absent skips.

### Not exercised (out of scope for this environment)
`webhook: rejected` (reachable only through a gateway tunnel), `automation: agent launched` / `event file written` (needs a configured trigger + agent fixture), the control-socket `browser open requested` / `file copy requested` / `browser prepared` flow, and `fleet server start failed`/`crashed` (needs failure injection). These are each a single `flog` call added onto a pre-existing branch and carry low regression risk, but they were **not** verified at runtime here.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the flog-coverage changes (head `deeff3c`). This is a clean, well-scoped, and unusually well-documented change. No correctness bugs or build-breakers found — a few minor completeness/consistency notes below.

### What I verified (no action needed)
- **Signature change is fully propagated.** `ConfigureInstanceBuildx` / `ConfigureInstanceApt` → `(configured bool, err error)` has exactly two production callers (`configureBuildkit`, `setupDebCache`), both updated, plus the unit tests. No missed callers — `go build` won't break. Mirroring imagecache's existing `func(bool, error)` result pattern is a nice consistency win.
- **Named-return + `defer` logging in `copy.go` is sound.** In `CopyInto`/`CopyFile` the `if err := …; err != nil { return err }` shadows are still captured by the deferred logger because `return err` assigns the named result before defers run; the final `return stream.SendAndClose(...)` also sets it. No false "copied in/out" success records on a failed transfer. (One benign edge: a late `SendAndClose` failure after the atomic rename logs `file copy in failed` even though the bytes landed — acceptable, the RPC genuinely didn't confirm.)
- **No log flooding.** `browser open requested` / `file copy requested` fire inside the per-envelope `control.Listen` handler (once per inbound event), not on each `syncRunning` poll tick. `Watch` subscribe/unsubscribe is per long-lived TUI stream, not per poll.
- **Multi-line warnings stay one line.** `WriteWarn` now mirrors to `fleet.log` via slog's `TextHandler`, which quotes/escapes values — so the multi-line blobs from `run_startup_scripts.go` / `dotfiles_install.go` are `\n`-escaped onto a single record rather than tearing the log. The `WriteWarn`/`WriteWarnQuiet` split correctly keeps the per-tick reinstall warnings out of the log.

### Minor notes
1. **Webhook rejection coverage is missing one path.** The PR traces the 413 / 404 / 400-non-json rejections, but the earliest rejection — the empty-name `400` at `internal/server/webhook.go:50` (`if name == "" { http.Error(... "webhook name required" ...) }`) — returns without a `webhook: rejected` record. For completeness against the stated goal ("rejected inbound events"), consider an `flog.Warn("webhook: rejected", "reason", "empty-name", "status", 400)` there too.
2. **`body-too-large` log omits the `name` key.** The 404 and 400 variants log `"name", clipName(name)`, but the 413 at `webhook.go:57` doesn't. `name` is already computed at that point; adding `"name", clipName(name)` would make the three rejection records uniformly greppable. (Inline comment below.)
3. **`clipName` truncates on a byte boundary.** `name[:max]` (`webhook.go`) can split a multi-byte UTF-8 rune since the URL path is already percent-decoded. The `TextHandler` escapes the result so there's no tearing — purely cosmetic — but truncating on a rune boundary would be tidier.
4. **Data-at-rest consideration (not a blocker).** `exec` and `session exec` now persist the full command string to `~/.fleet/fleet.log` (0644). These are the user's own authenticated, local actions so risk is low, but commands can carry inline secrets (e.g. a token/password as an argument). Flagging as a conscious acceptance, not a change request.

Nice touch keeping the conventional-keys list in `flog.go`'s doc comment in sync with the new keys, and addressing the prior round's copy-in/webhook-name feedback in `deeff3c`.

internal/server/webhook.go
```diff
@@ -54,6 +54,7 @@ func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 
 	body, err := io.ReadAll(http.MaxBytesReader(w, r.Body, maxWebhookBodySize))
 	if err != nil {
+		flog.Warn("webhook: rejected", "reason", "body-too-large", "status", 413)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Minor consistency nit: the 404 (`unknown-name`) and 400 (`non-json`) rejections both log `"name", clipName(name)`, but this 413 record omits it. `name` is already available here (`firstPathSegment(r.URL.Path)`), so adding `"name", clipName(name)` would make all three `webhook: rejected` records uniformly filterable by name.

Separately: the empty-name `400` at line 50 (`if name == ""`) is the one inbound rejection that still has no `webhook: rejected` record — worth tracing too for completeness.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behaviour-level QA of `deeff3c` (head of `feat/flog-action-coverage`). The binary was built and driven inside a throwaway Docker container (`mcr.microsoft.com/devcontainers/go:2-1.25-bookworm`) against a real devcontainer instance — no unit/integration tests were run; every record below was triggered by a real user action and read back out of `~/.fleet/fleet.log`.

| Test | Result | Notes |
|------|--------|-------|
| `go build ./...` + `cmd/fleet` binary | ✅ Pass | All 21 changed files + whole tree compile clean |
| CLI `--help` / subcommands render | ✅ Pass | No startup errors |
| `fleet server` boots + writes startup records | ✅ Pass | `mcp server started`, `fleet server started` land in fleet.log |
| `fleet server start failed` record | ✅ Pass | Forced `remove-socket` step → `ERROR …start failed step=remove-socket`, exit 1 |
| `watch subscribed` / `watch unsubscribed` (TUI) | ✅ Pass | Both logged with `runtime=true` on TUI open/quit |
| `fleet up` → real devcontainer instance | ✅ Pass | Instance reaches `running`, sibling container up |
| `fleet exec` → `container exec` record | ✅ Pass | Pre-existing record, no regression |
| `fleet logs` → `logs opened` record | ✅ Pass | `logs opened … follow=false` |
| `file copied out` record | ✅ Pass | `path=/etc/hostname ms=…` accurate |
| `file copied in` record | ❌ Fail | **Logged `path` is wrong on copy-and-rename** (see below) |
| Lifecycle `stop`/`start` + no spurious failure log | ✅ Pass | Works; no false `instance status change failed` on success |
| Log hygiene (no flood / no panics) | ✅ Pass | 30 lines for a full session, 0 panics, `WriteWarn`→log mirroring works |

**Overall: ❌ FAIL — 1 of 12 scenarios failing** (`file copied in` records a destination path that never existed on disk in the copy-and-rename case — a real accuracy defect in a record this PR adds, whose whole purpose is a trustworthy, greppable trail).

---

### ❌ `file copied in` logs an incorrect destination path on copy-and-rename

`CopyInto`'s single-file branch computes the logged path **before** destination resolution:

```go
destPath := path.Join(open.GetDest(), open.GetName())   // logged
...
finalDir, finalName, err := s.resolveCopyIntoDest(inst, open.GetDest(), open.GetName())  // actual write
```

When the client copies to a **full file path** (the rename form, e.g. `fleet copy ./x alpha:/tmp/y.txt`), the resolver correctly writes a single file at the requested path, but the log line joins `dest` + the *source* name, producing a path that points one level too deep into a "directory" that is actually the destination file.

**Reproduction** (verified live against a running `itest-fleet/alpha`):

```
$ echo rename-content > orig.txt
$ fleet copy orig.txt itest-fleet/alpha:/tmp/renamed.txt
Copied .../orig.txt -> /tmp/renamed.txt (15 bytes)

# inside the container — the file really is /tmp/renamed.txt:
$ fleet exec itest-fleet/alpha -- ls -la /tmp/renamed.txt   ->  -rw-r--r-- ... /tmp/renamed.txt
$ fleet exec itest-fleet/alpha -- ls /tmp/renamed.txt/       ->  Not a directory

# but fleet.log says:
msg="file copied in" fleet=itest-fleet instance=alpha path=/tmp/renamed.txt/orig.txt ms=898
```

Observed vs. expected:

| Copy form | Actual write | Logged `path` | |
|-----------|--------------|---------------|---|
| dir dest — `alpha:/tmp/` | `/tmp/dirfile.txt` | `/tmp/dirfile.txt` | ✅ correct |
| rename dest — `alpha:/tmp/renamed.txt` | `/tmp/renamed.txt` | `/tmp/renamed.txt/orig.txt` | ❌ wrong |
| rename dest — `alpha:/tmp/uploaded.txt` | `/tmp/uploaded.txt` | `/tmp/uploaded.txt/upload.txt` | ❌ wrong |

Why it matters: an operator grepping the audit trail for the file that was actually written (`/tmp/renamed.txt`) finds nothing, while the logged path (`/tmp/renamed.txt/orig.txt`) is a path that never existed. The rename form is a first-class, documented usage mode (it's in `fleet copy --help`'s own examples), and the PR description/review-response explicitly claims *"The logged destination is the client-requested target"* — but here the client target was `/tmp/renamed.txt`, not `/tmp/renamed.txt/orig.txt`. The same applies to the symmetric `file copy in failed` line. The copy operation itself is correct — **this is logging-accuracy only**, but it's squarely the defect class this PR exists to prevent.

Suggested fix: log `path.Join(finalDir, finalName)` after `resolveCopyIntoDest`. To keep tracing pre-resolution failures (the reason the `defer` was placed early), update `destPath` once resolution succeeds, or log the resolved path on the success branch and the raw requested dest on the error branch.

---

### ✅ What passed (detail)

- **Build / run gate.** `go build ./...` and the `cmd/fleet` binary both compile clean; `fleet --help` and all subcommands render.
- **Server lifecycle records.** A clean boot writes `mcp server started` + `fleet server started`. Forcing the new failure path (a non-empty dir at the socket path) produced `ERROR msg="fleet server start failed" step=remove-socket err="… directory not empty"` and exit 1 — the mkdir/listen/chmod/crashed siblings are the identical one-line pattern.
- **`watch subscribed` / `watch unsubscribed`.** Driving the TUI in tmux logged `watch subscribed … runtime=true` on open and `watch unsubscribed … runtime=true` on `q`; TUI exits cleanly.
- **Data plane on a real instance.** `fleet up` created a live devcontainer; `fleet logs` → `logs opened … follow=false`; `fleet copy <inst>:/etc/hostname ./out` → `file copied out path=/etc/hostname ms=440`; `fleet exec` still emits the pre-existing `container exec`. Copy-to-a-directory logs the correct path (only the rename form is wrong).
- **No regressions / hygiene.** `stop`→`start` round-trips and does **not** emit a false `instance status change failed`. A full session produced 30 log lines (no flooding), zero panics/goroutine dumps. The `WriteWarn`→fleet.log mirroring works (a non-fatal `devcontainer post-start failed` warning, pre-existing code, now surfaces in the trail as intended).

### ⏭️ Not exercised in this harness (no defect found; flagged for transparency)

These records need surfaces this container couldn't reach (MCP client, control socket, remote-gateway tunnel, automation fire, or specific cache settings) and were **not** directly triggered — they were reviewed in-diff only: MCP `exec` / `session spawned` / `session exec`, `browser open requested` / `file copy requested`, `browser prepared`, `automation: agent launched` / `event file written`, `webhook: rejected`, `group layout deleted`, `last seen version set`, and the buildx / apt / image-cache `… configured` records. The server-start `mkdir`/`listen`/`chmod`/`crashed` steps share the verified `remove-socket` step's pattern.

BenjaminBenetti:
Good catch on the QA re-run — fixed in c874e60.

**`file copied in` path accuracy (the FAIL):** root cause was my earlier move of the outcome `defer` *before* `resolveCopyIntoDest` (done so pre-resolution failures get traced), which left the logged `destPath` as `path.Join(dest, name)` — correct for a directory dest, wrong for the rename form. Now `destPath` is reassigned to `path.Join(finalDir, finalName)` right after resolution, so a successful record names the file actually written; pre-resolution failures still log the requested target. Verified:

```
copy ./orig.txt a:/tmp/renamed.txt   ->  msg="file copied in" path=/tmp/renamed.txt   (was /tmp/renamed.txt/orig.txt)
copy ./x a:/tmp/                      ->  msg="file copied in" path=/tmp/x             (unchanged, still correct)
```

**Also addressed the second code review's notes:**
- Empty-name `400` now logs `webhook: rejected reason=empty-name` (was the one untraced rejection path).
- `body-too-large` 413 now includes `name` for parity with the 404/400 records.
- `clipName` truncates on a rune boundary instead of a byte boundary.

The "late `SendAndClose` failure after a successful rename logs failure" edge you flagged as acceptable — left as-is (the RPC genuinely didn't confirm). Secrets-in-`exec`-logs likewise left as the conscious acceptance.

Build / vet / gofmt / golangci-lint / `internal/server` tests all green locally.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the diff for PR #222 (`feat(flog): log previously-untraced actions to fleet.log`) — 21 files, +192/-36. This is a clean, well-scoped observability pass. `go build ./...` passes locally. A few things stand out as worth a look before merge — none are blockers.

### What's good
- The `(configured bool, err error)` refactor on `ConfigureInstanceBuildx`/`ConfigureInstanceApt` correctly mirrors imagecache's existing pattern, so only *real* wiring is logged and the silent docker/apt-absent skips stay quiet. Tests updated to match.
- The flood-avoidance discipline is mostly right: `WriteWarnQuiet` for the per-tick reinstall poller, and the control-socket `browser open`/`file copy` records live inside the per-envelope `control.Listen` handler (registered once per instance), **not** in `syncRunning`'s poll loop — so they fire per real event, not per tick. 👍
- Respects the client/server log-ownership boundary (`watch.go` notes "client code must not write the server-owned event log"); all new writes are server-side.

### Findings

**1. (Medium) `CopyFile`/`CopyInto` log benign client cancellations at `ERROR`.**
`streamTarFile` returns client send errors verbatim, and a client that disconnects or Ctrl-C's mid-copy makes the RPC return a non-nil error. The new `defer` then emits `flog.Error("file copy out failed", …)` / `"file copy in failed"` for what is ordinary, expected behaviour — producing misleading ERROR records on every cancelled transfer. This contradicts the codebase's own established convention (see `probeFailureIsAlarming` in `schedule.go`: "a shutdown-driven cancel … stays quiet"). Suggest detecting `context.Canceled` / `codes.Canceled` (and likely client-gone `codes.Unavailable`) and either skipping the record or logging it at info/debug. (copy.go:60, 298, 314 — inline.)

**2. (Low–Medium) Unauthenticated webhook rejections are an unbounded-count log vector.**
`clipName` correctly bounds each line's *length*, but the `webhook: rejected` records fire on the unauthenticated path before any trigger is matched — in particular the `body-too-large` (413) record fires for **any** name, including completely unknown ones. A prober can spray the endpoint and append one WARN line to `fleet.log` per request, without limit on count. Worth considering rate-limiting/sampling these pre-auth rejections, or dropping the unknown-name/oversized cases to debug, so the endpoint can't be used to grow `fleet.log` arbitrarily. (webhook.go:58 — inline.)

**3. (Low) `exec` / `session exec` now persist full command lines to `fleet.log`.**
`flog.Info("exec", … "cmd", strings.Join(in.Command, " "))` (mcp_tools.go:646) and `"session exec", … "cmd", in.Command` (mcp_tools.go:709) write the complete command, including any secret passed as an argument (tokens, `-p<password>`, etc.), into a persistent file. It's roughly at parity with shell history and the log is local/per-user, so this may be acceptable — just flagging it as an intentional decision to confirm against the threat model, since `fleet.log` can be copied/shipped.

**4. (Nit) Trace coverage is slightly asymmetric vs. the doc comment.**
`CopyInto`'s comment says "validation, path resolution … and transfer failures are all traced," but the `defer` is registered only *after* the dir/file split, so the earlier `stream.Recv` / `resolveServerInstance` / not-running rejections (and `CopyFile`'s `resolveServerInstance` failure, before its defer) go untraced. Minor — consider tightening the comment or moving the defer earlier if those failures matter.

**5. (Observation) `watch subscribed`/`unsubscribed` can churn on reconnect.**
The TUI's `watchOnce` reconnects with backoff on every stream drop, so a flapping remote gateway yields repeated subscribe/unsubscribe pairs. Backoff bounds the rate, so probably fine — just noting it as the most likely source of repetitive lines among the new records.

Comments only — not requesting changes. Nice, careful PR overall.

internal/server/copy.go
```diff
@@ -46,12 +48,20 @@ const copyChunkSize = 64 * 1024
...
+func (s *service) CopyFile(req *fleetgrpc.CopyFileRequest, stream grpc.ServerStreamingServer[fleetgrpc.CopyFileChunk]) (err error) {
+	start := time.Now()
 	inst, err := resolveServerInstance(req.GetFleet(), req.GetInstance())
 	if err != nil {
 		return err
 	}
 	filePath := req.GetPath()
+	defer func() {
+		if err != nil {
+			flog.Error("file copy out failed", "fleet", req.GetFleet(), "instance", req.GetInstance(), "path", filePath, "err", err)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This `defer` logs **every** non-nil return at `ERROR` level — but `streamTarFile` forwards client `stream.Send` errors verbatim, so a client that disconnects or Ctrl-C's mid-copy is the most common way this path returns an error. That makes ordinary, expected cancellations show up as `file copy out failed` / `file copy in failed` ERROR records (same at lines 298 and 314 in `CopyInto`).

This also runs against the convention already established in this repo — see `probeFailureIsAlarming` in `schedule.go`, which deliberately keeps a "shutdown-driven cancel (`ctx.Err() == context.Canceled`) … quiet."

Consider checking for `context.Canceled` / `codes.Canceled` (and likely client-gone `codes.Unavailable`) in the defer and either skipping the record or logging it at info/debug, so genuine copy failures stay distinguishable from routine client disconnects.

internal/server/webhook.go
```diff
@@ -48,12 +48,14 @@ func (s *service) webhookHandler() http.Handler {
...
 	name := firstPathSegment(r.URL.Path)
 	if name == "" {
+		flog.Warn("webhook: rejected", "reason", "empty-name", "status", 400)
 		http.Error(w, "webhook name required", http.StatusBadRequest)
 		return
 	}
 
 	body, err := io.ReadAll(http.MaxBytesReader(w, r.Body, maxWebhookBodySize))
 	if err != nil {
+		flog.Warn("webhook: rejected", "reason", "body-too-large", "name", clipName(name), "status", 413)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

`clipName` nicely bounds each line's length, but this rejection fires on the unauthenticated path **before** any trigger name is matched, and the `body-too-large` (413) case fires for *any* name — including names that don't correspond to any configured webhook. So a prober can append one WARN line to `fleet.log` per request with no bound on the number of lines (only their length).

Worth considering rate-limiting/sampling these pre-auth rejection records, or dropping the unknown-name / oversized cases to debug, so an unauthenticated caller can't use this endpoint to grow `fleet.log` arbitrarily. The same applies to the `unknown-name` (404) record at line 74.

BenjaminBenetti:
Addressed the 3rd review's Medium finding in 8e8444a.

**#1 — copy cancellations logged as ERROR:** correct call. A client disconnect / Ctrl-C mid-copy returns a non-nil error (stream.Send / cancelled ctx), so the defer was emitting `ERROR "file copy ... failed"` for ordinary cancellations. The three copy log sites now route through a `logCopyOutcome` helper that classifies `context.Canceled` / `codes.Canceled` / `codes.Unavailable` as a benign cancellation (`info "file copy ... canceled"`) and reserves ERROR for genuine failures — matching the `probeFailureIsAlarming` convention you cited.

**#4 — comment/coverage:** the single-file `CopyInto` comment is scoped to that branch's failures (validation/resolution/transfer), which are all now traced; pre-header `stream.Recv` failures can't carry instance context so they stay untraced by design.

**#5 — watch churn:** left as-is; the TUI reconnect backoff bounds the rate.

Two items I've intentionally **not** changed because they trim/gate logs that were explicit asks — flagging for your decision rather than silently reversing:

- **#2 unauthenticated webhook rejection volume.** `clipName` bounds line length but not count; a prober with the capability URL can append one WARN per request (bounded only by the daily log rotation + 100-day retention). Options if you want it hardened: (a) rate-limit/sample the `webhook: rejected` records, (b) drop unknown-name/oversized to a quieter signal, or (c) leave as-is (capability-gated + rotated). 
- **#3 secrets in `exec`/`session exec` cmd.** Full command line persisted to `fleet.log` (0644, local/per-user) — at parity with shell history, but `fleet.log` can be shipped. Can redact or gate the `cmd` attribute behind a setting if you prefer.

Build / vet / gofmt / golangci-lint / `internal/server` tests green locally.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the latest push (`8e8444a`) on top of the prior three review/QA rounds. This is a clean, well-scoped logging PR, and the earlier findings have all landed in the current head — confirmed in-diff:

- **Copy cancel-vs-error (`8e8444a`)** — `logCopyOutcome`/`isClientGone` now route `context.Canceled` / `codes.Canceled` / `codes.Unavailable` to an info-level `"file copy … canceled"` record and reserve ERROR for genuine failures, matching the `probeFailureIsAlarming` convention. Correct, and applied symmetrically across all three copy sites (CopyFile defer, CopyInto dir branch, CopyInto single-file defer) with no double-logging.
- **`file copied in` rename-path defect (the QA FAIL → `c874e60`)** — `destPath` is reassigned to `path.Join(finalDir, finalName)` after `resolveCopyIntoDest`, so the closure logs the file actually written while pre-resolution failures still log the requested target. Verified in-diff. → The `qa-failed` label now appears **stale** (it predates this fix); worth clearing if a fresh QA pass confirms.
- **Named-return plumbing** — `CopyFile`/`CopyInto` use `(err error)` named returns; every exit is `return <expr>`, and the `inst, err :=` / `finalDir, finalName, err :=` sites reuse the function-scope `err` rather than shadowing it, so the deferred logger always observes the real returned value. ✓
- **Supporting refactors** — the `(configured bool, err error)` change on `ConfigureInstanceBuildx`/`ConfigureInstanceApt` (with tests updated) correctly logs only real wiring, and the `WriteWarn` → `writeWarnFile`/`WriteWarn`/`WriteWarnQuiet` split is sound: the only per-tick caller (`runtime_poller`) uses the quiet variant, and the remaining `WriteWarn` sites are all one-shot create/start/clone paths, so the new fleet.log mirroring won't flood.
- **Webhook hardening** — empty-name 400 now traced, 413 carries `name` for parity, and `clipName` truncates on a rune boundary. ✓

**One small forward-looking nuance** (non-blocking): in `isClientGone`, treating *all* `codes.Unavailable` as a benign cancellation can in principle misclassify a genuine mid-transfer transport/backend failure as `info "file copy … canceled"`. For `CopyInto` the write is atomic (temp file + rename-on-clean-finish), so a misclassified failure still means "no file written" — accurate enough — but the audit line would under-state it. In practice backend exec failures tend to surface as `Internal`/exit-status rather than `Unavailable`, so the risk is low; just noting it as the one edge the cancellation heuristic can't fully distinguish.

The two consciously-deferred items you already flagged — the full `cmd` in `exec`/`session exec` records (secrets-in-argv) and the unauthenticated `webhook: rejected` line *count* (capability-gated + bounded by log rotation) — read as genuine owner-decisions rather than defects; no objection to leaving them as-is with the rationale recorded.

Nice work — nothing blocking from this pass.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behavior-level QA of `feat/flog-action-coverage` @ `c874e60`, run end-to-end inside the project's `ghcr.io/benjaminbenetti/fleet-man/qa` Docker-in-Docker image. I drove the real `fleet` binary (and a real gateway tunnel) and inspected `~/.fleet/fleet.log` after each action — no automated unit/integration tests were run.

| Test | Result | Notes |
|------|--------|-------|
| `go build ./...` + build `fleet` binary | ✅ Pass | Compiles clean; `--help` renders all commands |
| `fleet up` create instance (default settings) | ✅ Pass | Instance runs; **no spurious cache logs** (skip-path correctly returns `configured=false`) |
| `file copied out` (download) | ✅ Pass | `fleet=…instance=…path=…ms=…` |
| `file copied in` (single file upload) | ✅ Pass | Round-trip content verified |
| `file copied in` (directory upload) | ✅ Pass | Logged once with `path=<dest dir>` |
| `file copy out failed` (missing src) | ✅ Pass | ERROR level + RPC error detail |
| `file copy in failed` (missing dest dir) | ✅ Pass | ERROR + best-effort resolved path, as documented |
| `logs opened` (`follow=false` / `follow=true`) | ✅ Pass | Both variants logged with correct `follow` value |
| `watch subscribed` / `watch unsubscribed` (TUI) | ✅ Pass | Both fire on TUI connect + quit; TUI renders correctly |
| `instance apt cache configured` (debcache refactor) | ✅ Pass | **Headline `(configured,err)` refactor** — apt-cacher container spun up, proxy written, record fires, instance still creates |
| `webhook: rejected` — unknown-name (404) | ✅ Pass | `reason=unknown-name status=404` |
| `webhook: rejected` — body-too-large (413) | ✅ Pass | `reason=body-too-large status=413` (>1 MiB body) |
| `webhook: rejected` — non-json (400) | ✅ Pass | json-path trigger + non-JSON body → `reason=non-json status=400` |
| `clipName` oversized webhook name bounding | ✅ Pass | 200-char name → 128 chars + `...(clipped)` in log |
| Webhook positive path (`webhook: received`) | ✅ Pass | Valid JSON → `fired=1`, trigger fired |
| `automation: event file written` | ✅ Pass | `bytes=19` matches payload size |
| `automation: agent launched` | ✅ Pass | `trigger=jtrig`, launched via fired webhook |
| `webhook: rejected` — empty-name (400) | ⏭️ Skipped | Not reachable via the gateway — it returns its own 404 for an empty path segment before the daemon sees it (daemon guard is defensive) |
| MCP `exec` / `session spawned` / `session exec` | ⏭️ Skipped | Require an MCP client driving `fleet_exec`/`fleet_session_*` |
| `instance status change failed`, `server start failed/crashed`, `instance warning`, buildx/image-cache configured, control-socket (`browser open/file copy requested`, `browser prepared`), `group layout deleted`, `last seen version set` | ⏭️ Skipped | Failure-injection / image-capability / TUI-only paths not cleanly drivable in this harness |

**Overall: ✅ PASS — every drivable scenario behaved as expected (17 passed, 0 failed; remaining records skipped as not externally drivable). No defects or regressions found.**

---

## Detailed findings

### Environment
- Pulled `ghcr.io/benjaminbenetti/fleet-man/qa:latest`, ran `--privileged` (its own DinD daemon), mounted the checked-out PR workspace.
- Built `/tmp/fleet` from `c874e60`. All `fleet` commands run via `docker exec` (non-TTY); the TUI run inside a container-side `tmux` session.
- Fleet log inspected at `/home/qa/.fleet/fleet.log`.

### Build (✅)
`go build ./...` → exit 0; `go build -o /tmp/fleet ./cmd/fleet` → exit 0; binary runs and lists all subcommands.

### Create flow + structural refactor (✅)
- Default `fleet up qa1` created a running devcontainer. The create log shows the existing lifecycle records and **no** `instance … configured` lines — confirming the refactored `ConfigureInstanceBuildx`/`ConfigureInstanceApt` skip-path now correctly returns `configured=false` (no spurious log) while the instance still provisions.
- Re-ran with a `debCacheServer:true` fleet (debian/apt fixture). The apt-cacher-ng container started, apt probe returned `PRESENT`, the proxy drop-in was written, and the new record fired:
  `INFO "instance apt cache configured" fleet=debcache-fleet workspace=…` — and the instance created successfully (`instance created … ms=10732`). This exercises the `configured=true` branch of the headline `(configured bool, err error)` change end-to-end.

### Copy & logs (✅)
Round-trip upload/download verified byte-for-byte. Records observed:
```
INFO  file copied in  …path=/tmp/qa-uploaded.txt ms=1031
INFO  file copied out …path=/tmp/qa-uploaded.txt ms=546
INFO  file copied in  …path=/tmp/qadir-dest ms=1031        # directory branch
ERROR file copy out failed …path=/tmp/does-not-exist-xyz.txt err="…NotFound…"
ERROR file copy in failed  …path=/root/protected/nope.txt/qa-upload.txt err="…destination directory \"/root/protected\" does not exist"
INFO  logs opened …follow=false
INFO  logs opened …follow=true
```
The copy-in failure logs the pre-resolution best-effort `dest/name` path, matching the documented behavior for failures before path resolution.

### Watch (TUI) (✅)
Launching the TUI logged `watch subscribed runtime=true`; pressing `q` logged `watch unsubscribed runtime=true`. The TUI rendered the fleet/instance tree correctly (regression sanity).

### Webhook rejections (✅)
Stood up a real `fleet gateway` (cert-less HTTP) + enabled the remote webhook tunnel (`webhook_enabled:true`), then POSTed to the gateway-assigned `public_webhook_url`:
```
WARN webhook: rejected reason=unknown-name   name=no-such-hook-xyz status=404
WARN webhook: rejected reason=body-too-large name=anyname        status=413
WARN webhook: rejected reason=non-json       name=jhook          status=400
WARN webhook: rejected reason=unknown-name   name=nnnn…(128)…(clipped) status=404
INFO webhook: received name=jhook fired=1
```
- HTTP status codes returned to the client matched the logged `status` in every case (404/413/400/200).
- **`clipName`**: a 200-char webhook name was bounded to 128 runes + `...(clipped)` in the log — confirming the reviewer-requested log-injection hardening.
- The **empty-name (400)** daemon guard is unreachable through the gateway: an empty path segment yields the gateway's own `404 page not found` before the request reaches the daemon, so that record can't be triggered externally (the guard remains as defense-in-depth).

### Automation (✅ bonus)
The valid webhook fire drove the full automation path, surfacing two more new records:
```
INFO automation: event file written …path=/tmp/fleet-event-…Z bytes=19   # == size of {"action":"opened"}
INFO automation: agent launched     …trigger=jtrig
```

### Key-consistency & levels
Every record I observed uses the conventional keys documented in `flog.go` (`fleet, instance, path, ms, follow, runtime, reason, name, status, bytes, trigger, workspace`), at sensible levels (INFO for actions, WARN for rejected inbound events, ERROR for failures). No malformed lines, no torn records, no unexpected ERROR noise.

### Not covered (transparency)
The MCP `exec`/`session_*` records, the failure-injection records (`instance status change failed`, `server start failed/crashed`, `instance warning`), instance-capability-gated cache records (buildx, image-cache), and TUI-only mutations (`group layout deleted`, `last seen version set`) plus control-socket/browser records were not exercised — they need an MCP client, induced failures, specific instance images, or TUI interactions outside this harness's clean reach. Nothing observed suggests a problem with them; they are simply unverified here.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Fresh manual QA of `8e8444a` (HEAD of `feat/flog-action-coverage`). Built the binary and drove the **real `fleet` CLI + TUI against a live devcontainer instance** (`itest-fleet/alpha`, debian-12 fixture) inside an isolated privileged docker container, inspecting `~/.fleet/fleet.log` after each action.

| Test | Result | Notes |
|------|--------|-------|
| `go build ./...` + `cmd/fleet` + `--help` | ✅ Pass | Compiles clean; binary runs |
| Server lifecycle records | ✅ Pass | `mcp server started`, `fleet server started` on boot |
| Instance create records | ✅ Pass | `instance created` + `container exec` as before |
| `logs opened` | ✅ Pass | `fleet logs` → `logs opened … follow=false` |
| `file copied out` (absolute + dir dest) | ✅ Pass | Correct `path`, `ms` |
| `file copied in` — **rename form** (prior FAIL) | ✅ Pass | Now logs `path=/tmp/renamed.txt` (the file actually written) — earlier defect fixed & holds |
| `file copied in` — dir dest | ✅ Pass | `path=/tmp/indir/up.bin` |
| `file copy in failed` — genuine failure | ✅ Pass | Missing dest dir → correct `ERROR` |
| **Copy cancellation — copy IN (Ctrl-C an upload)** | ❌ **Fail** | Benign client cancel still logged as **`ERROR "file copy in failed"`** — the exact thing this commit claims to fix |
| Copy cancellation — copy OUT | ⚠️ Partial | In-flight send-cancel is code-correct (verbatim `Unavailable`/`Canceled`→info), but a 0-byte early cancel logs `ERROR … read EOF` |
| stop → start round-trip | ✅ Pass | `instance status changed` both ways; **no** false `… change failed` |
| Watch subscribe/unsubscribe (TUI) | ✅ Pass | `watch subscribed`/`unsubscribed runtime=true`; clean `q` exit |
| Hygiene (no panics / no flood) | ✅ Pass | 41 log lines over a full session, no panics/goroutine dumps |
| MCP / control-socket / automation / webhook / cache records | ⏭️ Skipped | Need an MCP client, the gateway tunnel, automation config, or specific cache wiring — not reachable in this harness; reviewed in-diff only |

**Overall: ❌ FAIL — 1 of 13 exercised scenarios failing (1 partial).** The copy-cancellation reclassification added in this PR's latest commit (`8e8444a`, the commit's sole purpose) is reproducibly defeated for the **copy-in / upload** direction.

---

### ❌ Copy-in cancellations are still logged at ERROR (the regression this commit claims to fix)

This commit's stated goal (and the code-review item it answers): *"a client disconnect / Ctrl-C mid-copy … now route through `logCopyOutcome` … classifies `context.Canceled` / `codes.Canceled` / `codes.Unavailable` as a benign cancellation (`info "file copy … canceled"`) and reserves ERROR for genuine failures."*

That works for **copy-out** (`streamTarFile` returns `stream.Send` errors verbatim, so the raw `Canceled`/`Unavailable` code reaches `isClientGone`). It does **not** work for **copy-in**: `writeCopyIntoTar` re-wraps the `stream.Recv` error before the deferred `logCopyOutcome` ever sees it:

```go
// internal/server/copy.go:650-655 (writeCopyIntoTar)
chunk, err := stream.Recv()
if err == io.EOF { break }
if err != nil {
    return status.Errorf(codes.Internal, "recv: %v", err)   // <- Canceled buried inside a codes.Internal status, via %v (string), not %w
}
```

`isClientGone` (copy.go:54) only inspects the **outer** status code and `errors.Is`:
- `status.Code(err)` → `codes.Internal` (not `Canceled`/`Unavailable`) → no match
- `errors.Is(err, context.Canceled)` → `false` (the cause was string-formatted with `%v`, so it isn't in the unwrap chain)

…so `logCopyOutcome` falls through to the `ERROR` branch. **Every** cancellation of an in-progress upload is misclassified.

**Reproduction** (live against `itest-fleet/alpha`, deterministic — 4/4 attempts, both SIGKILL-disconnect and SIGINT/Ctrl-C):

```bash
# 400 MB local source so the upload is in flight long enough to interrupt
head -c 400000000 /dev/zero > /tmp/biglocal.bin
fleet exec itest-fleet/alpha -- mkdir -p /tmp/incancel

fleet copy /tmp/biglocal.bin itest-fleet/alpha:/tmp/incancel/x.bin &   # start upload
sleep 0.8; kill -INT %1                                                # Ctrl-C it mid-stream
```

`~/.fleet/fleet.log` (observed, every run):

```
level=ERROR msg="file copy in failed" fleet=itest-fleet instance=alpha \
  path=/tmp/incancel/x.bin \
  err="rpc error: code = Internal desc = recv: rpc error: code = Canceled desc = context canceled"
```

**Expected:** `level=INFO msg="file copy in canceled" …` (parity with the copy-out direction and with this commit's own stated contract).

Why it matters: cancelling an in-progress upload (Ctrl-C, TUI close) is ordinary user behaviour, not an error. This is precisely the "misleading ERROR records on every cancelled transfer" the review flagged and this commit set out to remove — it's removed for downloads but not uploads. Since accurate log levels are the entire deliverable of this PR, this is a real miss.

**Suggested fix:** classify before re-wrapping — e.g. in `writeCopyIntoTar`, `if isClientGone(err) { return err }` (or `status.Errorf(status.Code(err), "recv: %v", err)`), or have `isClientGone` also scan the wrapped message / unwrap via `%w`. The same `codes.Internal` rewrap exists for the directory-upload path (`copyDirInto`, copy.go:480-485), so a dir upload cancel would misclassify too.

### ⚠️ Copy-out early cancellation (minor, related)

A copy-**out** cancelled *before any data streams* (during the meta/source-read phase) kills the `tar` reader, so `tr.Next()` returns EOF → `codes.NotFound "read … EOF"` → logged `ERROR "file copy out failed"`. Narrow window (the common in-flight cancel is handled correctly), but it's the same "benign cancel surfaced as ERROR" family. Observed twice when killing at 0 bytes transferred.

### ✅ What passed (detail)

- **Build / run gate.** `go build ./...` and `go build ./cmd/fleet` compile clean; `fleet --help` renders. (Note: `go build` needs `git config --global --add safe.directory` in the throwaway container because of the bind-mount UID — environment-only, not a code issue.)
- **Rename-form copy-in (the previously-FAILED defect) is fixed and holds.** `fleet copy ./orig.txt alpha:/tmp/renamed.txt` writes a real file at `/tmp/renamed.txt` (verified `-d` is false in-container) and logs `path=/tmp/renamed.txt` — no longer the bogus `/tmp/renamed.txt/orig.txt`. Dir-dest still logs `path=/tmp/indir/up.bin`.
- **Genuine copy failures still ERROR correctly.** Upload into a non-existent dir → `ERROR "file copy in failed" … err="… destination directory \"/tmp/no-such-dir\" does not exist"`.
- **`logs opened`** fires with `follow=false`; **`file copied out`** logs source `path` + `ms`.
- **Lifecycle hygiene.** `stop`→`start` logs `instance status changed from=running to=stopped` / `from=stopped to=running` and emits **no** false `instance status change failed`.
- **`watch subscribed` / `watch unsubscribed`.** TUI open logged `watch subscribed runtime=true`; `q` logged `watch unsubscribed runtime=true` and exited cleanly.
- **No flooding / no panics.** A full session produced 41 log lines, no goroutine dumps; the only repeated record is the expected `container exec`.

### ⏭️ Not exercised (no defect found; flagged for transparency)

Records needing a surface this CLI/TUI harness can't reach were reviewed in-diff only: MCP `exec` / `session spawned` / `session exec` (need an MCP client), `browser open requested` / `file copy requested` (control socket), `browser prepared`, `automation: agent launched` / `event file written` (automation fire), **`webhook: rejected`** (the webhook receiver is a `remote.NewChanListener()` reachable only through the gateway tunnel — no local port), `group layout deleted`, `last seen version set`, and the buildx / apt / image-cache `… configured` records.

---
_Tested in an isolated privileged docker container (vfs); host instance never ran the app. Container torn down after the run._

BenjaminBenetti:
Re: review finding #2 (unauthenticated webhook log volume) — dropped the unknown-name (404) and oversized-body (413) rejection records in 1f3d09d; those fire pre-match on an attacker-controllable name. Kept the empty-name and non-json (json-path misconfig, #207) rejections, which indicate a real delivery problem. `clipName` still bounds the non-json record's name. Re: #3 — keeping the full `exec`/`session exec` command line as the intended audit behavior. Build/vet/gofmt/golangci-lint/tests green.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Re-review of #222 @ `1f3d09d`** (new commit on top of the prior review rounds).

`go build ./...` passes locally. The triggering commit (`fix(flog): don't log unauthenticated webhook probe-noise rejections`) is a tight, well-reasoned follow-up: it drops the 404 (unknown-name) and 413 (oversized) WARN records a prior round flagged as externally-drivable probe noise on the unauthenticated webhook path. That correctly resolves the earlier finding, and `clipName` stays in use (the `non-json` path), so nothing went dead.

Confirmed the rest of the earlier feedback is landed in head:
- copy cancel-vs-error classification (`isClientGone` / `logCopyOutcome`) — benign cancels log at info, genuine failures at error;
- copy-in validation + resolved-path tracing (`destPath` reassigned after `resolveCopyIntoDest`);
- the `(configured bool, err error)` refactor on `ConfigureInstanceBuildx`/`ConfigureInstanceApt`, with callers logging only real wiring and the `default:` success arms in `imagecache_setup.go` / `lifecycle.go` reading correctly.
- `WriteWarn` going loud is safe: the only per-tick caller (runtime poller reinstall) was moved to `WriteWarnQuiet`; every other `WriteWarn` site is one-shot provisioning/lifecycle, and the imagecache polling callback fires once per window, so no flood is introduced.

**One follow-up on this commit's own reasoning** (left inline on `webhook.go`): the `empty-name` 400 is still logged at WARN, yet it shares the exact probe-noise property the commit just removed from 404/413 — it fires on the unauthenticated path for a request needing *zero* knowledge of any configured name. Not a blocker (fixed-size line, no attacker-controlled data), just a consistency gap worth a second look.

Nothing else new stands out — careful, well-documented series.

internal/server/webhook.go
```diff
@@ -48,12 +48,16 @@ func (s *service) webhookHandler() http.Handler {
 func (s *service) serveWebhook(w http.ResponseWriter, r *http.Request) {
 	name := firstPathSegment(r.URL.Path)
 	if name == "" {
+		flog.Warn("webhook: rejected", "reason", "empty-name", "status", 400)
```
BenjaminBenetti, line 51: 🔍 **CodeReview Agent**

This `empty-name` rejection is still logged at WARN, but it has the same externally-drivable property `1f3d09d` just removed from the 404/413 records. It fires on the unauthenticated path (per the file header, the public URL is the only capability) for a request carrying *no* name segment — so an unauthenticated prober hitting `POST /webhook/` (or the base path) appends one WARN line per request, needing zero knowledge of any configured name. That's arguably *more* probe-prone than the 404 you just dropped, not less.

The commit message's rationale for keeping it — "a delivery that reached a configured surface" — doesn't quite fit here: an empty name is rejected at the entry, before any state load or trigger match, so it never reaches a configured surface.

It's only a disk-growth concern (the line is a fixed constant with no attacker data, so no injection risk), so not a blocker. But for consistency with the principle this commit establishes, consider dropping this WARN too (or logging it quietly / sampled). By contrast the `non-json` record below is defensible as-is — it requires `nameFound == true`, so the caller already had to know a configured name.

BenjaminBenetti:
Re: the `empty-name` 400 consistency nit — fair observation. It's kept deliberately (maintainer call): unlike the dropped 404/413, the line is fixed-size with no attacker-controlled content, and reaching `serveWebhook` at all requires the capability (tunnel) URL, so it's a single bounded signal that the endpoint is being hit with malformed (name-less) requests rather than an arbitrary-growth vector. Happy to drop it for full symmetry with the 404/413 decision if you'd prefer — low stakes either way. Everything else from the re-review is confirmed landed; PR is green on `1f3d09d`.

Issue #179, opened 2026-06-17
Start / Stop instance does not run PostStart script


BenjaminBenetti:
When starting and stopping an instance the PostStart script does not run. Like other devcontainer tools the postStart script should run when the container starts. 

## Tech Specs 

likely need to update `fleetd` with proper start / stop handling.

PR #225, opened 2026-06-29
fix(coder): stop provisioning hanging forever — stdin-EOF-safe file transfer (#223)

BenjaminBenetti:
Fixes #223.

## Root cause
Coder instances sat in `creating` forever (no error). Any provisioning command that **reads stdin to EOF** — `cat > file`, `tar -xf -`, `base64 -d` — never terminates over `coder ssh`, because the coder agent's SSH server (gliderlabs/ssh) never half-closes a remote command's stdin. The first such step in `finishProvision` (the `fleet.rc` write, or the binary copy when sudo is available) blocked `CombinedOutput()` forever, so `markInstanceRunning` was never reached. devcontainer (local `docker exec`) and codespaces (OpenSSH) deliver stdin EOF, so only coder hung.

## Approach — `Backend.CopyFile` strategy seam
A new interface method lets callers transfer a file into an instance without caring about the transport:

| backend | transport |
|---|---|
| devcontainer | `cat`-over-stdin (local docker exec delivers EOF) |
| codespaces | `cat`-over-stdin via a **non-PTY** exec (its normal exec PTY mangles binary — fixes a latent bug too) |
| coder | `scp` over `coder ssh --stdio` (length-framed, EOF-independent), then atomic `mv`+`chmod` |

Wired into the two callers, with no backend branching at the call site:
- **Binary staging** (`fleetlaunch`): `CopyFile` to a writable `/tmp` path, then a stdin-free `mv` install into `/usr/bin` (keeps the sudo/ETXTBSY handling).
- **`fleet copy`** (`server.CopyInto`, single file): buffers the upload to a host temp with exact-size validation, then `CopyFile` → `fc` now works on coder.

**Small in-memory files** (`fleet.rc`, Claude/auggie hook scripts + `settings.json`) move to **base64-in-argv** (`InlineWriteScript`) — stdin-free and uniform across backends. All staging execs gained a **timeout backstop** so no future stdin-reader can wedge create again.

Also fixes the same stdin-EOF hang in the automation event-file write (`writeAutomationEventFile` → `CopyFile`), refuses **directory** `fleet copy` on coder cleanly (`FailedPrecondition`) instead of hanging, and rejects control characters in copy filenames so they can't corrupt the scp transfer.

## Tests
- Real-`/bin/sh` round-trips for the inline/stream writers (binary bytes + nasty quoted paths), size guard, `ShellQuote`.
- Coder buffering helper + unique-stage-name test.
- Fake-backend test proving `copyBinary` stages-then-installs without `cat >`, and `EnsureFleetRC` writes inline.
- `CopyInto`/automation suites updated to the new seam; coder dir-copy gate test.
- Full non-TUI suite + `go vet` green.

## Adversarial review
Ran a multi-agent review (5 dimensions → per-finding verification). All 7 confirmed findings addressed: non-unique coder temp (race), `writeAutomationEventFile` hang, orphaned-temp cleanup, coder dir-copy hang→clean refusal, control-char filenames; OpenSSH-scp requirement and the missing cancellation-context documented.

## Known limitations / follow-ups
- The **coder scp path can't be exercised here** (no coder instance) — best-effort + 5-min timeout means a wrong guess degrades to "not staged", never a hang. **Needs validation on a real coder workspace.**
- `CopyFile`-based **recursive (directory)** copy on coder (currently refused).
- Threading a cancellation **context** into `Backend.CopyFile`.

### DIFF
--- a/internal/agentdetect/claude_provisioner.go
+++ b/internal/agentdetect/claude_provisioner.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"path"
 	"strings"
+
+	"github.com/BenjaminBenetti/fleet-man/internal/backend"
 )
 
 // ===========================================
@@ -66,14 +68,15 @@ func NewClaudeProvisioner(exec ContainerExecutor) *ClaudeProvisioner {
 // path into settings.json keeps Claude Code's hook command field
 // independent of any shell expansion behaviour.
 func (p *ClaudeProvisioner) Provision() error {
-	scriptPath, err := p.resolveScriptPath()
+	home, err := p.resolveHome()
 	if err != nil {
-		return fmt.Errorf("resolve script path: %w", err)
+		return fmt.Errorf("resolve home: %w", err)
 	}
+	scriptPath := path.Join(home, FleetManScriptSuffix)
 	if err := p.dropHookScript(scriptPath); err != nil {
 		return fmt.Errorf("install hook script: %w", err)
 	}
-	if err := p.injectSettings(scriptPath); err != nil {
+	if err := p.injectSettings(home, scriptPath); err != nil {
 		return fmt.Errorf("edit settings.json: %w", err)
 	}
 	return nil
@@ -83,10 +86,14 @@ func (p *ClaudeProvisioner) Provision() error {
 // Private helpers
 // ===========================================
 
-// resolveScriptPath asks the container for $HOME and joins it with
-// FleetManScriptSuffix to produce the absolute path the hook script
-// will be dropped at and that settings.json will reference.
-func (p *ClaudeProvisioner) resolveScriptPath() (string, error) {
+// settingsSuffix is the home-relative path of Claude Code's settings file.
+const settingsSuffix = ".claude/settings.json"
+
+// resolveHome asks the container for $HOME. The hook script and settings.json
+// both live under it, and their absolute paths are baked into the in-container
+// commands (and into settings.json's hook command field) so nothing depends on
+// later shell expansion.
+func (p *ClaudeProvisioner) resolveHome() (string, error) {
 	out, err := p.exec.Run([]string{"sh", "-c", "echo \"$HOME\""}, nil)
 	if err != nil {
 		return "", err
@@ -95,31 +102,29 @@ func (p *ClaudeProvisioner) resolveScriptPath() (string, error) {
 	if home == "" {
 		return "", fmt.Errorf("container reported empty $HOME")
 	}
-	return path.Join(home, FleetManScriptSuffix), nil
+	return home, nil
 }
 
-// dropHookScript writes the embedded hook script bytes to scriptPath
-// and sets the executable bit. mkdir of the parent directory
-// guarantees the path exists even on a fresh container that has
-// never had ~/.fleet/scripts before.
-//
-// All three operations are bundled into a single shell invocation
-// to minimise round-trips to the container exec layer.
+// dropHookScript writes the embedded hook script bytes to scriptPath and sets
+// the executable bit. It uses an inline (base64-in-argv) write rather than
+// streaming the bytes over stdin: a stdin-reading `cat > file` hangs forever on
+// the coder backend, whose transport never delivers the stdin EOF (issue #223).
+// The write mkdir's the parent and is atomic.
 func (p *ClaudeProvisioner) dropHookScript(scriptPath string) error {
-	parentDir := path.Dir(scriptPath)
-	script := fmt.Sprintf(
-		`mkdir -p %q && cat > %q && chmod +x %q`,
-		parentDir, scriptPath, scriptPath,
-	)
-	_, err := p.exec.Run([]string{"sh", "-c", script}, ClaudeHookScript)
+	argv, err := backend.InlineWriteScript(scriptPath, ClaudeHookScript, 0o755)
+	if err != nil {
+		return err
+	}
+	_, err = p.exec.Run(argv, nil)
 	return err
 }
 
 // injectSettings reads the current settings.json (or treats it as
 // absent), passes it through InjectFleetManHooks, and writes the
 // result back atomically.
-func (p *ClaudeProvisioner) injectSettings(scriptPath string) error {
-	current, err := p.readSettings()
+func (p *ClaudeProvisioner) injectSettings(home, scriptPath string) error {
+	settingsPath := path.Join(home, settingsSuffix)
+	current, err := p.readSettings(settingsPath)
 	if err != nil {
 		return fmt.Errorf("read: %w", err)
 	}
@@ -127,30 +132,32 @@ func (p *ClaudeProvisioner) injectSettings(scriptPath string) error {
 	if err != nil {
 		return fmt.Errorf("compute: %w", err)
 	}
-	if err := p.writeSettings(updated); err != nil {
+	if err := p.writeSettings(settingsPath, updated); err != nil {
 		return fmt.Errorf("write: %w", err)
 	}
 	return nil
 }
 
-// readSettings cats ~/.claude/settings.json. A missing file
-// produces empty bytes (treated by InjectFleetManHooks as "the file
-// does not exist"), not an error — distinguishing "missing" from
-// "present and empty" inside a shell pipeline is not worth the
-// complexity since both yield the same downstream behaviour.
-func (p *ClaudeProvisioner) readSettings() ([]byte, error) {
-	const script = `if [ -f "$HOME/.claude/settings.json" ]; then cat "$HOME/.claude/settings.json"; fi`
+// readSettings cats settingsPath. A missing file produces empty bytes (treated
+// by InjectFleetManHooks as "the file does not exist"), not an error —
+// distinguishing "missing" from "present and empty" inside a shell pipeline is
+// not worth the complexity since both yield the same downstream behaviour. The
+// read produces output and exits, so it has no stdin-EOF dependency.
+func (p *ClaudeProvisioner) readSettings(settingsPath string) ([]byte, error) {
+	q := backend.ShellQuote(settingsPath)
+	script := fmt.Sprintf(`if [ -f %s ]; then cat %s; fi`, q, q)
 	return p.exec.Run([]string{"sh", "-c", script}, nil)
 }
 
-// writeSettings writes content to ~/.claude/settings.json
-// atomically: same-directory tmp file then rename. mkdir of
-// ~/.claude handles the first-time case where Claude Code has
-// never been run in this container.
-func (p *ClaudeProvisioner) writeSettings(content []byte) error {
-	const script = `mkdir -p "$HOME/.claude" && ` +
-		`cat > "$HOME/.claude/settings.json.tmp" && ` +
-		`mv "$HOME/.claude/settings.json.tmp" "$HOME/.claude/settings.json"`
-	_, err := p.exec.Run([]string{"sh", "-c", script}, content)
+// writeSettings writes content to settingsPath atomically (same-directory tmp
+// then rename, handled by the inline writer, which also mkdir's the parent for
+// the first-time case). Inline (base64-in-argv) rather than stdin-streamed so it
+// cannot hang on the coder backend (issue #223).
+func (p *ClaudeProvisioner) writeSettings(settingsPath string, content []byte) error {
+	argv, err := backend.InlineWriteScript(settingsPath, content, 0o644)
+	if err != nil {
+		return err
+	}
+	_, err = p.exec.Run(argv, nil)
 	return err
 }
--- a/internal/agentdetect/claude_provisioner.go
+++ b/internal/agentdetect/claude_provisioner.go
@@ -1,6 +1,7 @@
 package agentdetect
 
 import (
+	"bytes"
 	"fmt"
 	"path"
 	"strings"
@@ -150,14 +151,14 @@ func (p *ClaudeProvisioner) readSettings(settingsPath string) ([]byte, error) {
 }
 
 // writeSettings writes content to settingsPath atomically (same-directory tmp
-// then rename, handled by the inline writer, which also mkdir's the parent for
-// the first-time case). Inline (base64-in-argv) rather than stdin-streamed so it
-// cannot hang on the coder backend (issue #223).
+// then rename, plus a parent mkdir for the first-time case — both guaranteed by
+// the CopyFile transport). It streams via CopyFile rather than an inline
+// (base64-in-argv) write because the merged settings.json is the one unbounded
+// payload here: a user's pre-existing file can be arbitrarily large, and the
+// inline writer hard-errors above its ARG_MAX-derived cap. CopyFile is both
+// uncapped and stdin-EOF-safe on every backend including coder (issue #223), so
+// it carries a large settings.json without the hang the inline path was added to
+// avoid. The fixed-size hook script keeps using the inline writer.
 func (p *ClaudeProvisioner) writeSettings(settingsPath string, content []byte) error {
-	argv, err := backend.InlineWriteScript(settingsPath, content, 0o644)
-	if err != nil {
-		return err
-	}
-	_, err = p.exec.Run(argv, nil)
-	return err
+	return p.exec.CopyFile(bytes.NewReader(content), settingsPath, 0o644)
 }
--- a/internal/backend/backend.go
+++ b/internal/backend/backend.go
@@ -1,6 +1,9 @@
 package backend
 
-import "os/exec"
+import (
+	"io"
+	"os/exec"
+)
 
 // Backend defines the strategy interface for container runtimes.
 // Implementations handle provisioning, lifecycle, execution,
@@ -91,6 +94,25 @@ type Backend interface {
 	// where logging every command would flood the event log. Behaviour is
 	// otherwise identical to ExecCommand.
 	ExecCommandQuiet(workspaceDir string, command []string) *Cmd
+
+	// CopyFile streams src INTO the instance, writing it to remotePath with the
+	// given mode. It is the single strategy seam for host→instance file transfer:
+	// callers (the fleet-launch binary stager, the `fleet copy` server handler)
+	// invoke it without caring which transport a backend uses. The write is
+	// atomic (same-directory temp + rename), creates remotePath's parent, and
+	// assumes that parent is writable by the exec user (a caller needing a
+	// privileged destination copies to a writable staging path and moves it into
+	// place separately).
+	//
+	// Crucially, an implementation MUST NOT rely on the remote command reading
+	// stdin to EOF: the coder backend's SSH transport never half-closes a remote
+	// command's stdin, so a `cat > file` there hangs forever. devcontainer (local
+	// docker exec) and codespaces (OpenSSH, no PTY) stream over stdin via
+	// StreamWriteScript; coder transfers out-of-band with scp.
+	//
+	// This is the INTO direction (host → instance); it is unrelated to the
+	// fleetgrpc CopyFile RPC, which streams a file OUT of an instance.
+	CopyFile(workspaceDir string, src io.Reader, remotePath string, mode int) error
 
 	// Stats returns CPU and memory usage for the given container IDs.
 	Stats(containerIDs []string) (map[string]*ContainerStats, error)
--- a/internal/backend/coder/coder_copy.go
+++ b/internal/backend/coder/coder_copy.go
@@ -57,6 +57,15 @@ func remoteStageName(remotePath string) string {
 // CopyTimeout rather than a caller context — Backend.CopyFile takes no context,
 // so a client that aborts an in-flight `fleet copy` cannot cancel the scp before
 // that timeout; the bound keeps an abandoned copy from leaking indefinitely.
+//
+// remotePath should be ABSOLUTE. Every caller in the staging path passes one
+// (the binary stages to /tmp, the automation event file to /tmp). A relative
+// remotePath resolves against scp's / `coder ssh`'s login dir (home) on coder —
+// which differs from the workspace folder a devcontainer/codespaces write would
+// use, and from the path the `fleet copy` handler reports — so a `fleet copy`
+// with an empty/relative dest lands in an undefined spot on coder. That path is
+// part of the still-to-be-validated coder `fleet copy` support; the #223 fix
+// (binary + automation staging) is unaffected because it is always absolute.
 func (coderBackend *CoderBackend) CopyFile(workspaceDir string, src io.Reader, remotePath string, mode int) error {
 	if _, err := exec.LookPath("scp"); err != nil {
 		return fmt.Errorf("copy into %q: scp not found on host (required for coder file transfer): %w", remotePath, err)
@@ -80,6 +89,13 @@ func (coderBackend *CoderBackend) CopyFile(workspaceDir string, src io.Reader, r
 
 	ctx, cancel := context.WithTimeout(context.Background(), backend.CopyTimeout)
 	defer cancel()
+	// The remote operand (target:remoteTmp) is intentionally NOT shell-quoted,
+	// unlike the mkdir/install/cleanup execs around it: scp's default SFTP
+	// transfer (OpenSSH 9.0+) takes the post-colon path LITERALLY, so a `'..'`
+	// wrapper would become part of the filename. (Quoting is only right for the
+	// pre-9.0 legacy protocol, where a remote shell expands the path — which this
+	// code never opts into via -O.) remoteTmp derives from remotePath, so the
+	// staging-path callers (always /tmp) carry no spaces regardless.
 	scp := exec.CommandContext(ctx, "scp",
 		"-o", "StrictHostKeyChecking=no",
 		"-o", "UserKnownHostsFile=/dev/null",
--- a/internal/backend/coder/coder_copy.go
+++ b/internal/backend/coder/coder_copy.go
@@ -1,0 +1,139 @@
+package coder
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/BenjaminBenetti/fleet-man/internal/backend"
+)
+
+// coderPrepTimeout bounds the small stdin-free helper execs (mkdir, chmod+mv,
+// cleanup) that bracket the scp transfer. They read no stdin, so they cannot hit
+// the coder stdin-EOF hang; the deadline only guards against a stuck tunnel.
+const coderPrepTimeout = 30 * time.Second
+
+// coderCopySeq makes the remote staging name unique within this process so two
+// concurrent CopyFile calls to the same remotePath never collide on it.
+var coderCopySeq atomic.Uint64
+
+// remoteStageName returns a hidden, per-call-unique sibling of remotePath to scp
+// into before the atomic rename — the out-of-band analogue of the in-shell
+// `$$`-suffixed temp the stdin-streaming backends use.
+func remoteStageName(remotePath string) string {
+	var b [6]byte
+	_, _ = rand.Read(b[:])
+	return fmt.Sprintf("%s.fleet-scp.%d.%x", remotePath, coderCopySeq.Add(1), b)
+}
+
+// CopyFile transfers src INTO the coder workspace (its nested devcontainer) at
+// remotePath. It cannot stream over `coder ssh` stdin like the other backends:
+// the coder agent's SSH server (gliderlabs/ssh) never half-closes a remote
+// command's stdin, so a `cat > file` reading to EOF hangs forever (issue #223).
+//
+// Instead it uses scp, whose length-framed protocol needs no stdin EOF,
+// tunnelled over `coder ssh --stdio` as an OpenSSH ProxyCommand — the transport
+// coder documents for scp/rsync, which lands the file inside the nested
+// devcontainer agent. Sequence:
+//
+//  1. materialise src to a host file (an *os.File source is used directly);
+//  2. mkdir -p the destination's parent (scp will not create it);
+//  3. scp the host file to a sibling temp next to remotePath;
+//  4. chmod + atomic rename into place via a normal exec (stdin-free, hang-proof).
+//
+// scp defaults to the SFTP subsystem, which the coder agent provides; if a
+// particular template's nested agent lacks it the transfer fails cleanly (the
+// caller treats binary staging as best-effort) rather than hanging.
+//
+// Requirements & limitations: `scp` must be on the host PATH and be an
+// OpenSSH-compatible build (the `-o ProxyCommand` syntax is OpenSSH-specific;
+// e.g. Dropbear's scp would reject it). The scp deadline is the standalone
+// CopyTimeout rather than a caller context — Backend.CopyFile takes no context,
+// so a client that aborts an in-flight `fleet copy` cannot cancel the scp before
+// that timeout; the bound keeps an abandoned copy from leaking indefinitely.
+func (coderBackend *CoderBackend) CopyFile(workspaceDir string, src io.Reader, remotePath string, mode int) error {
+	if _, err := exec.LookPath("scp"); err != nil {
+		return fmt.Errorf("copy into %q: scp not found on host (required for coder file transfer): %w", remotePath, err)
+	}
+
+	localPath, cleanup, err := localFileForCopy(src)
+	if err != nil {
+		return fmt.Errorf("copy into %q: stage source: %w", remotePath, err)
+	}
+	defer cleanup()
+
+	name := coderWorkspaceName(workspaceDir)
+	target := coderBackend.resolveSSHTarget(name)
+	remoteTmp := remoteStageName(remotePath)
+
+	// scp cannot create the destination directory; make sure it exists first.
+	mkdir := fmt.Sprintf(`mkdir -p "$(dirname %s)"`, backend.ShellQuote(remotePath))
+	if out, err := coderBackend.ExecCommand(workspaceDir, []string{"sh", "-c", mkdir}).CombinedOutputWithTimeout(coderPrepTimeout); err != nil {
+		return fmt.Errorf("copy into %q: prepare dir: %w (%s)", remotePath, err, strings.TrimSpace(string(out)))
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), backend.CopyTimeout)
+	defer cancel()
+	scp := exec.CommandContext(ctx, "scp",
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+		"-o", "LogLevel=ERROR",
+		"-o", "ProxyCommand=coder ssh --stdio %h",
+		localPath, target+":"+remoteTmp,
+	)
+	if out, err := scp.CombinedOutput(); err != nil {
+		// scp may have written a partial remote temp before failing.
+		coderBackend.removeRemote(workspaceDir, remoteTmp)
+		return fmt.Errorf("copy into %q: scp: %w (%s)", remotePath, err, strings.TrimSpace(string(out)))
+	}
+
+	// chmod to the requested mode then atomically rename into place. Neither
+	// command reads stdin, so they are immune to the coder stdin-EOF hang.
+	install := fmt.Sprintf(`set -e; chmod %o %[2]s; mv -f %[2]s %[3]s`,
+		mode, backend.ShellQuote(remoteTmp), backend.ShellQuote(remotePath))
+	if out, err := coderBackend.ExecCommand(workspaceDir, []string{"sh", "-c", install}).CombinedOutputWithTimeout(coderPrepTimeout); err != nil {
+		// The rename did not complete; don't leave the staged temp behind.
+		coderBackend.removeRemote(workspaceDir, remoteTmp)
+		return fmt.Errorf("copy into %q: install: %w (%s)", remotePath, err, strings.TrimSpace(string(out)))
+	}
+	return nil
+}
+
+// removeRemote best-effort deletes a leftover staging file inside the workspace
+// after a failed transfer, so repeated failures don't accumulate orphans.
+func (coderBackend *CoderBackend) removeRemote(workspaceDir, path string) {
+	rm := fmt.Sprintf(`rm -f %s`, backend.ShellQuote(path))
+	_, _ = coderBackend.ExecCommand(workspaceDir, []string{"sh", "-c", rm}).CombinedOutputWithTimeout(coderPrepTimeout)
+}
+
+// localFileForCopy yields a host filesystem path for src that scp can read. When
+// src is already an *os.File (e.g. the open fleet binary) its path is used
+// directly with a no-op cleanup; any other reader is buffered into a temp file
+// the caller must remove via the returned cleanup.
+func localFileForCopy(src io.Reader) (path string, cleanup func(), err error) {
+	noop := func() {}
+	if f, ok := src.(*os.File); ok && f.Name() != "" {
+		return f.Name(), noop, nil
+	}
+	tmp, err := os.CreateTemp("", "fleet-copy-*")
+	if err != nil {
+		return "", noop, err
+	}
+	remove := func() { _ = os.Remove(tmp.Name()) }
+	if _, err := io.Copy(tmp, src); err != nil {
+		_ = tmp.Close()
+		remove()
+		return "", noop, err
+	}
+	if err := tmp.Close(); err != nil {
+		remove()
+		return "", noop, err
+	}
+	return tmp.Name(), remove, nil
+}
--- a/internal/backend/coder/coder_copy.go
+++ b/internal/backend/coder/coder_copy.go
@@ -19,6 +19,13 @@ import (
 // the coder stdin-EOF hang; the deadline only guards against a stuck tunnel.
 const coderPrepTimeout = 30 * time.Second
 
+// coderScpWaitDelay is the grace period after the scp deadline fires (when
+// CommandContext kills scp) before Go force-closes the inherited stdout/stderr
+// pipe and lets cmd.Wait() return — so a hung tunnel can't keep the call blocked
+// past CopyTimeout. Generous relative to backend.Cmd's own 3s WaitDelay because
+// scp's nested `coder ssh --stdio` teardown can be slower than a local exec.
+const coderScpWaitDelay = 10 * time.Second
+
 // coderCopySeq makes the remote staging name unique within this process so two
 // concurrent CopyFile calls never collide on it.
 var coderCopySeq atomic.Uint64
@@ -124,6 +131,14 @@ func (coderBackend *CoderBackend) CopyFile(workspaceDir string, src io.Reader, r
 		"-o", "ProxyCommand=coder ssh --stdio %h",
 		localPath, target+":"+remoteTmp,
 	)
+	// Make the ctx deadline real. On timeout, CommandContext kills only scp; its
+	// ssh / `coder ssh --stdio` children inherit the stdout/stderr pipe, so
+	// without WaitDelay cmd.Wait() (inside CombinedOutput) would block on the
+	// output-copy goroutine until those children close it — a stuck tunnel could
+	// hang past CopyTimeout, the exact failure class this file removes. WaitDelay
+	// force-closes the pipe shortly after the kill, mirroring
+	// CombinedOutputWithTimeout's own WaitDelay.
+	scp.WaitDelay = coderScpWaitDelay
 	if out, err := scp.CombinedOutput(); err != nil {
 		// scp may have written a partial remote temp before failing.
 		coderBackend.removeRemote(workspaceDir, remoteTmp)
--- a/internal/backend/coder/coder_copy.go
+++ b/internal/backend/coder/coder_copy.go
@@ -20,16 +20,27 @@ import (
 const coderPrepTimeout = 30 * time.Second
 
 // coderCopySeq makes the remote staging name unique within this process so two
-// concurrent CopyFile calls to the same remotePath never collide on it.
+// concurrent CopyFile calls never collide on it.
 var coderCopySeq atomic.Uint64
 
-// remoteStageName returns a hidden, per-call-unique sibling of remotePath to scp
-// into before the atomic rename — the out-of-band analogue of the in-shell
-// `$$`-suffixed temp the stdin-streaming backends use.
-func remoteStageName(remotePath string) string {
+// coderStageDir is the fixed directory CopyFile scp's into before the install
+// rename. It is deliberately NOT a sibling of remotePath: the scp remote operand
+// is the one path in this file that cannot be ShellQuote'd (see the scp call),
+// and remotePath's parent can legitimately contain spaces — a `fleet copy` dest
+// directory is only `test -d`-probed, not character-restricted. Staging instead
+// in /tmp, whose name carries no shell metacharacters, keeps that operand safe
+// under both scp transports without quoting. /tmp is writable on every backend
+// (the binary and automation-event staging paths already rely on it).
+const coderStageDir = "/tmp"
+
+// remoteStageName returns a hidden, per-call-unique path in coderStageDir to scp
+// into before the install rename — the out-of-band analogue of the in-shell
+// `$$`-suffixed temp the stdin-streaming backends use. Its characters are
+// restricted to [0-9a-f.] so the unquoted scp operand is transport-safe.
+func remoteStageName() string {
 	var b [6]byte
 	_, _ = rand.Read(b[:])
-	return fmt.Sprintf("%s.fleet-scp.%d.%x", remotePath, coderCopySeq.Add(1), b)
+	return fmt.Sprintf("%s/.fleet-scp.%d.%x", coderStageDir, coderCopySeq.Add(1), b)
 }
 
 // CopyFile transfers src INTO the coder workspace (its nested devcontainer) at
@@ -44,8 +55,15 @@ func remoteStageName(remotePath string) string {
 //
 //  1. materialise src to a host file (an *os.File source is used directly);
 //  2. mkdir -p the destination's parent (scp will not create it);
-//  3. scp the host file to a sibling temp next to remotePath;
-//  4. chmod + atomic rename into place via a normal exec (stdin-free, hang-proof).
+//  3. scp the host file to a fixed metacharacter-free temp in /tmp;
+//  4. chmod + rename it into place via a normal exec (stdin-free, hang-proof).
+//
+// The rename in step 4 (not the scp in step 3) targets remotePath, so a `fleet
+// copy` destination that contains spaces is handled by the ShellQuote'd `mv` —
+// the scp operand never sees it. Because the temp lives in /tmp rather than
+// beside remotePath, that `mv` may cross filesystems (a copy+unlink, not an
+// atomic rename) when remotePath is outside /tmp; that is acceptable for the
+// best-effort staging this serves and matches no concurrent-reader expectation.
 //
 // scp defaults to the SFTP subsystem, which the coder agent provides; if a
 // particular template's nested agent lacks it the transfer fails cleanly (the
@@ -79,7 +97,7 @@ func (coderBackend *CoderBackend) CopyFile(workspaceDir string, src io.Reader, r
 
 	name := coderWorkspaceName(workspaceDir)
 	target := coderBackend.resolveSSHTarget(name)
-	remoteTmp := remoteStageName(remotePath)
+	remoteTmp := remoteStageName()
 
 	// scp cannot create the destination directory; make sure it exists first.
 	mkdir := fmt.Sprintf(`mkdir -p "$(dirname %s)"`, backend.ShellQuote(remotePath))
@@ -90,12 +108,15 @@ func (coderBackend *CoderBackend) CopyFile(workspaceDir string, src io.Reader, r
 	ctx, cancel := context.WithTimeout(context.Background(), backend.CopyTimeout)
 	defer cancel()
 	// The remote operand (target:remoteTmp) is intentionally NOT shell-quoted,
-	// unlike the mkdir/install/cleanup execs around it: scp's default SFTP
-	// transfer (OpenSSH 9.0+) takes the post-colon path LITERALLY, so a `'..'`
-	// wrapper would become part of the filename. (Quoting is only right for the
-	// pre-9.0 legacy protocol, where a remote shell expands the path — which this
-	// code never opts into via -O.) remoteTmp derives from remotePath, so the
-	// staging-path callers (always /tmp) carry no spaces regardless.
+	// unlike the mkdir/install/cleanup execs around it. The two scp transports
+	// disagree on how to read it: the default SFTP transfer (OpenSSH 9.0+) takes
+	// the post-colon path LITERALLY, so a `'..'` wrapper would become part of the
+	// filename, while the legacy pre-9.0 protocol shell-expands it (where quoting
+	// would be required). No single quoting choice is right for both — so instead
+	// remoteTmp is constructed to need none: it lives in /tmp with only [0-9a-f.]
+	// characters (see remoteStageName), which is safe verbatim under either
+	// transport. The spacey final path is reached by the ShellQuote'd `mv` below,
+	// never by scp.
 	scp := exec.CommandContext(ctx, "scp",
 		"-o", "StrictHostKeyChecking=no",
 		"-o", "UserKnownHostsFile=/dev/null",
--- a/internal/backend/copy_file.go
+++ b/internal/backend/copy_file.go
@@ -1,0 +1,104 @@
+package backend
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// copy_file.go holds the transport-agnostic helpers behind the Backend.CopyFile
+// strategy and the small-file inline writer. They exist because a remote command
+// that reads stdin until EOF (e.g. `cat > file`) hangs forever on transports
+// that do not half-close a remote command's stdin — notably `coder ssh`, whose
+// agent SSH server never delivers the stdin EOF. Two stdin-EOF-free techniques
+// live here:
+//
+//   - WriteFileInline / InlineWriteScript embed the payload in the command as
+//     base64 (decoded in-container), so no stdin is involved at all. Use it for
+//     SMALL files; the whole command line is bounded by the host ARG_MAX.
+//
+//   - StreamWriteScript is the classic stdin-streamed write (`cat > tmp`) used by
+//     backends whose transport DOES deliver stdin EOF (devcontainer's local
+//     docker exec, codespaces' OpenSSH). The coder backend cannot use it and
+//     ships its own scp-based CopyFile instead.
+
+// CopyTimeout bounds a single CopyFile transport so a wedged copy can never stall
+// instance provisioning (which is otherwise best-effort but has no deadline of
+// its own). Generous enough for a multi-tens-of-MB binary over a slow tunnel.
+const CopyTimeout = 5 * time.Minute
+
+// maxInlineRaw caps the raw payload an inline (base64-in-argv) write accepts.
+// The base64 expansion (+~33%) plus the surrounding script must fit comfortably
+// under the host ARG_MAX (as low as 128KB on some systems), so the cap is kept
+// well below that. Every inline caller (fleet.rc, agent hook scripts and
+// settings.json) is a few KB, so this is never hit in practice; it exists to
+// fail loudly rather than silently exceed ARG_MAX if a payload ever grows.
+const maxInlineRaw = 64 * 1024
+
+// ShellQuote wraps value in single quotes with embedded single quotes escaped
+// via the '\” idiom, so any path is safe to drop into a /bin/sh command
+// literal. Exported so backend implementations (e.g. coder's scp install step)
+// share one quoting routine; kept here rather than imported from
+// internal/dotfiles to avoid that package's transitive state dependency.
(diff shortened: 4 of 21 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Solid, well-scoped fix. The root cause (no stdin half-close over `coder ssh`) is correctly identified, and the `Backend.CopyFile` strategy seam is a clean way to keep transport choice out of the call sites. Atomic temp+rename writes, the per-backend timeout backstops, the codespaces non-PTY fix, and the control-char filename guard are all good. I built the PR head locally (`36a7d222`): `go build ./...`, `go vet`, and the tests for `backend/…`, `server`, `fleetlaunch`, and `agentdetect` are all green.

Findings below are mostly on the coder path, which you already flag as unvalidated.

---

**1. (medium) Coder + a relative/empty `fleet copy` dest can land the file in the wrong dir and report an untruthful path.**

`resolveCopyIntoDest` returns `finalDir="."` for an empty dest, so `finalPath` becomes a bare relative name. On devcontainer/codespaces the in-container write runs via an exec whose cwd is the workspace folder, so a relative name lands there. On coder, `CopyFile` does `scp <target>:<relativeName>` — scp resolves a relative remote path against the **SSH login dir** (home), and the bracketing `coder ssh` `mkdir`/`mv` execs also start from the login dir, not the workspace folder. Two consequences:
- the file lands under `$HOME` instead of the workspace dir, diverging from the other backends;
- `reportedPath` still resolves the relative dest against `inst.WorkspaceDir`, so the reply advertises a path the file isn't at.

If scp's cwd and the exec's cwd ever disagree, the install `mv` just fails (clean error, not a hang), so it degrades safely — but the result is wrong/misleading. Anchoring a relative dest to an absolute path *before* calling `CopyFile` would make every backend agree and keep the reply truthful. Worth confirming during the coder validation you've already scheduled.

**2. (low) `stagePath` is a fixed constant, unlike every other staging name in this PR.**

`InlineWriteScript` uses `.fleet-inline.$$`, coder's `remoteStageName` uses a seq+random suffix, and the old copyinto temp was randomized — but the binary stage is a constant `/tmp/.fleet-launch-stage`. Two concurrent `copyBinary` calls into the same instance could clobber each other's stage between `CopyFile` and the install `mv`. Concurrent staging into one instance is unlikely (low severity), but a `$$`/random suffix would match the rest of the PR's care around uniqueness.

---

**Nits (non-blocking):**
- The `maxInlineRaw` (64 KiB) cap is enforced on the agent `settings.json` inline write; a user with an unusually large existing `settings.json` would fail provisioning with a clear error (degraded, never hung) — acceptable and already documented, just noting the user-visible edge.
- `base64 -d` is the GNU/busybox spelling — correct for the Linux in-container target, but the host-side round-trip tests (`TestInlineWriteScriptRoundTrip`, etc.) would fail on a BSD/macOS host where `base64` expects `-D`. CI is Linux, so this is cosmetic.

Nice test coverage overall — the real-`/bin/sh` round-trips and the fake-backend `copyBinary` assertion (no `cat >`) are exactly the right guards against regressing the hang.

internal/backend/coder/coder_copy.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), backend.CopyTimeout)
+	defer cancel()
+	scp := exec.CommandContext(ctx, "scp",
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+		"-o", "LogLevel=ERROR",
+		"-o", "ProxyCommand=coder ssh --stdio %h",
+		localPath, target+":"+remoteTmp,
```
BenjaminBenetti, line 132: 🔍 **CodeReview Agent**

`scp <target>:<remoteTmp>` resolves a relative `remoteTmp` against the SSH **login dir** (home), and the `mkdir`/`mv` execs above/below run via `coder ssh` which also starts in the login dir — not the workspace folder that the devcontainer/codespaces writes use. So a `fleet copy` with an empty or relative dest (`resolveCopyIntoDest` returns `finalDir="."`) would land under `$HOME` on coder while `reportedPath` still resolves it against `inst.WorkspaceDir` — a wrong location plus a path the reply doesn't actually point at. Consider anchoring a relative `remotePath` to an absolute path before this transfer so all backends agree. (Safe-fails to a clean `mv` error rather than a hang, but worth checking in the planned coder validation.)

internal/fleetlaunch/fleetlaunch.go
```diff
@@ -115,19 +115,28 @@ fi`, RemotePath, RemotePath)
...
-// copyBinary streams the running fleet binary into the container at
-// RemotePath. It unlinks the old file first rather than truncating it:
-// a process the caller has just signalled may still have the old
-// binary mmap'd, and overwriting a mapped executable surfaces as
-// ETXTBSY. Unlink-then-create allocates a fresh inode so the running
-// process keeps its old copy and the new file is written cleanly.
+// stagePath is the user-writable path the binary is transferred to before being
+// installed into the privileged RemotePath. /tmp is writable without sudo on
+// every backend, so CopyFile (which is itself sudo-free) can always land it here.
+const stagePath = "/tmp/.fleet-launch-stage"
```
BenjaminBenetti: 🔍 **CodeReview Agent**

`stagePath` is a fixed constant, unlike the other staging names introduced in this PR (`InlineWriteScript`'s `.fleet-inline.$$`, coder's per-call `remoteStageName`). Two concurrent `copyBinary` calls into the same instance could clobber each other's stage between `CopyFile` and the install `mv`. Unlikely in practice (low severity), but a `$$`/random suffix would keep it consistent with the rest of the change.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behavior-level QA of `fix(coder): stdin-EOF-safe file transfer` @ `36a7d22`. Because no coder workspace is available in this environment, testing focused on the **devcontainer backend** — exercising the *shared* seam this PR introduces (`Backend.CopyFile`, the base64-in-argv inline writer, `validateCopyName`, binary staging) and confirming the provisioning hang fix introduces **no regression** on the path real users hit most. All app execution ran inside an isolated nested docker-in-docker container; nothing ran in the host session.

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` (static, `go build ./cmd/fleet`) | ✅ Pass | Compiles clean, `--help` runs |
| Provisioning reaches `running` (no hang) | ✅ Pass | `fleet up` → running in ~17s, never stuck in `creating` |
| Binary staging → `/usr/bin/fleet` (CopyFile + `mv` install) | ✅ Pass | 46,255,899-byte binary staged & installed, byte-exact |
| `fleet.rc` inline write (base64-in-argv) | ✅ Pass | Correct path/mode 644, `FLEET_INSTANCE_NAME='alpha'` quoted & sourced, `.bashrc` wired in |
| `fleet exec` into instance | ✅ Pass | Runs as remoteUser `vscode` |
| `fleet copy` IN (single file) | ✅ Pass | Content + mode `755` preserved |
| `fleet copy` INTO a directory dest | ✅ Pass | Source basename kept |
| `fleet copy` OUT (single file) | ✅ Pass | Content + mode `750` preserved |
| Binary 1 MB round-trip (sha256, in & out) | ✅ Pass | Byte-identical both directions; exact size |
| Nasty quoted dest path (`ShellQuote`) | ✅ Pass | `"/tmp/has space's.bin"` copied intact |
| Copy OUT a missing file | ✅ Pass | Non-zero exit, no partial dest left behind |
| Control-char filename rejected (`validateCopyName`) | ✅ Pass | Dest name with embedded newline cleanly refused |
| Recursive directory copy in/out (devcontainer) | ✅ Pass | Still supported; nested tree preserved both ways |
| Copy BETWEEN instances | ✅ Pass | `alpha:→bravo:` 1 MB, sha256 matches |
| Stop/start sanity (state survives) | ✅ Pass | Files + `fleet.rc` survive; postStart hook re-runs (#179) |
| **Coder `scp`-over-`coder ssh` CopyFile** (root-cause fix) | ⏭️ Skipped | **No coder workspace available** — cannot exercise the actual fixed transport |
| Codespaces non-PTY CopyFile fix | ⏭️ Skipped | No codespaces environment |
| Agent (Claude/auggie) hook-script inline write | ⚠️ Partial | Base fixture has no agent tool, so not triggered; shares the same `WriteFileInline` helper proven by `fleet.rc` |
| Automation event-file write hang fix | ⏭️ Skipped | No trigger configured; shares the same `CopyFile` seam |

**Overall: ✅ PASS — 15 of 15 exercisable scenarios passed, 0 defects.** The coder-specific transport that is the literal subject of this PR could **not** be validated here (no coder workspace) and still needs a run on a real coder workspace, exactly as the PR description already flags under *Known limitations*.

---

## Detailed findings

**Environment:** isolated nested DinD container (Alpine 3.22 / Docker 28.5.2 / `@devcontainers/cli` 0.87.0 / node 22), fixture = `integration/fixture` (`mcr.microsoft.com/devcontainers/base:debian-12` + tmux), remoteUser `vscode`.

### ✅ Provisioning no longer hangs (the regression this guards)
`fleet up alpha --repo file:///root/itest-fleet` reached `running` in **17.4s** and `fleet list` showed `running`. The first provisioning step that previously could wedge on a stdin-EOF read (the `fleet.rc` write / binary staging) completed, so `markInstanceRunning` was reached. No `creating`-forever stall on the devcontainer path.

### ✅ The new shared file-transfer seam is correct
- **Binary staging** (`fleetlaunch` → `CopyFile` to `/tmp`, then stdin-free `mv` install): `/usr/bin/fleet` present, executable, `46,255,899` bytes — identical to the host build, confirming the staged binary is not truncated/corrupted.
- **Inline writer** (`WriteFileInline`, base64-in-argv): `/home/vscode/.fleet/fleet.rc` written mode `644`; the per-instance block emitted `export FLEET_INSTANCE_NAME='alpha'`; a login shell resolved `FLEET_INSTANCE_NAME=alpha`; `.bashrc` carries exactly one source line.
- **`fleet copy` (`CopyInto` buffer → `copyFileInto` → backend `CopyFile`)**: in/out/between all preserve content **and** mode; a **1 MB random blob** is sha256-identical after IN, OUT, and instance→instance hops.

### ✅ Edge cases the PR hardens
- **`ShellQuote`**: copying to `/tmp/has space's.bin` (space + single quote) landed intact — the atomic temp+rename script quotes the path safely.
- **`validateCopyName`**: a dest basename containing a newline was rejected with a clean error (no file created) — matches the control-char guard added for the coder scp framing.
- **Failed copy hygiene**: copying a non-existent source exits non-zero and leaves **no** destination file.
- **Recursive dir copy** remains functional on devcontainer (the PR only refuses it on coder/codespaces).

### ✅ No regressions in lifecycle
`fleet stop` → `stopped`, `fleet start` → `running`; a previously-copied `/tmp/up.bin` and the staged `fleet.rc` both survived the restart, and the devcontainer `postStartCommand` re-fired on start (#179 behavior intact).

### ⏭️ / ⚠️ Not exercised here (scope limits, not defects)
- **Coder `scp` CopyFile** — the actual root-cause fix — needs a real coder workspace; **untested** here. This is the one item I'd want validated before full confidence in #223's closure.
- **Codespaces non-PTY exec** fix — no codespaces env.
- **Agent hook-script inline writes** (Claude/auggie) — the debian-base fixture ships no agent tool, so the provisioners short-circuited; the underlying `WriteFileInline` they call is the same helper proven correct above.
- **Automation event-file write** — no trigger configured; routes through the same `CopyFile` seam validated above.

_Verdict label reflects the devcontainer-backend result; it is **not** a sign-off on the coder transport, which remains a tracked follow-up._

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the stdin-EOF-safe file-transfer rework. This is a solid, unusually well-documented change: the `Backend.CopyFile` strategy seam is a clean boundary, the base64-in-argv inline writer neatly sidesteps the coder stdin hang for small files, and the test coverage (real `/bin/sh` round-trips, size guard, shell-quoting, fake-backend staging, the coder dir-copy gate) is thorough. `go build`, `go vet`, and the changed-package tests (`backend/...`, `agentdetect/...`, `fleetlaunch/...`, `server/...`) all pass locally.

**Verified during review**
- `resolveCopyIntoDest` still probes the parent dir for existence (NotFound) before the write, so `CopyFile`'s `mkdir -p` does **not** silently create a missing destination directory — no behavior regression versus the old `tar -xf - -C <dir>`.
- The coder `mkdir`/`install`/`removeRemote` execs and the scp `ProxyCommand` all resolve to the same `workspace.agent` target, so the staged temp, the rename, and the cleanup all act on one consistent filesystem.
- `CopyIntoReply.Written` now reports `written`, which is validated to equal the declared size — equivalent to the previous `open.GetSize()`.
- base64's alphabet contains no single quote, so the single-quoted `printf %s '<b64>'` embedding is safe; the `maxInlineRaw` cap keeps the command well under ARG_MAX.

**Worth a look (none blocking)** — two inline notes below, plus:
- **Doc/name nit** (`internal/agentdetect/auggie_provisioner.go:75`): the comment reads `// settingsSuffix is the home-relative path…` but the constant is `auggieSettingsSuffix`. Stale identifier in the doc comment (the Claude provisioner's matching const is actually named `settingsSuffix`, so the comment looks copy-pasted).
- **The coder scp path remains the main untested risk** — already called out in the PR description. +1 on validating it against a real coder workspace before relying on it; in particular confirm whether the agent's scp negotiates SFTP vs the legacy protocol, since the quoting note below depends on it.

Inline comments follow.

internal/backend/coder/coder_copy.go
BenjaminBenetti, line 132: 🔍 **CodeReview Agent**

The scp **remote** path isn't shell-quoted here: `target+":"+remoteTmp`. The surrounding `mkdir`, `install` (`chmod`/`mv`), and `removeRemote` steps all wrap their paths in `backend.ShellQuote`, but this scp argument does not.

`remoteTmp` derives from `remotePath`, and for `fleet copy` the destination *directory* can legitimately contain spaces or shell metacharacters — it's only `test -d`-probed in `resolveCopyIntoDest`, not character-restricted (`validateCopyName` constrains the basename, not the dir). With OpenSSH's legacy SCP protocol (the pre-9.0 default) the remote path is expanded by a remote shell, so a destination like `/home/user/my docs/` would break the transfer or land the temp in the wrong place; the SFTP default (9.0+) takes it literally and is fine — so the behavior silently depends on the agent's scp version.

Quoting the remote portion (e.g. `target+":"+backend.ShellQuote(remoteTmp)`) makes it transport-version-independent and matches the quoting the adjacent steps already rely on.

internal/server/copy.go
```diff
@@ -353,76 +354,88 @@ func (s *service) CopyInto(stream grpc.ClientStreamingServer[fleetgrpc.CopyIntoC
...
-			}
-			return writeErr
+// drainCopyIntoToHostFile streams the client's data chunks into a host temp file,
+// enforcing the declared size exactly: a stream that ends short or overruns is an
+// InvalidArgument (so a truncated upload never lands as a silent partial file),
+// and any other failure is Internal. On success it returns the temp file
+// rewound to the start (ready to be read by CopyFile) and the byte count; the
+// caller owns closing and removing it.
+func drainCopyIntoToHostFile(stream copyIntoChunkReceiver, size int64) (*os.File, int64, error) {
+	f, err := os.CreateTemp("", "fleetcopyin-*")
```
BenjaminBenetti, line 409: 🔍 **CodeReview Agent**

`CopyInto` now buffers the **entire** upload to a host temp file before handing it to `CopyFile`, whereas the old `tar` path streamed straight through to the instance. For coder this temp is reused directly by scp (good — no second copy), but the buffering means a large `fleet copy` first writes the full payload to the server host's default temp dir.

If `$TMPDIR`/`/tmp` on the host is small or RAM-backed (tmpfs), a multi-GB copy can now exhaust it (or apply memory pressure) where the streaming path didn't. The exact-size validation does require a rewindable source, so buffering to disk is a reasonable trade — but it's a new host-side requirement. Consider one of: documenting it, honoring `$TMPDIR` placement explicitly, and/or enforcing a maximum accepted `size`. Low severity; flagging because it's a behavioral change for large transfers.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` binary | ✅ Pass | `go build ./cmd/fleet` clean (Go 1.25.9) |
| `fleet up` provisions → **running** (bug #223) | ✅ Pass | Reaches `running`, no hang in `creating`; exit 0 |
| Binary staging into instance (`CopyFile`→`mv`) | ✅ Pass | `/usr/bin/fleet` staged, exec user owns it |
| `fleet.rc` inline write + `fc`/`fl` aliases (`InlineWriteScript`) | ✅ Pass | `~/.fleet/fleet.rc` (1605B) with `alias fc=`/`fl=` |
| `fleet copy` upload single file (mode preserved) | ✅ Pass | content + `0755` intact in instance |
| `fleet copy` into directory dest (keeps basename) | ✅ Pass | lands as `dir/up.bin` |
| `fleet copy` 100 KB binary integrity | ✅ Pass | sha256 host == instance |
| `fleet copy` download out of instance | ✅ Pass | content intact |
| `fleet copy` recursive directory upload (devcontainer) | ✅ Pass | nested tree copied |
| `fleet copy` into non-existent dest dir | ✅ Pass | clean `NotFound`, **no hang** (timeout-guarded) |
| `fleet copy` control-char (newline) filename | ✅ Pass | rejected `InvalidArgument … contains a control character` |
| `fleet copy` missing local source | ✅ Pass | clean non-zero exit |
| Stop → Start lifecycle (re-stages binary) | ✅ Pass | back to `running`, no hang, artifacts persist |
| **Coder** backend scp transport (subject of #223) | ⏭️ Skipped | No coder workspace in this env — see below |
| **Codespaces** backend non-PTY copy | ⏭️ Skipped | No codespace in this env |

**Overall: ✅ PASS — 13 of 13 exercisable scenarios passing; 2 backend-specific paths not exercisable here (documented).**

---

### Environment
All execution isolated in a privileged Docker-in-Docker container (never the runner's own session):
`docker:27-dind` (Alpine 3.21.3, Docker 27.5.1) + Go 1.25.9 + `@devcontainers/cli` 0.87.0 + tmux/git/node. Fleet built from PR head `f492af1`, driven against the repo's `integration/fixture` devcontainer (`mcr.microsoft.com/devcontainers/base:debian-12`).

### Detailed findings

**Core fix — provisioning no longer hangs (bug #223).** `fleet up alpha --repo file://…` was driven exactly as a user would. The instance pulled its image, ran `postCreate`, and reached:
```
Creating itest-fleet/alpha (backend: devcontainer)...
Instance itest-fleet/alpha is running (container: 6e260564bcb5)
EXIT_CODE=0
```
It reached **`running`** rather than wedging in `creating`. The provisioning steps the PR reworked all completed: the `fleet` binary was staged to `/usr/bin/fleet` (the `CopyFile`→`/tmp`→`mv` path) and `~/.fleet/fleet.rc` was written via the base64-in-argv `InlineWriteScript` (verified its `alias fc='fleet copy'` / `alias fl='fleet launch'` content in-container).

**`fleet copy` (rewritten `CopyInto`).** Every direction/edge behaved correctly:
- Single-file upload preserved content and `0755` mode; directory dest kept the source basename.
- A 100 KB random binary round-tripped with matching sha256 (exercises the buffered-to-host-temp + exact-size validation + `CopyFile` stream).
- Download out of the instance and recursive directory upload both succeeded.
- **No-hang error handling** (each guarded with `timeout`): upload into a missing instance dir → `NotFound` and exits; missing local source → clean non-zero. Neither blocks.
- Control-char guard: a filename containing a newline is rejected with `InvalidArgument … "ba\nd.txt" contains a control character`.

**Lifecycle sanity.** `fleet stop` → `stopped`, `fleet start` → `running` (no hang on the re-stage), and post-restart the staged binary and a previously uploaded file were still intact. `fleet ls` rendered correctly throughout.

### Not exercisable in this environment (no defect observed — flagging coverage gaps)
- **Coder backend** is the actual subject of #223 (the `scp`-over-`coder ssh --stdio` transport that replaces the stdin-EOF `cat`). It **cannot be exercised here** without a real coder workspace — the same limitation the PR description calls out ("Needs validation on a real coder workspace"). Its correctness rests on the PR's unit tests + code review, not this manual pass. Recommend a smoke test of `fleet up` + `fleet copy` against a live coder workspace before relying on it in production.
- **Codespaces** non-PTY copy path likewise not exercised (no codespace available).

The devcontainer and codespaces backends share the new `Backend.CopyFile`/`InlineWriteScript` seam that I did exercise end-to-end on devcontainer, so the common refactor is validated; only the coder-specific transport remains unverified by manual QA.

BenjaminBenetti:
Thanks for the QA + review pass. Addressed the CodeReview findings in `bf15bd8`:

- **`fleetlaunch.go:121` — fixed `stagePath` could clobber on concurrent stages** → `stageBinaryPath()` now returns a per-call-unique `/tmp/.fleet-launch-stage.<rand>` path, matching the `$$`/random uniqueness used by the inline writer and `remoteStageName`.

- **`copy.go:401` — host-temp buffering for large copies** → documented. The buffer is what makes the exact-size validation possible before anything is written into the instance (and lets coder scp it with no second copy). It lives in `os.TempDir()` so it honors `$TMPDIR`; noted that a large `fleet copy` needs scratch space there.

- **`coder_copy.go:88` — scp remote path not shell-quoted** → this one is intentional, now documented inline. scp's **default SFTP** transport (OpenSSH 9.0+) takes the post-`:` path **literally**, so wrapping it in `ShellQuote` would make the quotes part of the filename. Quoting is only correct for the pre-9.0 legacy protocol (remote-shell expansion), which this code never opts into (no `-O`). The adjacent `mkdir`/`mv`/`rm` execs *do* run through `sh -c`, so those stay quoted. Staging-path callers always use `/tmp` (no spaces) regardless.

- **`coder_copy.go:88` — relative dest lands in `$HOME` on coder** → real edge, now documented on `CopyFile`: `remotePath` should be absolute. Every #223 staging caller passes an absolute path (binary → `/tmp`, automation event → `/tmp`), so the hang fix is unaffected; only `fleet copy` with an empty/relative dest is affected, and that's part of the coder `fleet copy` support still flagged for validation on a real coder workspace.

The QA Agent's devcontainer PASS (15/15) covers the shared seam; the coder scp transport remains the one item needing a real coder workspace, as called out under Known limitations.

BenjaminBenetti:
Second review pass addressed in `b020738`.

**`coder_copy.go` — scp remote operand quoting (re-raise).** You're right that the earlier doc-only response didn't actually close this: staging the scp temp as a *sibling* of `remotePath` let a spacey `fleet copy` dest dir flow into the one argument that can't be quoted. And the literal suggestion (`ShellQuote(remoteTmp)`) can't be the fix either — it's only correct for the legacy pre-9.0 protocol; under the SFTP default the quotes become part of the filename and the follow-up `mv` then can't find the staged file. There's no single quoting choice that's right under both transports.

So I removed the need to quote it at all: the scp temp now stages into a **fixed `/tmp` path** (`remoteStageName()` → `/tmp/.fleet-scp.<seq>.<rand>`, characters restricted to `[0-9a-f.]`), which is safe verbatim under both SFTP and legacy scp. The spacey final path is reached only by the already-`ShellQuote`'d `mv`, never by scp. Tradeoff documented on `CopyFile`: that `mv` may now cross filesystems (copy+unlink, not an atomic rename) when `remotePath` is outside `/tmp` — acceptable for this best-effort staging. Added a metacharacter assertion to `TestRemoteStageNameUnique`.

**`copy.go:409` — host-temp buffering for large copies (re-raise).** Keeping as-is, documented. The temp already lives in `os.TempDir()`, so `$TMPDIR` placement *is* honored explicitly — the doc comment calls that out and advises pointing `$TMPDIR` at a roomy filesystem when the default `/tmp` is small or RAM-backed. I'm deliberately not adding a hard `size` cap: the old streaming path had none, any fixed limit is arbitrary and would reject legitimate large transfers, and the exact-size validation (the reason for buffering) is what makes a truncated upload fail loudly instead of landing as a silent partial file. The buffering is the price of that guarantee.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the stdin-EOF-safe file-transfer rework (the `Backend.CopyFile` strategy seam plus the inline/stream writers). This is a solid, carefully documented change: the abstraction is clean, the per-backend transport choices are each justified in comments, and the hardening — exact-size guard, atomic temp-then-rename, control-char rejection, per-exec timeouts — is the right set of things to add. Test coverage is genuinely thorough (real-`/bin/sh` round-trips, fake-backend staging proof, codespaces/coder gate tests). I cloned the branch and confirmed `go build ./...`, `go vet ./internal/...`, and the affected package tests all pass.

A few things worth a look — all **non-blocking**, leaving as comments only:

**1. (Medium) The coder `scp` transfer's 5-minute timeout may not actually be enforced.** Every other coder exec in `coder_copy.go` runs through `CombinedOutputWithTimeout` (which sets `WaitDelay = 3s`), but the `scp` call uses `exec.CommandContext(ctx, …)` + `scp.CombinedOutput()` with no `WaitDelay`/`Cancel`. `scp` spawns `ssh`, which spawns the `coder ssh --stdio` ProxyCommand; those children inherit `scp`'s combined-output pipe write-end. On `ctx` timeout Go kills only `scp` (`Process.Kill`), and with `WaitDelay == 0` it will *not* force-close the inherited pipe — so `CombinedOutput()` (`cmd.Wait`) blocks on the output-copy goroutine until the lingering `ssh`/`coder ssh` children exit. That's the same "wedge provisioning forever" failure mode this PR is eliminating, just softer. Details inline.

**2. (Low) Orphaned staging file on install failure.** In `copyBinary`, if `CopyFile` succeeds but the subsequent install `mv` exec fails (non-writable `/usr/bin` + `sudo -n` denied is the realistic case), the staged `/tmp/.fleet-launch-stage.*` is left behind in the instance. The coder path cleans up its own `remoteTmp` on failure, but nothing removes this one, so repeated provisions accumulate orphans. Inline.

**3. (Low) `stageBinaryPath` dropped the degraded-RNG guard.** `remoteStageName` (and the former `copyIntoTempName`) fold an `atomic.Uint64` into the name so an ignored `crypto/rand.Read` failure can't collapse two names to the same string; `stageBinaryPath` relies on rand alone. Practically harmless, just inconsistent with its sibling helpers. Inline.

**Notes (no action needed):** single-file `fleet copy` now buffers the whole upload to host `$TMPDIR` before transfer — a deliberate, already-documented tradeoff; and the coder arbitrary-dest `fleet copy` path control-char-validates only the basename, not `finalDir`. Both are fine given the coder `fleet copy` path is a tracked follow-up — just worth remembering when it's enabled.

Nice work on the root-cause writeup and the adversarial-review pass.

internal/backend/coder/coder_copy.go
```diff
@@ -0,0 +1,155 @@
...
+	// code never opts into via -O.) remoteTmp derives from remotePath, so the
+	// staging-path callers (always /tmp) carry no spaces regardless.
+	scp := exec.CommandContext(ctx, "scp",
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+		"-o", "LogLevel=ERROR",
+		"-o", "ProxyCommand=coder ssh --stdio %h",
+		localPath, target+":"+remoteTmp,
+	)
+	if out, err := scp.CombinedOutput(); err != nil {
```
BenjaminBenetti, line 142: 🔍 **CodeReview Agent**

This is the one exec in this file that doesn't go through `CombinedOutputWithTimeout`. `exec.CommandContext` will `Process.Kill()` **only `scp`** on `ctx` timeout, and because `WaitDelay` is unset (0) Go won't force-close the stdout/stderr pipe that `scp`'s `ssh` / `coder ssh --stdio` children inherited. So `CombinedOutput()` → `cmd.Wait()` blocks on the output-copy goroutine until those children close the pipe — a stuck tunnel can keep this call hanging well past `CopyTimeout`, which is the exact failure class this PR is removing.

Setting `WaitDelay` makes the deadline real, mirroring `CombinedOutputWithTimeout`'s own `WaitDelay = 3s`:

```go
scp := exec.CommandContext(ctx, "scp",
    ...
    localPath, target+":"+remoteTmp,
)
scp.WaitDelay = 10 * time.Second
```

internal/fleetlaunch/fleetlaunch.go
```diff
@@ -139,18 +156,21 @@ func copyBinary(instanceBackend backend.Backend, workspaceDir string) error {
 	}
 	defer bin.Close()
 
-	script := fmt.Sprintf(`
-write='rm -f %s && cat > %s && chmod +x %s'
-if [ -w %s ]; then
+	stagePath := stageBinaryPath()
+	if err := instanceBackend.CopyFile(workspaceDir, bin, stagePath, 0o755); err != nil {
```
BenjaminBenetti, line 166: 🔍 **CodeReview Agent**

If this `CopyFile` succeeds but the install exec below fails (the realistic case being a non-writable `/usr/bin` with `sudo -n` denied), `stagePath` is left behind in the instance's `/tmp`. The coder backend cleans up its internal `remoteTmp` on failure, but nothing removes this staged copy — so repeated provision attempts accumulate `/tmp/.fleet-launch-stage.*` orphans.

Consider a best-effort `rm -f %s` of `stagePath` on the install error path (via `ExecCommand` + `CombinedOutputWithTimeout`), matching the cleanup `coder.CopyFile`/`removeRemote` already does.

internal/fleetlaunch/fleetlaunch.go
```diff
@@ -115,19 +116,35 @@ fi`, RemotePath, RemotePath)
...
-// process keeps its old copy and the new file is written cleanly.
+// stageBinaryPath returns a per-call-unique, user-writable path the binary is
+// transferred to before being installed into the privileged RemotePath. /tmp is
+// writable without sudo on every backend, so CopyFile (which is itself sudo-free)
+// can always land it here. The random suffix keeps two concurrent stagers from
+// clobbering each other's staging file between CopyFile and the install mv,
+// matching the per-call uniqueness used elsewhere in the staging path.
+func stageBinaryPath() string {
+	var b [6]byte
+	_, _ = rand.Read(b[:])
```
BenjaminBenetti, line 133: 🔍 **CodeReview Agent**

Minor / consistency: `remoteStageName` (coder) and the former `copyIntoTempName` deliberately fold an `atomic.Uint64` into the name so that an *ignored* `crypto/rand.Read` error (which leaves `b` zeroed) can't produce two identical names. `stageBinaryPath` relies on the random bytes alone, so under a degraded RNG two concurrent stagers could collide. The collision window is vanishingly small, but adding the same seq counter here keeps the staging helpers uniform.

BenjaminBenetti:
Third review pass — all three addressed in `cbda93f`. Two were real, one a worthwhile consistency fix.

**`coder_copy.go` — scp not bounded by `WaitDelay` (real, good catch).** This is the same hang class the PR removes, and the scp call was the one exec in the file not going through `CombinedOutputWithTimeout`. On the ctx deadline `CommandContext` kills only `scp`; its inherited `coder ssh --stdio` children keep the stdout/stderr pipe open, so `cmd.Wait()` (inside `CombinedOutput`) would block on the output-copy goroutine until they close it — a stuck tunnel hanging past `CopyTimeout`. Set `scp.WaitDelay = 10s` (generous vs. `backend.Cmd`'s own 3s, since the nested-ssh teardown is slower than a local exec) so the pipe is force-closed shortly after the kill and `Wait()` returns. Documented why.

**`fleetlaunch.go:160` — orphaned stage temp on install failure (real).** Right — and the randomized staging name made it worse: every retried provision leaked a *fresh* `/tmp/.fleet-launch-stage.*`. Added a best-effort `rm -f` (ShellQuote'd) of `stagePath` on the install error path, mirroring `coder.removeRemote`. Covered by a new `TestCopyBinaryCleansStageOnInstallFailure` (a `failExec` fake-backend mode that fails the install exec and asserts the cleanup `rm` runs against the staged path).

**`fleetlaunch.go:127` — seq counter for RNG-failure uniqueness (consistency).** Done. Folded an `atomic.Uint64` into `stageBinaryPath` so an ignored `crypto/rand.Read` failure (zeroed buffer) still can't collide, uniform with `remoteStageName`. Vanishing window, but it's free and keeps the two staging helpers identical in spirit.

Local: build / vet / gofmt clean, backend + fleetlaunch + server tests green. CI running on `cbda93f`.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-review of head `b020738` (this `synchronize` event). This exact commit already received a [detailed pass](https://github.com/BenjaminBenetti/fleet-man/pull/225#pullrequestreview-4588379964) at 00:50 UTC, so I won't repeat those findings. On a fresh independent read I **confirm no new blocking issues**: the `Backend.CopyFile` seam, the inline (`base64`-in-argv) and stdin-stream writers, the exact-size guard + atomic temp/rename, the codespaces non-PTY fix, and the control-char filename guard all look correct, and I re-verified the affected helpers (`CombinedOutputWithTimeout`, dest resolution, negative-size rejection, quoting/`base64` safety) line up.

Consolidated status of the still-open items (all already acknowledged, **none blocking**):

- **(medium) The coder `scp` 5‑min timeout may not actually be enforced.** `exec.CommandContext(ctx, "scp", …)` + `scp.CombinedOutput()` sets no `WaitDelay`/`Cancel`; on `ctx` timeout Go kills only `scp`, and the inherited combined-output pipe held open by the `ssh` / `coder ssh --stdio` ProxyCommand children can keep `cmd.Wait()` blocked — a softer version of the very "wedge forever" mode this PR removes. This is the one item I'd most want closed before relying on the coder path. (Matches the prior pass; flagging as still-present in `b020738`.)
- **(low) Orphaned `/tmp/.fleet-launch-stage.*`** in `copyBinary` if `CopyFile` succeeds but the install `mv` exec fails (non-writable `/usr/bin` + `sudo -n` denied) — the coder path cleans its own `remoteTmp`, but this stage temp is left behind.
- **(low) `stageBinaryPath` uses `crypto/rand` alone**, unlike `remoteStageName`/the former `copyIntoTempName`, which fold in an `atomic.Uint64` so an ignored `rand.Read` failure can't collapse two names.
- **(note) Single-file `fleet copy` now buffers the whole upload to host `$TMPDIR`** before transfer, capped only by the client-declared size — deliberate and documented; just worth remembering for host-scratch usage on very large copies and as an unbounded-input surface once coder `fleet copy` is enabled.
- **(follow-ups) Coder `fleet copy`** remains the tracked, still-unvalidated surface: relative/empty dest landing under `$HOME` with an untruthful reported path, directory copy (cleanly refused for now), no cancellation context, and the cross-filesystem `mv` from `/tmp` being a non-atomic copy+unlink.

Net: solid, well-documented change in good shape to merge. The main ask stays validating the coder scp path — including the scp-timeout enforcement above — against a real coder workspace.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build (`go build ./...`) | ✅ Pass | Clean compile of all packages in an isolated container |
| CLI sanity (`fleet --help`, subcommands) | ✅ Pass | All subcommands render; `copy`/`up` help correct; no panic |
| Instance provisioning (`fleet up`) — the bug | ✅ Pass | Reached **running** in ~16s, **no hang** in `creating` (container `5e0d8a3fbc71`) |
| Provisioning artifacts | ✅ Pass | `~/.fleet/fleet.rc` inline-written (1611B, 0644); `/usr/bin/fleet` staged & executable; `~/.bashrc` sources the rc |
| `fleet copy` IN — text file | ✅ Pass | Content + mode (`0640`) preserved |
| `fleet copy` IN — binary integrity | ✅ Pass | 200 KB random (NUL/0xFF) sha256 matches end-to-end |
| `fleet copy` IN — directory dest keeps name | ✅ Pass | `…:/tmp/` landed as `/tmp/bin.dat` |
| `fleet copy` OUT — download | ✅ Pass | 150 KB sha256 matches |
| Control-char filename rejected | ✅ Pass | `InvalidArgument … contains a control character` |
| `fleet copy` IN — directory (devcontainer) | ✅ Pass | Nested tree (`a.txt`, `sub/b.txt`) extracted correctly |
| Non-existent dest dir | ✅ Pass | Clean `NotFound` error, no hang |
| Stop / start lifecycle | ✅ Pass | State + copied files persist across restart |
| **Coder backend (real bug target)** | ⏭️ Skipped | No coder workspace available — PR-acknowledged limitation |

**Overall: ✅ PASS — 12 of 12 testable scenarios pass; 1 skipped (coder, not testable here).**

---

## Method

All build/run/interaction was performed inside an **isolated Docker container** (`qa-225`), never in the runner session. The container carried its own toolchain (Go 1.25, Node 20, devcontainer CLI 0.87.0, tmux, git) and drove the real `fleet` binary built from the PR head (`b020738`). A devcontainer instance was provisioned as a sibling container against the repo's `integration/fixture` (`mcr.microsoft.com/devcontainers/base:debian-12`), exactly mirroring `integration/common.sh`. The container is torn down after the run.

> ⚠️ The coder backend — where issue #223 actually manifested — could not be exercised (no coder workspace). The devcontainer path validates the **shared** `CopyInto` → `Backend.CopyFile` → `StreamWriteScript`/`InlineWriteScript` seam this PR introduces; the coder-specific `scp`-over-`coder ssh --stdio` transport and the dir-copy `FailedPrecondition` refusal remain validated only by unit tests, consistent with the PR's "Known limitations".

## Detailed findings

### Provisioning no longer hangs (root fix) ✅
`fleet up qa-instance --repo file://…` returned exit 0 with `Instance itest-fleet/qa-instance is running` in ~16s (`00:59:57` → `01:00:13`). Before this fix the instance would sit in `creating` forever because the first stdin-to-EOF provisioning step blocked. Reaching `running` proves `markInstanceRunning` was reached, i.e. the `fleet.rc` write + binary staging completed. Confirmed inside the instance:
- `~/.fleet/fleet.rc` — present, 1611 bytes, mode `0644` (the `InlineWriteScript` base64-in-argv path).
- `/usr/bin/fleet` — present, executable, 46 MB (the `CopyFile` → `/tmp` → `mv` staging path).
- `~/.bashrc` line 160: `[ -f ~/.fleet/fleet.rc ] && . ~/.fleet/fleet.rc` (hook wiring).

### `fleet copy` (the `CopyInto`/`CopyFile` seam) ✅
- Upload text → content and mode `0640` preserved.
- Upload 200 KB of `/dev/urandom` → `sha256` identical in-instance (`649eb5fe…`), proving the devcontainer `cat > tmp`-over-stdin transport **terminates** (gets EOF) and is **binary-safe**.
- Directory-style dest (`:/tmp/`) keeps the source basename.
- Download (copy OUT) of a 150 KB file → `sha256` identical (`af3d8ec1…`).
- Recursive directory copy into the devcontainer extracts the nested tree correctly.

### Defensive behavior ✅
- A source basename containing a newline is rejected: `code = InvalidArgument desc = file name "ev\nil.txt" contains a control character` (new `validateCopyName`).
- A copy to a non-existent dest directory fails fast with `code = NotFound desc = destination directory "/no/such/dir" does not exist` — a clean error, not a hang.

### Lifecycle regression sanity ✅
`fleet stop` → `stopped`, `fleet start` → `running` (postStart re-ran, `outcome: success`); the previously copied `/tmp/hello.txt` and `/tmp/bin.dat` survived intact.

### Note (benign, not a regression)
`fleet version` prints an empty line in these runs. This is identical for a binary built from `main` and is just the dev build lacking ldflags version injection (set at release/`install.sh`); unrelated to this PR.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the latest commit (`cbda93f`) on top of the full diff. I cloned the branch and ran `go build ./...`, `go vet`, and the affected suites (`backend`, `fleetlaunch`, `agentdetect`, `server`) locally — **all green**. The core design is sound: a single `Backend.CopyFile` strategy seam with per-transport implementations is the right abstraction, the stdin-EOF reasoning is carefully documented, and the new tests (binary round-trips, awkward-quoted paths, size guard, control-char filenames, fake-backend staging) are genuinely good coverage. All three backends implement the new interface method, so the seam is complete.

A few non-blocking observations, roughly in priority order:

**1. (Low–Medium) The 64 KB inline cap converts a previously-unbounded settings write into a hard failure.**
`writeSettings` (claude + auggie provisioners) now routes the *merged user settings.json* through `InlineWriteScript`, which returns an error once `content > maxInlineRaw` (64 KB). The old `cat >`-over-stdin write had no size limit. Because the existing on-disk `settings.json` is read, merged, and rewritten, a user whose `~/.claude/settings.json` (or `~/.augment/settings.json`) is already large would have hook (re)provisioning fail. It's non-fatal — `provision.go` / `runtime_poller.go` log it as a warning and continue — so the effect is "state-detection hooks silently don't get installed," not a crash. The embedded *hook scripts* are fixed-size constants so they're safe; only the merge path is exposed. Consider routing the settings write through the streaming `CopyFile` seam (no cap) rather than the inline writer, or at minimum surfacing the cap reason in the warning. 64 KB is unlikely in practice, but it's a real behavior change from "always worked" to "fails above N". See inline note.

**2. (Low) Failed single-file `fleet copy` can leave an orphaned temp on devcontainer/codespaces.**
The removed `CopyInto` path explicitly called `removeContainerFile` to clean its extraction temp on *any* error. The replacement `StreamWriteScript`/`InlineWriteScript` writers rely on `set -e` and have no cleanup, so a failure between the `cat > "$t"` / `printf | base64 -d > "$t"` and the final `mv` leaves a hidden `.fleet-copy.$$` / `.fleet-inline.$$` orphan in the destination directory. Notably the *coder* `CopyFile` (`removeRemote`) and the fleetlaunch stager both do clean up their temps, so this is an inconsistency rather than a systemic gap. A `trap 'rm -f "$t"' EXIT` in the two shared writers would restore parity. See inline note.

**3. (Nit / already self-flagged) coder `scp` ProxyCommand `%h` substitution is unquoted** and `mv` from `/tmp` to a cross-filesystem dest is a non-atomic copy+unlink for `fleet copy`. Both are acknowledged in the code comments and gated behind the "needs validation on a real coder workspace" caveat, so I'm only noting them for completeness — no action needed for the #223 fix itself (binary + automation staging always target `/tmp`).

Nice hardening touches: rejecting control characters in copy filenames (protects the legacy scp record framing), the `WaitDelay` backstop so a wedged tunnel can't outlive `CopyTimeout`, the exact-size validation before anything touches the instance, and the per-call-unique staging names guarding against `crypto/rand` degradation. Overall this looks like a solid, well-tested fix for the hang.

internal/agentdetect/claude_provisioner.go
```diff
@@ -95,62 +102,62 @@ func (p *ClaudeProvisioner) resolveScriptPath() (string, error) {
...
-	const script = `mkdir -p "$HOME/.claude" && ` +
-		`cat > "$HOME/.claude/settings.json.tmp" && ` +
-		`mv "$HOME/.claude/settings.json.tmp" "$HOME/.claude/settings.json"`
-	_, err := p.exec.Run([]string{"sh", "-c", script}, content)
+// writeSettings writes content to settingsPath atomically (same-directory tmp
+// then rename, handled by the inline writer, which also mkdir's the parent for
+// the first-time case). Inline (base64-in-argv) rather than stdin-streamed so it
+// cannot hang on the coder backend (issue #223).
+func (p *ClaudeProvisioner) writeSettings(settingsPath string, content []byte) error {
+	argv, err := backend.InlineWriteScript(settingsPath, content, 0o644)
```
BenjaminBenetti: 🔍 **CodeReview Agent**

The merged user `settings.json` is written via `InlineWriteScript`, which hard-errors above the 64 KB `maxInlineRaw` cap. The previous `cat >`-over-stdin write was unbounded, so a user with an already-large existing `settings.json` would now have hook injection fail here (non-fatal — logged as a warning in `provision.go`/`runtime_poller.go` — but the state-detection hooks silently won't install). The embedded hook *scripts* are fixed-size constants and safe; only this read-merge-rewrite path is exposed. Consider routing the settings write through the streaming `CopyFile` seam (no cap) instead of the inline writer. Same applies to `auggie_provisioner.go:139`.

internal/backend/copy_file.go
```diff
@@ -0,0 +1,104 @@
...
+// reads file content from STDIN and writes it to remotePath with the given mode.
+// The write is atomic (same-dir temp + rename) and creates the parent directory.
+//
+// This is the stdin-streamed counterpart used by CopyFile on backends whose
+// transport delivers stdin EOF. It must NOT be used on the coder backend, whose
+// `cat` would block forever waiting for an EOF that never arrives.
+func StreamWriteScript(remotePath string, mode int) []string {
+	qPath := ShellQuote(remotePath)
+	body := fmt.Sprintf(
+		`set -e; d=$(dirname %[1]s); mkdir -p "$d"; t="$d/.fleet-copy.$$"; cat > "$t"; chmod %[2]o "$t"; mv -f "$t" %[1]s`,
```
BenjaminBenetti: 🔍 **CodeReview Agent**

`StreamWriteScript` (and `InlineWriteScript` just above) relies on `set -e` with no cleanup, so a failure between `cat > "$t"` / `base64 -d > "$t"` and the final `mv` leaves an orphaned `.fleet-copy.$$` / `.fleet-inline.$$` in the destination directory. The old `CopyInto` path explicitly cleaned its temp via `removeContainerFile` on any error, and the coder `CopyFile` (`removeRemote`) + the fleetlaunch stager both clean up theirs — so devcontainer/codespaces single-file copies are the inconsistent case. A `trap 'rm -f "$t"' EXIT` after `t=...` in both writers would restore parity. Low severity.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual QA of `fix/coder-copyfile-stdin-eof` @ `cbda93f`. Since no coder/codespaces server is reachable in this environment, I exercised the **devcontainer** transport end-to-end (the regression surface) by building the PR binary and driving real `fleet` flows against a live devcontainer instance, inside an isolated dedicated Docker container.

| Test | Result | Notes |
|------|--------|-------|
| Build gate (`go build ./...` + `fleet` binary) | ✅ Pass | Builds clean; binary runs |
| Create devcontainer instance reaches `running` | ✅ Pass | **#223 regression** — provisioning completed, never wedged in `creating` (container `94026b6…`) |
| fleet-launch binary staging (`CopyFile`→`StreamWriteScript`) | ✅ Pass | 46 MB binary staged to `/usr/bin/fleet`, **md5 byte-exact** vs host build; terminated |
| `fleet.rc` inline write (`InlineWriteScript`, base64-in-argv) | ✅ Pass | Lands at `~/.fleet/fleet.rc` (1605 B), correct content/owner |
| Claude + auggie hook scripts + `settings.json` (inline) | ✅ Pass | Scripts present & `0755` (mode preserved); `settings.json` valid JSON wired to hooks |
| `fleet copy` upload — text + ~1 MB odd-size binary | ✅ Pass | md5 matches in-instance; exact-size drain works |
| `fleet copy` download — binary round-trip | ✅ Pass | Downloaded md5 == original; no hang |
| Directory copy into instance (devcontainer) | ✅ Pass | Nested tree extracted with correct contents |
| Control-character filename rejected | ✅ Pass | `InvalidArgument: … "ba\ndname" contains a control character` |
| Path-traversal filename rejected | ✅ Pass | `InvalidArgument: invalid file name ".."` |
| Sanity: `fleet exec` / `status` / bad-instance error | ✅ Pass | Exec runs as `vscode`; bad target → clean `NotFound`, no hang |
| Coder `scp` transport | ⏭️ Skipped | No coder workspace available — matches the PR's own "needs validation on a real coder workspace" note |
| Codespaces non-PTY `CopyFile` | ⏭️ Skipped | No codespace available |

**Overall: ✅ PASS — 11/11 testable scenarios pass; 2 transports not exercisable in this environment (no coder/codespaces server).**

## Detailed findings

**Environment.** Built the PR's `fleet` binary, then ran the fleet client+server and a real devcontainer instance inside a dedicated throwaway container (`qa-225`) that shares an isolated inner dockerd. The instance under test is itself a separate devcontainer (`mcr.microsoft.com/devcontainers/base:debian`).

**Headline regression (#223) — provisioning no longer hangs.** `fleet up` of a devcontainer instance completed to `running`. This path runs exactly what hung on coder: the `fleet.rc` write and the binary copy. On devcontainer (local docker exec delivers stdin EOF) both finished and `markInstanceRunning` was reached.
```
Creating testrepo/inst1 (backend: devcontainer)...
Instance testrepo/inst1 is running (container: 94026b6578e0)
```

**Binary staging integrity (`CopyFile` → `StreamWriteScript`).** The 46 MB fleet-launch binary copied into the instance is byte-identical to the host build:
```
in-instance:  bf43339669ea7da4982c213136de140f  /usr/bin/fleet
host build:   bf43339669ea7da4982c213136de140f
```

**Inline writers (`InlineWriteScript`).** `~/.fleet/fleet.rc`, `~/.fleet/scripts/{claude,auggie}-state-hook.sh` (both `0755`), and `~/.claude/settings.json` (valid JSON, hook commands pointing at the scripts) were all written correctly via the base64-in-argv path — no stdin involved.

**`fleet copy` (`CopyInto` → host-temp + size validation → `CopyFile`).**
```
Copied ./up.txt -> /tmp/up.txt (46 bytes)
Copied ./up.bin -> /tmp/up.bin (1000003 bytes)            # ~1MB, odd size
Copied testrepo/inst1:/tmp/up.bin -> ./down.bin (1000003 bytes)
up.bin / in-instance / down.bin md5 all == 80a4c504a3b80bd361826cac6078ab48
```
Upload, download, and round-trip are all md5-exact and terminate promptly.

**Guards.** Control-character (`\n`) and `..` destination basenames are rejected with `InvalidArgument`; a copy targeting a non-existent instance returns a clean `NotFound` (no wedge). Directory copy into a devcontainer instance extracts the full tree.

**Not exercised (no defect):** the coder `scp`-over-`coder ssh --stdio` path and the codespaces non-PTY exec path — no such servers are reachable here. These remain validated only by the PR's unit tests + review, consistent with the author's stated follow-up to validate coder on a real workspace.

BenjaminBenetti:
Fourth review pass — both addressed in `9cdee96`. Both real; the first is a genuine regression I'd missed.

**`claude_provisioner.go` / `auggie_provisioner.go` — settings.json write hit the inline 64 KB cap (regression).** Good catch. The old `cat >`-over-stdin write was unbounded, so routing the merged settings.json through `InlineWriteScript` regressed the >64 KB case to a hard failure on devcontainer/codespaces (non-fatal but the hooks silently don't install). And settings.json is precisely the *unbounded* payload here — a read-merge-rewrite of arbitrary user content — i.e. exactly what the new `CopyFile` seam was built for. So I widened `ContainerExecutor` with a `CopyFile` method (`BackendExecutor` delegates to `backend.CopyFile`) and routed **both** provisioners' `writeSettings` through it: uncapped *and* stdin-EOF-safe on every backend including coder. The fixed-size hook *scripts* stay on the inline writer (the safe, hang-proof path for small constants). Test fakes updated to record the streamed write (`fakeExec.copies`; the integration `localExecutor` writes to disk so the e2e settings assertions still exercise a real file). Applies to both `:157` (claude) and `:139` (auggie).

**`copy_file.go:100` — orphaned writer temp on mid-write failure.** Fixed in both `InlineWriteScript` and `StreamWriteScript`: added `trap 'rm -f "$t"' EXIT` right after `t=…`, so a `set -e` abort between the write and the `mv` cleans the `.fleet-inline.$$` / `.fleet-copy.$$` temp instead of leaving it behind — restoring parity with the coder `CopyFile` (`removeRemote`) and the fleetlaunch stager. On success the temp is already renamed away, so the trap's `rm` is a harmless no-op.

Local: build / vet / gofmt clean; `agentdetect` (incl. the host e2e provisioning tests), `backend`, `create`, `server` all green. CI running on `9cdee96`.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the latest push (through `9cdee96`). This is a careful, well-documented fix: the `Backend.CopyFile` seam cleanly removes the stdin-EOF hang, the inline-vs-stream split is well-reasoned, and the host-side helpers have strong tests (real `/bin/sh` round-trips, binary bytes, size guards, quoting). The four prior review rounds already closed the obvious gaps (per-call uniqueness, orphan cleanup, scp `WaitDelay`, control-char filenames), and I confirmed the scp tunnel and the follow-up `mv` both resolve to the same `coder ssh` target, so they share one filesystem namespace. Two remaining observations:

**1. The `Backend.CopyFile` atomicity guarantee isn't upheld by the coder backend (medium-low).**
The interface doc states *"The write is atomic (same-directory temp + rename)"*, and the inline/stream writers honor that (temp + rename in the **destination** dir). The coder implementation does not: it scps to a fixed `/tmp/.fleet-scp.*` and then `mv -f`s into `remotePath`. When `remotePath` is outside `/tmp` — which is every non-#223 caller: `~/.claude` / `~/.augment/settings.json`, automation event files, and `fleet copy` destinations — `/tmp` is frequently a separate filesystem (tmpfs) from `$HOME` (overlay), so that `mv` degrades to a non-atomic copy+unlink. During that window a concurrent reader can see ENOENT or a partially-written file — exactly what the contract is meant to prevent. coder_copy.go documents the cross-fs `mv` locally, but the *interface* still advertises atomicity, so a caller trusting the contract is misled.

Two ways to resolve:
- Soften the interface doc to spell out the coder cross-fs caveat; **or**
- Restore real atomicity without reopening the scp-quoting problem you deliberately fixed: keep scping to the metacharacter-free `/tmp/.fleet-scp.*` (the unquoted operand stays safe), but have the **install** exec publish through a *same-directory* temp — `cp` the staged file to a `ShellQuote`'d sibling of `remotePath`, `chmod`, then `mv` (a true intra-dir rename). The install step is a normal quoted exec, so the spacey/destination path is no obstacle there, and you get the atomic rename back.

**2. Whole-upload host buffering in `drainCopyIntoToHostFile` (low / by-design).**
Single-file `fleet copy` now buffers the entire upload into host `os.TempDir()` before transfer (it previously streamed straight through). This is needed for the exact-size check and is reasonable, but two consequences are worth surfacing more loudly than a function comment: (a) a large copy now needs full-size host scratch and doubles I/O, and (b) it's a new host-disk-pressure surface — the buffered size is bounded only by the client's declared size and available disk. Consider mentioning the `$TMPDIR` requirement somewhere user-facing (the `fleet copy` help/docs), since a small or RAM-backed `/tmp` on the server host will now cap copy size.

Neither is blocking — the layering and the inline rationale comments are genuinely good. Nice work.

internal/backend/backend.go
```diff
@@ -92,6 +95,25 @@ type Backend interface {
 	// otherwise identical to ExecCommand.
 	ExecCommandQuiet(workspaceDir string, command []string) *Cmd
 
+	// CopyFile streams src INTO the instance, writing it to remotePath with the
+	// given mode. It is the single strategy seam for host→instance file transfer:
+	// callers (the fleet-launch binary stager, the `fleet copy` server handler)
+	// invoke it without caring which transport a backend uses. The write is
+	// atomic (same-directory temp + rename), creates remotePath's parent, and
```
BenjaminBenetti, line 102: 🔍 **CodeReview Agent**

This promises *"atomic (same-directory temp + rename)"*, but the coder `CopyFile` stages in `/tmp` and `mv -f`s into `remotePath`. When `remotePath` is outside `/tmp` (settings.json, automation files, most `fleet copy` dests), `/tmp` is often a different filesystem, so the `mv` becomes a non-atomic copy+unlink and a concurrent reader can observe ENOENT or a partial file — contradicting this guarantee.

Either soften this sentence to note the coder cross-fs caveat, or have coder's **install** exec publish through a `ShellQuote`'d *same-directory* temp (`cp` the `/tmp` stage → sibling of `remotePath` → `mv`) so the rename is atomic again. The scp operand stays the metacharacter-free `/tmp` path, so the quoting tradeoff you fixed is unaffected. (See the main review for detail.)

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behavior-level QA of the stdin-EOF-safe file-transfer change, driven through the real `fleet` binary (built from `9cdee96`) inside an isolated Docker container that shares this instance's nested Docker daemon. Backend exercised: **devcontainer** (the only backend provisionable here). The fixture instance was created with `fleet up alpha --repo file://…/itest-fleet`, then driven with `fleet copy` / `exec` / `stop` / `start`.

| Test | Result | Notes |
|------|--------|-------|
| Build & launch (`fleet --help`, `fleet ls`) | ✅ Pass | Binary runs, empty state renders |
| **Provisioning completes — no hang** (`fleet up`) | ✅ Pass | Reaches `running`, exit 0; the #223 symptom (stuck in `creating`) does not reproduce on devcontainer |
| `fleet.rc` inline-written | ✅ Pass | `~/.fleet/fleet.rc` present, `FLEET_INSTANCE_NAME='alpha'` exported, `.bashrc` sources it |
| Fleet binary staged (CopyFile→/tmp→mv) | ✅ Pass | `/usr/bin/fleet` present, runnable, **sha256 byte-exact** vs host (46 MB) |
| Copy OUT single file + mode | ✅ Pass | Content + `755` preserved |
| Copy OUT into directory | ✅ Pass | Keeps source basename |
| Copy OUT relative path | ✅ Pass | Resolves against workspace folder (see note 1) |
| Copy OUT missing file | ✅ Pass | Non-zero exit, no dest file left behind |
| Copy IN single file + mode (new buffer→CopyFile) | ✅ Pass | Content + `755` preserved |
| Copy IN into directory | ✅ Pass | Keeps source basename |
| `host:` prefix up/down | ✅ Pass | Both directions correct |
| Copy IN into non-existent dir | ✅ Pass | Clean non-zero exit |
| Copy IN missing local file | ✅ Pass | Clean non-zero exit |
| Instance→instance relay | ✅ Pass | Content preserved |
| Directory tree IN / OUT / relay | ✅ Pass | Nested files intact all three directions |
| **Binary integrity round-trip** (100 KB, NUL bytes, crosses 64 KB chunk) | ✅ Pass | sha256 byte-exact IN **and** OUT |
| Nasty quoted remote path (spaces + `'`) | ✅ Pass | `ShellQuote` handles `/tmp/odd dir/we ird'name.bin` |
| Control-char filename rejected | ✅ Pass | Newline-in-name upload refused (`InvalidArgument`) |
| Lifecycle stop/start + re-stage | ✅ Pass | `stopped`→`running`, staged binary intact after restart |
| Coder scp `CopyFile` path | ⏭️ Skipped | No coder workspace available — the PR's primary fix could not be exercised directly (author flagged this as needing real-coder validation) |
| Codespaces non-PTY binary path | ⏭️ Skipped | No codespace available |
| Coder dir-copy clean refusal / size-validation reject | ⏭️ Skipped | Coder path needs a real workspace; exact-size truncation isn't triggerable from the CLI (client sends the true size) |

**Overall: ✅ PASS — 19 of 19 executed scenarios pass; 3 coder/codespaces-only paths could not be exercised in this environment.**

---

## Detailed findings

### Core fix — provisioning no longer hangs (devcontainer regression surface)
`fleet up alpha --repo file://…/itest-fleet` completed in well under the 6-minute hard timeout, printed `Instance itest-fleet/alpha is running (container: ccfdda004459)` and exited `0`. The refactored provisioning (`EnsureFleetRC` via `WriteFileInline`, binary staging via `CopyFile`→`/tmp/.fleet-launch-stage.*`→`mv`) all completed:
- `~/.fleet/fleet.rc` (1605 B) written, with the `export FLEET_INSTANCE_NAME='alpha'` per-instance block, and `~/.bashrc` contains `[ -f ~/.fleet/fleet.rc ] && . ~/.fleet/fleet.rc`.
- `/usr/bin/fleet` staged, executable, and **`sha256 = b49696fe…3039` identical to the host source binary** — the 46 MB streamed-then-installed payload is byte-exact, confirming the `StreamWriteScript` `cat > tmp` transport terminates correctly on devcontainer.

### `fleet copy` matrix (22 assertions)
Ran every direction the command supports — OUT, IN, `host:` prefix, instance↔instance relay, and recursive directory copy — asserting content **and** mode. The two integrity-focused cases are the most relevant to this PR:
- **L. Binary round-trip:** a 100 007-byte payload (random + explicit `\x00\x01\x02\xff\xfe\x7f\x80`, crossing the 64 KB chunk boundary) copied IN and back OUT with sha256 unchanged both ways — the new buffer-to-host-temp + `CopyFile` IN path preserves arbitrary bytes.
- **M. Nasty path:** uploading into `/tmp/odd dir/we ird'name.bin` (spaces + a single quote) landed correctly — `ShellQuote` is sound.
- **N. Control-char rejection:** a local file whose name contained a newline was refused with a non-zero exit, matching `validateCopyName`.

Error cases (missing source OUT, missing local file IN, missing/non-existent dest dir IN) all exit non-zero and leave no partial destination.

### Notes
1. **Copy OUT relative path (initial transient):** my first attempt failed only because the cloned workspace under `/workspaces/itest-fleet` is owned by `root:root` (the QA harness runs `fleet` as root, so `git clone` produced root-owned files), and the container's `vscode` user got `Permission denied` creating the seed file — i.e. the test fixture couldn't be written, not a copy bug. After seeding `ws-rel.txt` as root, `fleet copy itest-fleet/alpha:ws-rel.txt …` resolved it against the workspace folder and copied `relative-bytes` correctly. Counted as Pass. This ownership quirk is an artifact of running fleet as root in the sandbox, not a product issue.

### Scope / limitations of this run
- The change's **headline target is the coder backend**, which cannot be provisioned in this environment (no coder workspace). The coder-specific `scp`-over-`coder ssh --stdio` `CopyFile`, the coder directory-copy `FailedPrecondition` refusal, and the codespaces non-PTY exec path were **not exercised** — consistent with the PR's own "Known limitations" note that the coder path "needs validation on a real coder workspace."
- What this run *does* establish: the shared `Backend.CopyFile` / `InlineWriteScript` / `StreamWriteScript` seam and the rewired `fleetlaunch` binary staging, `EnsureFleetRC`, and `server.CopyInto` callers all work correctly on the devcontainer backend, with byte-exact transfers and intact error handling — so the refactor introduced no regression on the backend that was already working.

_Tested at commit `9cdee96`. Container & fixture instance torn down after the run._

PR #226, opened 2026-07-03
feat(tui): tabs in panes — tmux windows as a per-pane tab bar

### LINKED ISSUE #168: Fleet TUI - Tabs in panes
Bit of a moon shot idea, not sure exactly if this is possible but, 

Imagine that if in any given pane within a session, you could hit a key (tmux new window is 'ctrl + b then c') and it would spawn a new "tab" the tabs would be visualized across the top of the pane, as mouse intractable links. The user would also be able to cycle tabs using a keyboard shortcut. 

In this way we can increase the information density of `fleet` even more. Letting users get more out of a single session, instead of requiring flipping between sessions.

#### Mockup
```
--------------------------------------------------------------------------------------
[tab 1] [tab 2] [tab 3] 

normal terminal content here.




--------------------------------------------------------------------------------------

normal terminal content here of another pane that has only
one tab 




--------------------------------------------------------------------------------------
```

There is no obvious way to pull this off popping in to my head, so Fable 5 label, need to wait for the un ban 🤞 

(may attempt with Opus, if I can first understand a clear path my self, for now just recording the idea).

BenjaminBenetti:
Closes #168.

## What

Per-pane tabs in the fleet TUI, exactly as mocked in the issue: a clickable tab strip across the top of each session pane, keyboard shortcuts to create/cycle tabs, independent per pane.

## How — the moon shot turned out to be a config change

Each TUI pane is a host-tmux split pane attached (via `fleet shell`) to a full tmux **session** inside the container. Tmux **windows** inside that session are the tabs — and the inner tmux status line, which fleet previously hid in split mode (`set -g status off`), is the tab bar. No terminal emulation, no new rendering, no data-plane round trips:

- **Tab bar** = inner status line at `status-position top`, styled to the TUI palette (`styles.go` greys + colour39 accent), one entry per window.
- **Click-to-switch** = tmux's default `MouseDown1Status` binding; mouse passthrough already works (it's how drag-copy inside panes works today).
- **Keys** = tmux defaults, now visible and documented: `ctrl+x c` new tab, `ctrl+x n`/`p` cycle, `ctrl+x 0-9` jump (split mode; `ctrl+b` prefix full-screen). These technically "worked" before — into an invisible window with zero indication. The keybindings overlay lists them now.
- **Auto-hide** (split mode): `window-linked`/`window-unlinked` hooks hide the bar while a session has a single window, so the common one-tab case keeps full pane height — matching the issue mockup. The hooks are global on the container's shared tmux server, so a session-scoped `@fleet_tab_autohide` marker keeps them from hiding the bar on full-screen attaches, where it stays always-on and carries the existing detach hint.
- **Switching is instant** — window select/create is local to the container's tmux server; no ~2-3s exec round trip, nothing per-keystroke.
- Tabs persist across detach/reattach/TUI restarts for free (they live in the inner session).
- Session rows in the fleet list now show an `(N tabs)` badge sourced from the already-polled `#{session_windows}` count (previously plumbed but unused).

## Verified

Ran the exact emitted attach scripts (both modes) in an isolated nested two-server tmux harness (outer host layer + inner layer, isolated sockets):

- split mode: bar hidden at 1 window → appears at top with `0:bash 1:vim` + `ctrl+x c: new tab` hint on second window → hides again after kill
- full-screen: bar always on with session name + detach hint
- mode transition (same session attached split then full-screen): autohide flips, `status-left` restored
- `MouseDown1Status select-window -t=` confirmed present in default bindings (actual clicking not injectable headlessly)
- `go test ./...` green, `golangci-lint` 0 issues

## Known limitations (documented, deliberate v1 scope)

- Screen capture, agent-activity detection, and MCP `fleet_session_exec`/`read` target a session's *active* window — an agent running in a background tab is invisible to them until you switch back. Extending the capture/poll scripts per-window is a natural follow-up if this bites.
- Layout presets snapshot outer pane geometry only; a session's tab set isn't captured.
- *Concurrent* mixed-mode attaches to the same session (a TUI split pane and a full-screen `fleet shell` at once) get last-attach-wins semantics for the session-scoped prefix / auto-hide / status options — tmux options have no per-client scope. Sequential transitions are handled; concurrent same-mode attaches (multi-TUI) are consistent by construction. If it ever bites, `client-attached`/`client-detached` hooks re-running the mode chain are the escape hatch.

### DIFF
--- a/internal/tui/dotfiles.go
+++ b/internal/tui/dotfiles.go
@@ -86,19 +86,48 @@ func ShellCommandForSession(config *configutil.Config, session string, cols, row
 		tmuxSize = fmt.Sprintf(` -x %d -y %d`, cols, rows-1)
 		resizeHook = ` \; set -g aggressive-resize on \; set -g window-size latest`
 	}
+	// Tabs (issue #168): tmux windows inside the inner session act as
+	// per-pane tabs. The status line is the tab bar, rendered across the
+	// top of the pane with one entry per window; tmux's default
+	// MouseDown1Status binding makes each entry click-to-switch, and the
+	// default prefix keys (c / n / p / 0-9) create and cycle tabs.
+	//
+	// tabToggle hides the bar while a session has a single window and
+	// shows it otherwise. It runs once at attach time and again from the
+	// window-linked/unlinked hooks on every window create/kill. The hooks
+	// are global on the container's shared tmux server, so they are
+	// guarded by the session-scoped @fleet_tab_autohide marker: nested
+	// panes auto-hide (the TUI provides all other UI) while full-screen
+	// attaches keep the bar always on because status-right also carries
+	// the detach hint there.
+	tabToggle := `if -F '#{&&:#{@fleet_tab_autohide},#{==:#{session_windows},1}}' 'set status off' 'set status on'`
+	tabBarConf := ` \; set -g status-position top` +
+		` \; set -g status-style 'bg=default,fg=colour245'` +
+		` \; set -g window-status-format ' #I:#W '` +
+		` \; set -g window-status-current-format '#[fg=colour39,bold] #I:#W #[default]'` +
+		` \; set-hook -g window-linked "` + tabToggle + `"` +
+		` \; set-hook -g window-unlinked "` + tabToggle + `"`
 	// When nested inside a host tmux (split pane mode), use Ctrl+X as
 	// the inner prefix so it doesn't conflict with the outer Ctrl+B.
 	// Pane navigation (h/j/k/l) is handled by the outer tmux, so the
 	// inner tmux only needs prefix and session keys.
-	prefixConf := ""
+	modeConf := ""
 	statusRight := ` ctrl+q/ctrl+o: detach `
 	if nested {
 		// In nested mode, Ctrl+Q/O are handled by the outer tmux
 		// (they close all split panes). The inner tmux only needs
 		// the prefix override and session navigation. The status bar
-		// is hidden because the outer tmux provides all the UI.
-		prefixConf = ` \; set -g prefix C-x \; bind-key C-x send-prefix \; set -g status off`
+		// doubles as the tab bar and is hidden while the session has a
+		// single window (the outer tmux provides all other UI).
+		modeConf = ` \; set -g prefix C-x \; bind-key C-x send-prefix` +
+			` \; set status-left '' \; set status-right ' ctrl+x c: new tab '` +
+			` \; set @fleet_tab_autohide 1 \; ` + tabToggle
 		statusRight = ""
+	} else {
+		// Full-screen mode keeps the bar always on: it carries the
+		// detach hint and doubles as the tab bar. -u resets a
+		// status-left cleared by an earlier nested attach.
+		modeConf = ` \; set @fleet_tab_autohide 0 \; set status on \; set -u status-left`
 	}
 	// Session navigation: Ctrl+PageUp/Down are handled by the outer
 	// tmux to cycle session groups. prefix+T creates a new session
@@ -152,7 +181,7 @@ func ShellCommandForSession(config *configutil.Config, session string, cols, row
 		` \; unbind -n MouseDown2Pane`
 	clipboardFeatures := ` \; set -as terminal-features ',*:clipboard'`
 	inner := setup + tmuxEnsureInstalled + sizefix + sshAgentFix + hookClear + fmt.Sprintf(
-		`exec tmux -u new-session -A -s %s`+tmuxSize+` \; set -g mouse on`+clipboardConf+mouseBindings+detachKeys+statusConf+prefixConf+sessionKeys+resizeHook+clipboardFeatures,
+		`exec tmux -u new-session -A -s %s`+tmuxSize+` \; set -g mouse on`+clipboardConf+mouseBindings+detachKeys+tabBarConf+statusConf+modeConf+sessionKeys+resizeHook+clipboardFeatures,
 		shQuote(session),
 	)
 	return []string{"sh", "-c", inner}
--- a/internal/tui/dotfiles.go
+++ b/internal/tui/dotfiles.go
@@ -95,18 +95,21 @@ func ShellCommandForSession(config *configutil.Config, session string, cols, row
 	// tabToggle hides the bar while a session has a single window and
 	// shows it otherwise. It runs once at attach time and again from the
 	// window-linked/unlinked hooks on every window create/kill. The hooks
-	// are global on the container's shared tmux server, so they are
-	// guarded by the session-scoped @fleet_tab_autohide marker: nested
-	// panes auto-hide (the TUI provides all other UI) while full-screen
+	// and the status options are session-scoped so they never touch a
+	// user's own tmux sessions on the container's shared server (or
+	// clobber their global hooks); only the window-status formats are
+	// global, as window options have no session scope. The session-scoped
+	// @fleet_tab_autohide marker distinguishes the modes: nested panes
+	// auto-hide (the TUI provides all other UI) while full-screen
 	// attaches keep the bar always on because status-right also carries
 	// the detach hint there.
 	tabToggle := `if -F '#{&&:#{@fleet_tab_autohide},#{==:#{session_windows},1}}' 'set status off' 'set status on'`
-	tabBarConf := ` \; set -g status-position top` +
-		` \; set -g status-style 'bg=default,fg=colour245'` +
+	tabBarConf := ` \; set status-position top` +
+		` \; set status-style 'bg=default,fg=colour245'` +
 		` \; set -g window-status-format ' #I:#W '` +
 		` \; set -g window-status-current-format '#[fg=colour39,bold] #I:#W #[default]'` +
-		` \; set-hook -g window-linked "` + tabToggle + `"` +
-		` \; set-hook -g window-unlinked "` + tabToggle + `"`
+		` \; set-hook window-linked "` + tabToggle + `"` +
+		` \; set-hook window-unlinked "` + tabToggle + `"`
 	// When nested inside a host tmux (split pane mode), use Ctrl+X as
 	// the inner prefix so it doesn't conflict with the outer Ctrl+B.
 	// Pane navigation (h/j/k/l) is handled by the outer tmux, so the
--- a/internal/tui/dotfiles.go
+++ b/internal/tui/dotfiles.go
@@ -119,18 +119,24 @@ func ShellCommandForSession(config *configutil.Config, session string, cols, row
 	if nested {
 		// In nested mode, Ctrl+Q/O are handled by the outer tmux
 		// (they close all split panes). The inner tmux only needs
-		// the prefix override and session navigation. The status bar
+		// the prefix override and session navigation. The prefix is
+		// session-scoped so full-screen attaches and a user's own
+		// sessions on the shared server keep the default Ctrl+B
+		// (bind-key tables are inherently global; prefix+C-x =
+		// send-prefix is inert under other prefixes). The status bar
 		// doubles as the tab bar and is hidden while the session has a
 		// single window (the outer tmux provides all other UI).
-		modeConf = ` \; set -g prefix C-x \; bind-key C-x send-prefix` +
+		modeConf = ` \; set prefix C-x \; bind-key C-x send-prefix` +
 			` \; set status-left '' \; set status-right ' ctrl+x c: new tab '` +
 			` \; set @fleet_tab_autohide 1 \; ` + tabToggle
 		statusRight = ""
 	} else {
 		// Full-screen mode keeps the bar always on: it carries the
-		// detach hint and doubles as the tab bar. -u resets a
-		// status-left cleared by an earlier nested attach.
-		modeConf = ` \; set @fleet_tab_autohide 0 \; set status on \; set -u status-left`
+		// detach hint and doubles as the tab bar. -u resets the
+		// prefix and status-left a prior nested attach overrode, so
+		// the documented full-screen keys (ctrl+b …) hold regardless
+		// of attach order.
+		modeConf = ` \; set @fleet_tab_autohide 0 \; set status on \; set -u status-left \; set -u prefix`
 	}
 	// Session navigation: Ctrl+PageUp/Down are handled by the outer
 	// tmux to cycle session groups. prefix+T creates a new session
--- a/internal/tui/dotfiles_test.go
+++ b/internal/tui/dotfiles_test.go
@@ -244,10 +244,10 @@ func TestShellCommandTabBar(t *testing.T) {
 	for _, nested := range []bool{true, false} {
 		script := shellCommand(config, "agent-1", 80, 24, nested)[2]
 		for _, want := range []string{
-			"set -g status-position top",
+			"set status-position top",
 			"set -g window-status-format ' #I:#W '",
-			`set-hook -g window-linked "` + toggle + `"`,
-			`set-hook -g window-unlinked "` + toggle + `"`,
+			`set-hook window-linked "` + toggle + `"`,
+			`set-hook window-unlinked "` + toggle + `"`,
 		} {
 			if !strings.Contains(script, want) {
 				t.Errorf("nested=%v script missing %q: %s", nested, want, script)
--- a/internal/tui/dotfiles_test.go
+++ b/internal/tui/dotfiles_test.go
@@ -231,9 +231,53 @@ func TestShellCommandNestedNoInnerPaneKeys(t *testing.T) {
 	if !strings.Contains(script, "set -g prefix C-x") {
 		t.Errorf("nested script missing prefix override: %s", script)
 	}
-	// Status bar should be hidden in nested mode.
-	if !strings.Contains(script, "set -g status off") {
-		t.Errorf("nested script should hide status bar: %s", script)
+	// The status bar doubles as the tab bar and auto-hides while the
+	// session has a single window.
+	if !strings.Contains(script, "set @fleet_tab_autohide 1") {
+		t.Errorf("nested script missing tab auto-hide marker: %s", script)
+	}
+}
+
+func TestShellCommandTabBar(t *testing.T) {
+	config := state.DefaultConfig()
+	toggle := `if -F '#{&&:#{@fleet_tab_autohide},#{==:#{session_windows},1}}' 'set status off' 'set status on'`
+	for _, nested := range []bool{true, false} {
+		script := shellCommand(config, "agent-1", 80, 24, nested)[2]
+		for _, want := range []string{
+			"set -g status-position top",
+			"set -g window-status-format ' #I:#W '",
+			`set-hook -g window-linked "` + toggle + `"`,
+			`set-hook -g window-unlinked "` + toggle + `"`,
+		} {
+			if !strings.Contains(script, want) {
+				t.Errorf("nested=%v script missing %q: %s", nested, want, script)
+			}
+		}
+	}
+}
+
+func TestShellCommandTabBarNested(t *testing.T) {
+	config := state.DefaultConfig()
+	script := shellCommand(config, "agent-1", 80, 24, true)[2]
+	// Auto-hide on, and the initial toggle runs at attach so a reattach
+	// to a multi-window session shows the bar immediately.
+	if !strings.Contains(script, `set @fleet_tab_autohide 1 \; if -F`) {
+		t.Errorf("nested script missing attach-time tab toggle: %s", script)
+	}
+	if !strings.Contains(script, "ctrl+x c: new tab") {
+		t.Errorf("nested script missing new-tab hint: %s", script)
+	}
+}
+
+func TestShellCommandTabBarFullScreen(t *testing.T) {
+	config := state.DefaultConfig()
+	script := shellCommand(config, "agent-1", 80, 24, false)[2]
+	// Full-screen keeps the bar always on (it carries the detach hint).
+	if !strings.Contains(script, `set @fleet_tab_autohide 0 \; set status on`) {
+		t.Errorf("full-screen script should force status on: %s", script)
+	}
+	if !strings.Contains(script, "ctrl+q/ctrl+o: detach") {
+		t.Errorf("full-screen script missing detach hint: %s", script)
 	}
 }
 
--- a/internal/tui/dotfiles_test.go
+++ b/internal/tui/dotfiles_test.go
@@ -227,9 +227,11 @@ func TestShellCommandNestedNoInnerPaneKeys(t *testing.T) {
 	if strings.Contains(script, "bind-key k select-pane") {
 		t.Errorf("nested script should not have k keybinding (pane nav is on outer tmux): %s", script)
 	}
-	// Should still have the prefix override for inner tmux.
-	if !strings.Contains(script, "set -g prefix C-x") {
-		t.Errorf("nested script missing prefix override: %s", script)
+	// Should still have the prefix override for inner tmux,
+	// session-scoped so it doesn't leak to full-screen attaches or a
+	// user's own sessions on the shared server.
+	if !strings.Contains(script, `set prefix C-x`) || strings.Contains(script, "set -g prefix") {
+		t.Errorf("nested script should set session-scoped prefix: %s", script)
 	}
 	// The status bar doubles as the tab bar and auto-hides while the
 	// session has a single window.
@@ -279,6 +281,11 @@ func TestShellCommandTabBarFullScreen(t *testing.T) {
 	if !strings.Contains(script, "ctrl+q/ctrl+o: detach") {
 		t.Errorf("full-screen script missing detach hint: %s", script)
 	}
+	// A prior nested attach session-scopes prefix C-x; full-screen must
+	// reset it so the documented ctrl+b keys hold regardless of order.
+	if !strings.Contains(script, `set -u prefix`) {
+		t.Errorf("full-screen script should reset session prefix: %s", script)
+	}
 }
 
 func TestShellCommandNestedVimKeysDisabled(t *testing.T) {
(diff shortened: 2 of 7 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**PR #226 @ d096b07** — reviewed `internal/tui/dotfiles.go`, the row/view badge plumbing, keybindings, and tests.

**Overall**: solid approach. Reusing the inner tmux status line as the tab bar keeps this a config-only change, and the emitted-script tests pin the important invariants in both modes plus the nested↔full-screen transition. One incidental fix worth calling out: the old nested path's `set -g status off` was **global** on the shared server, so a single nested attach used to hide the detach-hint bar for every full-screen attach too — the new session-scoped toggle fixes that.

Findings (no blockers):

1. **Global blast radius on the shared tmux server** — `set -g status-position top`, the styling globals, and the global `window-linked`/`window-unlinked` hooks leak into sessions fleet doesn't own on the container's default socket. Details and containment options in the inline comment on `dotfiles.go`.
2. **`prefix + T` sessions never get the marker/toggle** — `bind-key T new-session` switches the client to a session that never runs the attach chain, so in a nested pane it shows tmux's *default* status content (session name + time/hostname), always on, at the top via the new global `status-position`. Before this PR the global `status off` hid it. Probably acceptable v1 behavior, but worth a look — e.g. `bind-key T new-session \; set @fleet_tab_autohide 1 \; <toggle>` would bring those sessions in line.
3. **tmux version floor (minor)** — `set-hook`, `window-linked`/`window-unlinked`, and `#{&&:}` formats need tmux ≥ 2.4. Because the whole attach is a single command sequence, a parse error on an ancient preinstalled tmux aborts the entire attach (unlike `terminal-features`, which this file deliberately appends last for exactly that reason). Almost certainly academic given `tmuxEnsureInstalled`, but worth confirming it enforces a floor rather than mere presence.
4. **Documented limitation → automation impact** — the active-window-only scope of capture/agent-activity/MCP exec is documented, but note it can *silently* affect automation: an agent finishing (or hanging) in a background tab won't flip activity detection until someone switches back to that tab. Suggest a tracking issue so the follow-up doesn't get lost.

Checked and found sound: `tabToggle` quoting through `sh -c` → tmux command-string parsing; `#{&&:}` false-on-`0`/empty semantics for the marker; full-screen re-attach after a nested attach correctly restores `status-right` (via the session-scoped `statusConf`) and `status-left` (via `set -u`); hook ordering within the attach chain; `tabCount` only set for single-session rows (multi-pane groups keep the panes suffix) with a test covering both; `Windows` sourced from the already-polled runtime discovery as described.

internal/tui/dotfiles.go
```diff
@@ -86,19 +86,48 @@ func ShellCommandForSession(config *configutil.Config, session string, cols, row
...
+	// tabToggle hides the bar while a session has a single window and
+	// shows it otherwise. It runs once at attach time and again from the
+	// window-linked/unlinked hooks on every window create/kill. The hooks
+	// are global on the container's shared tmux server, so they are
+	// guarded by the session-scoped @fleet_tab_autohide marker: nested
+	// panes auto-hide (the TUI provides all other UI) while full-screen
+	// attaches keep the bar always on because status-right also carries
+	// the detach hint there.
+	tabToggle := `if -F '#{&&:#{@fleet_tab_autohide},#{==:#{session_windows},1}}' 'set status off' 'set status on'`
+	tabBarConf := ` \; set -g status-position top` +
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Everything in `tabBarConf` is server-global on the container's shared default-socket tmux server, so it leaks into sessions fleet doesn't own (a user's personal `tmux` inside the container): their bar jumps to the top, their window list gets restyled, and — via the global hooks — the marker-unset else-branch runs `set status on` in *their* session on every window create/kill, stomping a deliberate `status off`. The global `set-hook -g` also silently replaces any user-defined `window-linked`/`window-unlinked` hook.

Two containment options, composable:

- `status-position` and `status-style` are **session** options — dropping `-g` (setting them in the attach chain like `statusConf` already does) scopes them to fleet-attached sessions. (`window-status-format`/`window-status-current-format` are window options, so `-g` is the pragmatic choice there; their values are near-default anyway.)
- `set-hook` **without** `-g` sets a session-scoped hook: it still fires on window create/kill in fleet sessions whether or not a client is attached, never runs in foreign sessions, and doesn't clobber user hooks. The `@fleet_tab_autohide` marker then only has to distinguish nested vs full-screen, which it already does.

Not a blocker — this file already sets globals (`mouse on`, `aggressive-resize`, …) — but these are the first that visibly restyle non-fleet sessions rather than just tweaking input behavior.

BenjaminBenetti: Fixed in 773d0eb — `status-position`/`status-style` and both hooks are session-scoped now; only the window-status formats stay global (window options have no session scope), per your note. Re-verified in the isolated two-server harness, including a foreign session on the same server with a deliberate `status off`: window ops in it no longer flip its status, and it picks up no styling or hooks. Fleet-session auto-hide behavior is unchanged.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**PR #226 @ 773d0eb** — incremental review of the new commit (`fix(tui): session-scope the tab-bar status options and hooks`) since my previous review @ d096b07.

**Finding #1 (global blast radius on the shared tmux server): resolved.** `status-position`/`status-style` dropped `-g` — both are session options, applied in the attach chain against the fleet session — and both `window-linked`/`window-unlinked` hooks are now session-scoped `set-hook`, so they fire for fleet-session window ops (attached or not) without running in foreign sessions or clobbering user-defined global hooks. Window-status formats stay global as discussed in the inline thread (window options have no session scope; values near default) — accepted residual. The updated test assertions genuinely pin the scoping: `strings.Contains` on `"set status-position top"` and `` `set-hook window-linked "` `` cannot match the former `-g` variants, so a scope regression fails the tests.

Two small notes on this commit, neither blocking:

1. **prefix+T delta**: with the styling globals gone, a `prefix+T` session in a nested pane now shows the completely stock tmux bar (bottom, default theme) rather than the fleet-styled top bar it inherited at d096b07. Same root cause as finding #2 last time (those sessions never run the attach chain) — just noting the manifestation changed; the previously suggested `bind-key T new-session \; set @fleet_tab_autohide 1 \; <toggle>` would cover it if you want tab parity there.
2. **Stale globals from the prior build**: a long-lived tmux server that ran the d096b07 script keeps its global hooks/styling until restart — the new script doesn't unset them. Academic (d096b07 was never released); only relevant if testing both builds against the same container.

Prior findings #3 (tmux ≥ 2.4 floor — unchanged; session-scoped `set-hook` and the `window-linked`/`unlinked` hooks are the same 2.4+ vintage) and #4 (tracking issue for background-tab automation blindness) remain open and non-blocking.

No new issues found in this commit. LGTM from my side.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build fleet binary from PR branch | ✅ Pass | `go build ./cmd/fleet` clean in QA image |
| Provision instance (`fleet up` → devcontainer) | ✅ Pass | itest fixture instance running, no regressions |
| TUI launch & fleet list renders | ✅ Pass | No errors on startup |
| Keybindings overlay lists tab shortcuts | ✅ Pass | `prefix + c`, `n / p`, `0-9` all shown under "Tmux (inner session)" |
| Split attach, 1 window → no tab bar | ✅ Pass | Full pane height kept, matches issue mockup |
| `ctrl+x c` creates tab, bar appears at top | ✅ Pass | ` 0:bash  1:bash ` + `ctrl+x c: new tab` hint |
| `ctrl+x n` / `p` cycle, `ctrl+x 0` jump | ✅ Pass | Active window flips, per-tab content fully isolated |
| Active-tab styling | ✅ Pass | Bold colour39 vs grey colour245 — matches TUI palette |
| Click-to-switch tabs | ⚠️ Partial | `MouseDown1Status → select-window -t=` binding + `mouse on` verified in inner tmux; real click not injectable headlessly |
| Kill tab → bar auto-hides | ✅ Pass | Bar gone immediately at 1 window (split mode) |
| `(N tabs)` badge in fleet list | ✅ Pass | `● 19179f (2 tabs)` at 2 windows; badge clears at 1 |
| Badge suppressed on multi-pane groups | ✅ Pass | Group shows `(2 panes)`, no tab badge despite root having 2 windows |
| Tabs persist across detach/reattach | ✅ Pass | Bar shows immediately on split reattach (attach-time toggle) |
| Full-screen attach (`fleet shell`, no tmux) | ✅ Pass | Bar always on: session name + tabs + `ctrl+q/ctrl+o: detach` |
| Full-screen at 1 window keeps bar | ✅ Pass | Bar stays on after killing 2nd window (autohide off) |
| Mode transition split ↔ full-screen | ✅ Pass | `status-left` restored full-screen (`-u` fix); autohide re-engages back in split |
| `ctrl+q` detach | ✅ Pass | Clean exit, session persists |
| TUI quit sanity | ✅ Pass | `q` exits cleanly |

**Overall: ✅ PASS — 17 of 18 scenarios passing, 1 partial (environmental limitation only, no defect)**

## Environment

QA image `ghcr.io/benjaminbenetti/fleet-man/qa:latest` (privileged, DinD), PR head `d096b07`. Built `fleet` from the branch, provisioned a real fixture instance via `fleet up alpha --repo file://…/itest-fleet` (debian-12 devcontainer, tmux 3.3a inside), and drove the TUI/CLI in headless tmux exactly as a user would — keystrokes in, `capture-pane` out. No unit/integration suites were run.

## Detailed findings

### Split-pane tabs (core feature) — works as mocked in #168
1. `Enter` on the alpha row opened the split pane attached to inner session `alpha~19179f` — **top row was shell content, no tab bar** with a single window.
2. `ctrl+x c` → tab bar appeared instantly across the top of the pane: ` 0:bash   1:bash … ctrl+x c: new tab `. Confirmed 2 windows server-side (`tmux list-windows`).
3. Typed distinct markers into each tab; `ctrl+x p` / `ctrl+x n` / `ctrl+x 0` each switched the active window (verified via `#{window_index}` and by the marker content actually swapping — no cross-tab bleed). Switching was instant, no exec round-trip lag.
4. Raw capture shows the current tab rendered `ESC[1m ESC[38;5;39m` (bold, colour39) against colour245 greys — consistent with `styles.go` accent.
5. `exit` in one tab → back to 1 window → **bar auto-hid immediately**, pane reclaimed the row.

### Fleet list badge
- With 2 windows the session row rendered `● 19179f (2 tabs)`; the badge disappeared after the tab was killed (discovery poll picked both up within seconds).
- Added a group pane (`fleet shell itest-fleet/alpha --group 19179f`) while the root session had 2 windows: row switched to `● 19179f (2 panes)` with **no tab badge** — matches the intended single-session-row-only behavior.

### Full-screen mode & transitions
- `fleet shell itest-fleet/alpha --session 19179f` from a plain (non-tmux) terminal: status bar **always on** at top — `[alpha~19…  0:bash 1:bash … ctrl+q/ctrl+o: detach`. `status-left` (session name) was correctly restored even though a prior nested attach had cleared it — the `-u status-left` reset works.
- Killed a window down to 1 in full-screen: bar **stayed on** (autohide correctly scoped to split mode via `@fleet_tab_autohide`).
- `ctrl+q` detached cleanly; a later split reattach to the same session flipped autohide back on (bar hidden at 1 window).
- Tabs survived detach/TUI restarts: a fresh TUI process reattached to the same session and the bar + both tabs rendered immediately at attach time.

### Partial: click-to-switch
Real mouse clicks can't be injected into a headless tmux, so click-to-switch was verified structurally instead: the inner tmux has `mouse on` and the default root-table binding `MouseDown1Status → select-window -t=` present, which is the exact mechanism the PR relies on (same passthrough path as the existing drag-copy). No defect observed; flagging for a quick human click when convenient.

No regressions spotted in surrounding flows (instance provisioning, session create/attach/detach, settings page, TUI navigation/quit).

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build binary & launch TUI | ✅ Pass | `go build` clean; fleet list renders with instance state |
| `fleet up` instance (devcontainer, DinD) | ✅ Pass | `tabs-demo/agent-1` provisioned and running |
| Split attach, 1 window → no tab bar | ✅ Pass | Bar hidden, full pane height (matches issue mockup) |
| `ctrl+x c` → new tab, bar appears at top | ✅ Pass | ` 0:bash  1:bash ` + `ctrl+x c: new tab` hint on row 1 of the pane |
| Tab bar styling | ✅ Pass | Active tab bold colour39, inactive grey 245, `status-position top` |
| `ctrl+x n` / `p` / `0` cycle & jump | ✅ Pass | Pane content switches between per-tab marker shells |
| Auto-hide when back to 1 window | ✅ Pass | `exit` in a tab → bar disappears |
| `(N tabs)` badge in fleet list | ✅ Pass | `● 79cb3c (2 tabs)` appears live at 2 windows, drops at 1 |
| Click-to-switch wiring | ⚠️ Partial | Mouse not injectable headlessly; verified `mouse on` + `MouseDown1Status → select-window -t=` on the inner server |
| Session scoping (user's own session on shared server) | ✅ Pass | Manually created tmux session: no fleet hooks, no status overrides |
| Full-screen attach: bar always on + detach hint | ✅ Pass | Bar stays on at 1 window; `[session] 1:bash … ctrl+q/ctrl+o: detach`; `ctrl+q` detaches, session+tabs persist |
| Mode transition split → full-screen (same session) | ✅ Pass | Autohide flips off, `status-left` restored, detach hint replaces tab hint |
| Full-screen `ctrl+b` prefix | ⚠️ Partial | Works on a pristine container; dead after any split attach (pre-existing global `set -g prefix C-x` — details below) |
| Keybindings overlay lists tab keys | ✅ Pass | `prefix + c`, `prefix + n / p`, `prefix + 0-9` entries shown |

**Overall: ✅ PASS — 12 of 14 scenarios pass, 2 partial; no regressions introduced by this PR.**

## Environment

QA ran fully isolated in the repo's own QA image (`ghcr.io/benjaminbenetti/fleet-man/qa:latest`, Docker-in-Docker, `--privileged`) at head `773d0eb`. A real devcontainer instance (`mcr.microsoft.com/devcontainers/base:debian-12` + tmux) was provisioned with `fleet up`, and the TUI/attach flows were driven interactively through tmux `send-keys` / `capture-pane` — no unit/integration tests were run.

## Detailed findings

### Core tab flow (split panes) — works as mocked in #168

From the TUI (which self-wraps into its own host tmux, so split mode is the default user path), `enter` on the instance opened a split pane attached to the inner session. With one window: no bar, prompt on row 1. After `ctrl+x c`:

```
 0:bash   1:bash                                            ctrl+x c: new tab
vscode ➜ /workspaces/tabs-demo (master) $
```

`ctrl+x p` / `n` / `0` switched between tabs (verified with distinct marker output per tab), the active entry renders bold colour39, and killing back down to one window hid the bar again. The fleet list showed `● 79cb3c (2 tabs)` while 2 windows existed and dropped the badge at 1 window — polled live, no manual refresh needed.

### Full-screen attach

`fleet shell tabs-demo/agent-1 --session …` from a bare (non-tmux) terminal: bar always on at the top with session name on the left and ` ctrl+q/ctrl+o: detach ` on the right, and it correctly does **not** auto-hide at 1 window. Re-attaching a session that had previously been attached split flips the autohide marker and restores `status-left` as described in the PR. Tabs and their shell state persisted across detach/reattach.

### Scoping

Created a plain user tmux session on the instance's shared server: it picked up none of the fleet status options or `window-linked`/`window-unlinked` hooks (the session-scoping in `773d0eb` works). Only the window-status formats are global, matching the documented deliberate scope.

### ⚠️ Finding: full-screen `ctrl+b` prefix is order-dependent (pre-existing, not a regression)

The PR description says tabs in full-screen mode use the default `ctrl+b` prefix. That holds **only until the container's tmux server has seen one split-pane attach**: the nested chain runs `set -g prefix C-x`, which is **global (server-wide) and never reset** by the full-screen chain, so it outlives the split and applies to all later full-screen attaches (and to a user's own sessions on that server) until the tmux server exits.

Repro:
1. Open any session from the TUI as a split pane (the default TUI flow), close it.
2. `fleet shell <fleet>/<instance>` from a bare terminal (full-screen).
3. Press `ctrl+b c` → no new tab; the `c` is typed as a literal character into the shell (into a running agent, in the worst case). `ctrl+x c` works instead.

Since the TUI always attaches split-first, most real containers will be in the `ctrl+x` state, making the PR-documented `ctrl+b` path effectively unreachable. Verified the same `set -g prefix C-x` exists on `main` (old `prefixConf`), so this PR neither introduced nor worsened it — and the in-product keybindings overlay only advertises `ctrl+x` as "(nested)", which stays accurate. Suggested follow-up: `prefix` is a session option, so the nested chain could set it session-scoped (`set prefix C-x`) exactly like this PR already did for the status options and hooks.

BenjaminBenetti:
Re: the QA finding on the order-dependent full-screen prefix — fixed in 897f7b6 rather than deferred, since it made the PR-documented full-screen `ctrl+b` tab keys unreachable in practice. Same pattern as the review fix: the nested chain now sets `prefix C-x` session-scoped (the global default stays `ctrl+b`), and the full-screen chain resets a leftover session override with `set -u prefix`, mirroring the `status-left` reset. The prefix-table `C-x = send-prefix` binding stays global (bind tables have no session scope) and is inert under other prefixes.

Verified QA's exact repro in the isolated harness: split attach → global prefix still `C-b`, user session unaffected → full-screen attach of the same session shows effective prefix `C-b`.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**PR #226 @ 897f7b6** — incremental review of the new commit (`fix(tui): session-scope the nested C-x prefix, reset it full-screen`) since my previous review @ 773d0eb.

**QA's order-dependent full-screen prefix finding: resolved.** The nested chain now sets `prefix C-x` at session scope (the server global stays `ctrl+b`), and the full-screen chain unsets any leftover session override with `set -u prefix`, mirroring the `status-left` reset from the previous round. Nice side-benefit of `-u` over hard-setting `C-b`: a user's own global prefix customization (container `tmux.conf`) is inherited rather than clobbered. Test coverage is right: the nested test asserts both the session-scoped form *and* the absence of `set -g prefix`, and the full-screen test pins `set -u prefix`, so the fix can't silently regress in either direction.

**One observation (non-blocking, inline):** session-scoped options are client-agnostic, so *simultaneous* nested + full-screen attaches to the same session get last-attach-wins semantics — see the inline note on `dotfiles.go:139`. Sequential transitions (the tested path) are handled correctly.

**Nano-nit, comment precision only:** the nested-mode comment says the global `prefix`-table `C-x = send-prefix` binding is "inert under other prefixes". Not quite — in a session with prefix `ctrl+b`, `ctrl+b ctrl+x` now behaves like `ctrl+b ctrl+b` (sends a literal `C-b` to the pane) rather than doing nothing. This bind is pre-existing (predates this PR) and harmless in practice, so no action needed; just flagging that the comment slightly oversells it.

No new blocking issues. All three rounds' findings are addressed; the emitted-script tests now pin every invariant this feature depends on.

internal/tui/dotfiles.go
```diff
@@ -86,19 +86,57 @@ func ShellCommandForSession(config *configutil.Config, session string, cols, row
...
+			` \; set status-left '' \; set status-right ' ctrl+x c: new tab '` +
+			` \; set @fleet_tab_autohide 1 \; ` + tabToggle
 		statusRight = ""
+	} else {
+		// Full-screen mode keeps the bar always on: it carries the
+		// detach hint and doubles as the tab bar. -u resets the
+		// prefix and status-left a prior nested attach overrode, so
+		// the documented full-screen keys (ctrl+b …) hold regardless
+		// of attach order.
+		modeConf = ` \; set @fleet_tab_autohide 0 \; set status on \; set -u status-left \; set -u prefix`
```
BenjaminBenetti, line 139: 🔍 **CodeReview Agent**

**Observation (non-blocking):** `prefix`, `@fleet_tab_autohide`, `status-left/right` are all single session-scoped values shared by every client on the session, so *concurrent* mixed-mode attaches get last-attach-wins semantics. Concretely: with a TUI split pane holding the session, a simultaneous full-screen `fleet shell` attach runs this chain and flips the session prefix back to `ctrl+b` — which the *outer* tmux intercepts in the split pane, making tab keys unreachable from that pane until it re-attaches (and vice-versa: a nested attach hijacks a live full-screen client's prefix to `ctrl+x` and flips auto-hide on).

Sequential transitions — the tested path — are correct, and this is inherent to per-session (not per-client) options, so I don't think it needs code changes for v1. Worth a line in the PR's known-limitations list, and if it ever bites, `client-attached`/`client-detached` hooks re-running the mode chain (or keying the toggle off `#{client_flags}`) would be the escape hatch.

BenjaminBenetti: Agreed — inherent to session-scoped options (tmux has no per-client option scope), and concurrent same-mode attaches (the multi-TUI path) are consistent by construction since both attach chains set identical values. Added to the PR's known-limitations list verbatim with the `client-attached` hook escape hatch noted; no code change for v1.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual QA of `897f7b6` (tabs in panes, #168), run in the `qa/env` DinD image against a real devcontainer instance (`fleet up` → debian-12 fixture), driving the TUI and attach flows exactly as a user would via tmux inside the container.

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` from PR branch | ✅ Pass | `go build ./cmd/fleet` clean |
| `fleet up` + `fleet ls` (sanity) | ✅ Pass | Instance provisions and reports running |
| TUI launch, fleet list hydrates | ✅ Pass | No errors on startup |
| Create session via `a` dialog | ✅ Pass | `work` row appears, "Session created" |
| Split attach, single window | ✅ Pass | No tab bar — auto-hide as mocked in #168 |
| `ctrl+x c` creates tab | ✅ Pass | Bar appears at top: ` 0:bash  1:bash ` + `ctrl+x c: new tab` hint |
| Active-tab styling | ✅ Pass | Bold colour39 accent on current, grey (245) elsewhere |
| `ctrl+x n` / `p` / `0-9` | ✅ Pass | Cycle + jump; pane content follows the window |
| Click-to-switch on tab bar | ✅ Pass | Injected a real SGR mouse click on `1:bash` — switched windows through both tmux layers |
| `(N tabs)` badge in fleet list | ✅ Pass | `● work (2 tabs)` at 2 windows; badge clears at 1 |
| Auto-hide on tab close | ✅ Pass | `exit` in 2nd window → bar disappears, drops to window 0 |
| Keybindings overlay | ✅ Pass | Lists all 3 new entries under "Tmux (inner session)" |
| Full-screen attach (`fleet shell`, no TMUX) | ✅ Pass | Bar always on at 1 window: `[alpha~wor 0:bash … ctrl+q/ctrl+o: detach` |
| Full-screen prefix reset | ✅ Pass | `ctrl+b c` creates tab; `ctrl+x` inert (no leak from prior nested attach) |
| Detach + persistence | ✅ Pass | `ctrl+q` detaches; session + tabs survive detach/reattach and TUI quit |
| Mode transition (full-screen → split) | ✅ Pass | status-right flips to new-tab hint, status-left cleared, `ctrl+x` prefix restored, auto-hide re-armed |
| Shared-server isolation | ✅ Pass | User-owned session on same inner server keeps default status/prefix; toggle hooks exist only session-scoped on `alpha~work` (global hook slots empty) |

**Overall: ✅ PASS — 17 of 17 scenarios passing**

## Details

**Environment.** `ghcr.io/benjaminbenetti/fleet-man/qa:latest` (privileged, Docker-in-Docker), PR head `897f7b6`. Fixture repo from `integration/fixture`, `fleet up alpha --repo file://…`, TUI driven headless in a 160×40 tmux (`send-keys` + `capture-pane`), inner state cross-checked with `tmux` inside the devcontainer.

**Split-pane tab flow (core of #168).** Opened `work` in a split from the TUI. With one window the pane top line is the shell prompt — no bar, matching the issue mockup's single-tab pane. `ctrl+x c` immediately rendered the bar on row 1 with both entries and the `ctrl+x c: new tab` hint; capture with escapes confirmed the active entry is `bold` + `colour39` on grey — consistent with the TUI palette. Cycling (`n`/`p`), jumping (`0`), and switching all landed on the right window, verified by distinct marker output (`echo FIRST/SECOND-TAB-CONTENT`) per window. Killing the second window auto-hid the bar and returned to window 0 with scrollback intact.

**Click-to-switch was tested for real.** Injected an SGR mouse press/release (`\e[<0;12;1M` / `…m`) at the `1:bash` entry into the host pane: the click passed through the host tmux into the inner server and switched the active window. Mouse passthrough + `MouseDown1Status` works end-to-end, not just config-present.

**Full-screen mode.** Note for reviewers: plain `fleet` outside tmux relaunches itself into a host tmux wrapper (`fleet-<hex>`), so the TUI path always split-attaches; genuine full-screen (`nested=false`) was exercised via `fleet shell tabsqa-fleet/alpha --session work` outside tmux. There the bar is always on even at 1 window, `status-left` carries the session name and `status-right` the detach hint, and `ctrl+b c` (default prefix) creates tabs while `ctrl+x` is inert — the prior nested attach's prefix override did not leak.

**Mode transition.** Same session attached full-screen (2 windows created) → `ctrl+q` detach → reattach split from the TUI: bar stayed visible (2 windows), hint flipped from detach to `ctrl+x c: new tab`, `status-left` cleared, `ctrl+x n` cycled again, and after killing the extra window auto-hide kicked back in. The fleet list badge tracked live: `(2 tabs)` ↔ no badge.

**Isolation.** Created an unrelated `usersess` on the same inner tmux server: its `status-position` stays `bottom` (global default untouched), prefix stays `C-b`, and `show-hooks -g` carries no toggle commands — they exist only on `alpha~work`. The globally-set `window-status-format` styling is the one shared knob, as documented in the code comments.

No defects found. Known limitations in the PR description (background-tab invisibility to capture/MCP, layout presets not snapshotting tab sets) were not contradicted by anything observed.

PR #228, opened 2026-07-08
refactor/protoconv-single-source one source of truth for proto converters, paths, quoting

BenjaminBenetti:
## Summary

Maintainability pass focused on single sources of truth and removing duplication (SOLID/DRY). Net **−375 lines** with strictly more test coverage.

- **`internal/protoconv` — one home for all domain↔proto conversion.** The same converters were maintained in triplicate (server `convert.go`/`mutations.go`/`config.go`, TUI `client.go`, CLI `automation.go`/`jobs.go`) because the depguard client boundary forbids clients from importing the server — the documented "6-converter trap": every new Trigger/FleetSettings field had to be added in up to 6 places. Like `configutil`, the new package is a deliberate carve-out: pure mapping over DTO shapes, no state I/O, importable from both sides of the boundary (noted in `.golangci.yml`).
- **Fixes a real data-loss bug the trap caused:** the CLI's trigger converters were missing `Trigger.Disabled`, so any `fleet trigger`/`fleet agent` write round-tripped the full trigger list through the lossy converters and silently **re-enabled every disabled trigger** in the fleet (exactly the drift predicted in PR #195's notes; it was masked by a round-trip test that never set `Disabled: true`).
- **Trap-killer tests:** `protoconv`'s reflection round-trip suite fills *every field* of every domain struct with distinct non-zero values and requires deep equality after to-proto → from-proto — a future field added without converter support fails CI automatically, naming the exact field. Plus exhaustive enum, tri-state presence (`PreferFleetLaunch`), wire presence rules, and nil/empty convention tests consolidated from the three per-surface test files.
- **One deliberate divergence kept, now explicit:** `ConfigFromProto(pc, base)` — the server's SetConfig write path passes a zero `Config` (never mask a field the caller cleared), the TUI read path passes `DefaultConfig()` (absent fields render as defaults). Both behaviors are locked by tests.
- **`state` path helpers delegate to `fleetpaths`.** `FleetDir`/`WorkspacesDir`/`WarnPath` were defined in both packages with keep-in-sync comments. `WarnPath` is a writer/reader contract across the boundary (server writes the banner file, TUI reads it) — drift would silently kill post-create warnings.
- **`fleetgrpc.CountInstanceStatuses`** — the running/stopped/other bucketing was duplicated between `fleet status` and the MCP `fleet_status` tool; the rule now lives once with the other pure proto-type domain methods.
- **`internal/shellquote`** — the injection-sensitive `'\''` single-quote idiom was written out four times (backend, dotfiles, TUI ×2 variants, fleet). One dependency-free leaf implementation now; established names (`backend.ShellQuote`, `dotfiles.ShQuote`) remain as thin delegates.
- **`fleetclient` owns the remote-endpoint env contract** — admiralmcp re-implemented the exact body of `fleetclient.IsRemote()`; the `FLEET_GATEWAY`/`FLEET_SERVER`/`FLEET_TOKEN` names are now constants in the package that owns their semantics.

## Verification

- `make test` (depguard lint + full unit suite) passes.
- Integration subset run in an isolated Docker container: CLI lifecycle (010–057), `fleet status` (040), TUI startup (100), TUI settings (200), automation CLI CRUD (392) — all pass. 025_copy fails in that sandbox with a permission error, but fails **identically with a main-built binary** in the same container (root-vs-`vscode` UID artifact of the sandbox, not a regression); the full suite runs in CI.

## Deferred follow-ups (mapped, deliberately not in this PR)

- Backend `CaptureAllSessions`/`ListSessions`/`RunScript`/`AgentToolProbe` wrappers repeated across the three backends (extractable via a script-command seam).
- Docker-service seam boilerplate (`runDocker`/`execCommand`/per-fleet locks) repeated across buildkit/debcache/imagecache/fleetnet, and the larger debcache↔imagecache structural mirror.
- TUI form-dialog scaffold similarity (subtle, low payoff/risk ratio).
- A few hand-rolled temp+rename writes that could adopt `internal/atomicfile`.


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed `4aab582` (all 6 commits). I rebuilt the branch and ran the full unit suite locally (`go build ./...` + `go test ./...`) — everything passes. Overall this is a well-executed consolidation; the findings below are minor.

**Verified claims**

- **The `Disabled` data-loss bug is real and fixed.** The deleted CLI converters (`internal/cli/automation.go`: `protoTriggersToFleet` / `fleetTriggersToProto`) both omitted `Disabled`, so any `fleet trigger`/`fleet agent` write would silently re-enable every disabled trigger — and the deleted round-trip test indeed never set `Disabled: true`, which is why it passed. The new `protoconv` converters carry it both ways, and the reflection round-trip suite would now catch a regression by field name.
- **Behavior preservation.** I diffed each consolidated converter copy (server `convert.go`/`config.go`/`mutations.go`, TUI `client.go`, CLI `automation.go`/`jobs.go`) against `protoconv` line-by-line: presence rules, nil/empty conventions, enum mappings, and the zero-base (SetConfig) vs `DefaultConfig` base (TUI read) fork are all faithfully preserved. The one seeming difference — old `if i.Automated { pi.Automated = true }` vs new unconditional assignment — is semantically identical for a proto3 bool.
- CLI `mutateAutomation` reuses the live proto settings as the carrier and nil-guards `pf.GetSettings()`, so untouched settings fields survive the wholesale write. 👍
- `fleetgrpc.CountInstanceStatuses` matches both prior bucketing copies exactly (UNSPECIFIED → Other, per-fleet total == instance count).
- No string literals of `FLEET_GATEWAY`/`FLEET_SERVER`/`FLEET_TOKEN` remain outside `fleetclient/endpoint.go` (comments aside); `admiralmcp`'s guard is behavior-identical via `fleetclient.IsRemote()`.

**Findings**

1. **Missed fifth copy of the quoting idiom** — `internal/fleetlaunch/fleetrc.go:45` (`renderFleetRC`) still hand-rolls the single-quote escape:
   ```go
   quoted := "'" + strings.ReplaceAll(instanceName, "'", `'\''`) + "'"
   ```
   This is exactly the injection-sensitive path the PR argues should live once (an instance name is spliced into a sourced rc file). `fleetlaunch` can import the new leaf `internal/shellquote` with no cycle — worth folding into this PR (`shellquote.Single(instanceName)` is a two-line change), or at least adding to the deferred-follow-ups list so the "one implementation" claim stays true.

2. **Enum exhaustiveness test is hand-maintained** (inline comment on `protoconv_test.go`) — the reflection filler pins one status/backend member, so a *new* domain status added without a proto mapping is only caught if the developer also extends `TestEnumsExhaustive`'s literal list. Suggestion inline.

**Nit (no action needed):** `ConfigFromProto` mutates `base` in place and returns it — this is documented and both current callers pass fresh allocations, but future callers passing a shared/global config would be surprised. Fine as is given the package is documented as transitional.

internal/protoconv/protoconv_test.go
```diff
@@ -0,0 +1,266 @@
...
+	in.DefaultBackend = string(fleet.BackendCodespaces)
+	requireEqual(t, ConfigFromProto(ConfigToProto(&in), &state.Config{}), &in)
+}
+
+// --- Intent-specific cases the fuzz fill can't express -----------------------
+
+// TestEnumsExhaustive proves every enum member survives both directions; a new
+// domain status/backend added without a proto mapping falls to UNSPECIFIED and
+// fails here.
+func TestEnumsExhaustive(t *testing.T) {
```
BenjaminBenetti, line 151: 🔍 **CodeReview Agent**

Small gap in the trap-killer guarantee: the comment says "a new domain status/backend added without a proto mapping falls to UNSPECIFIED and fails here" — but that only holds if the developer also extends this hand-maintained literal list. The reflection filler pins `StatusRebuilding`/`BackendCoder`, so a new `fleet.Status*` constant added without converter support sails through both the round-trip suite and this test.

Since there's no canonical `[]InstanceStatus` slice in `internal/fleet` to iterate, one option that closes at least the proto side automatically: iterate the generated `fleetgrpc.InstanceStatus_name` / `fleetgrpc.BackendType_name` maps and assert every non-UNSPECIFIED member survives `FromProto → ToProto` — a proto enum member added without a domain mapping then fails by name. The domain-side list would still be manual, but a one-line comment here ("extend this list when adding a status") would make the maintenance contract explicit.

BenjaminBenetti:
Addressed the review in edc2a56:

1. **fleetlaunch quoting copy** — folded into `internal/shellquote`: `renderFleetRC` now uses `shellquote.Single(instanceName)`. Good catch; that was the exact injection-sensitive case the leaf package exists for.
2. **Enum exhaustiveness test** — now derives its member lists from the generated `fleetgrpc.InstanceStatus_name` / `BackendType_name` maps instead of a hand-maintained literal list. Any proto enum member added without a domain mapping fails by name with no test edit. (The domain side stays implicit as you noted — a new domain status can't reach the wire without a proto member, and adding that member is what trips the test.)
3. `ConfigFromProto` base-mutation nit — leaving as documented, per your note.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` binary (`go build ./cmd/fleet`) | ✅ Pass | Builds clean at `4aab582`; `--help` renders all commands |
| TUI launch & dashboard renders | ✅ Pass | Fleet/instance rows, status indicators, keybind bar — no errors |
| Create instance (`fleet up` → devcontainer) | ✅ Pass | Instance `qa-fleet/alpha` created and running |
| `fleet ls` / `fleet status` bucketing | ✅ Pass | "No fleets" empty case → 1 running → 1 stopped → running again, counts correct each time |
| `fleet exec` with quote-hostile args | ✅ Pass | `it's a "quoted" $(noexec) \`noexec2\` arg` passed through literally — no injection/mangling (shellquote path) |
| **Trigger `Disabled` preserved across CLI writes** (headline fix) | ✅ Pass | Stayed disabled through `trigger edit`, `agent create`, and `agent delete` writes |
| TUI automation view + disable toggle | ✅ Pass | `m` opens view, `s` disables, `· disabled` marker persists across TUI restart |
| TUI settings: defaults + toggle round-trip | ✅ Pass | Defaults render (`ConfigFromProto` default-base read path); toggle persists; sibling settings untouched (zero-base write path) |
| `stop`/`start` + `down`/`destroy` lifecycle | ✅ Pass | Clean transitions, correct status after each step |
| MCP `fleet_status` tool / gateway remote env contract | ⏭️ Skipped | Not exercised manually; CLI half of the shared `CountInstanceStatuses` bucketing verified above |

**Overall: ✅ PASS — 9 of 9 executed scenarios passing (1 skipped)**

---

### Environment

All testing ran isolated in the repo's own QA image (`ghcr.io/benjaminbenetti/fleet-man/qa:latest`, privileged DinD), PR branch checked out at head `4aab582`. Binary built inside the container with `go build -o /tmp/fleet ./cmd/fleet`. No automated tests were run — everything below was driven interactively (CLI + headless tmux TUI) as a user would.

### Detailed findings

**1. Headline fix — disabled triggers survive CLI automation writes.** This is the data-loss bug the PR claims to fix, and it holds:

1. `fleet agent create qa-fleet builder --backend devcontainer --system-prompt "be terse"`
2. `fleet trigger create qa-fleet nightly --agent builder --cron "0 0 1 1 *" --prompt "run nightly"` and a second trigger `weekly`
3. In the TUI: `m` → automation view, cursor to `nightly`, `s` → row shows `nightly ⏱ 0 0 1 1 * → builder · disabled`, status line confirms `Disabled trigger "nightly"`; `state.json` shows `"disabled": true`
4. CLI write to the *other* trigger: `fleet trigger edit qa-fleet weekly --prompt "run weekly CHANGED"` → after the write, state shows `[('nightly', True), ('weekly', False)]` with weekly's prompt updated — **nightly is still disabled** (on `main` this round-trip silently re-enabled it)
5. Repeated with `fleet agent create qa-fleet builder2 …` and `fleet agent delete qa-fleet builder2` — nightly stayed disabled through both, and a reopened TUI still renders the `· disabled` marker

**2. Status bucketing (`CountInstanceStatuses` consolidation).** `fleet status` reported correctly at every lifecycle stage: `No fleets` when empty; `1 instances (1 running, 0 stopped)` after `up`; `(0 running, 1 stopped)` after `stop`; back to `(1 running, 0 stopped)` after `start`; `0 instances` after `down`; `No fleets` after `destroy`.

**3. Shell quoting (`internal/shellquote` consolidation).** `fleet exec qa-fleet/alpha printf '%s\n' "it's a \"quoted\" \$(noexec) \`noexec2\` arg with  spaces"` printed the argument byte-for-byte — single quotes, double quotes, `$( )`, backticks, and double spaces all survived; nothing was executed. A `bash -c` payload with nested single quotes and `$(whoami)` also behaved (`who: vscode`).

**4. Settings read/write divergence (`ConfigFromProto(pc, base)`).** Fresh settings page rendered defaults (`Tmux vim keys [on]`, `Show help text [on]`, dotfiles `(not set)`, etc. — the `DefaultConfig()` read base). Toggling `Tmux vim keys` off, closing, and reopening showed it persisted as `[off]` while `Show help text` stayed `[on]` — a cleared/changed field didn't clobber siblings (the zero-base write path). Restored to `[on]` afterwards.

**5. Instance lifecycle sanity.** `up` → `ls` → `exec` → `stop` → `start` (postStartCommand re-ran, `outcome: success`) → `down` → `destroy` all completed with clean output and no daemon errors.

### Non-blocking observations (not regressions)

- `fleet trigger list` gives no indication of disabled state — `nightly` looked identical to enabled triggers in CLI output; the state is only visible in the TUI (and `state.json`). Pre-existing UX gap this PR neither caused nor fixed; a `DISABLED` column/marker might be a nice follow-up.
- `fleet exec <name>` cwd-inference requires the current repo to have an `origin` remote; with a local `file://`-only fixture the qualified `fleet/instance` form is needed. Environment quirk of my fixture, unrelated to this PR.

BenjaminBenetti:
Thanks QA — note the run was at `4aab582`; two review-response commits landed after (`edc2a56`, `7537ccb`): fleetlaunch's quoting copy folded into shellquote, the enum test now derives from the generated proto maps, protoconv moved inside the depguard client rule (machine-checked purity via the configutil aliases), `StatusCounts.Merge`, and the TUI's `shQuote` pointing straight at the leaf. All are behavior-identical to what QA exercised; CI is re-running on the new head.

On the non-blocking observation: agreed `fleet trigger list` (and the `fleet_trigger_*` MCP output) should surface the `Disabled` flag now that it reliably survives CLI writes — left it out of this PR to keep the refactor behavior-neutral. Worth a small follow-up issue.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed the update (`4aab582..edc2a56`). Both findings from the previous review are addressed — verified, not just eyeballed:

1. **Fifth quoting copy folded in** — `renderFleetRC` now uses `shellquote.Single`, which is byte-identical to the replaced expression (`"'" + ReplaceAll(name, "'", `'\''`) + "'"`), so rendered rc output is unchanged. The quote-injection tests (`TestRenderFleetRCQuotesSingleQuotesInName`) still pass, and a repo-wide grep confirms no hand-rolled copies of the idiom remain outside `internal/shellquote` (the `automation_match.go` hit is a tokenizer character-class check, not the escape idiom).

2. **Enum test no longer hand-maintained** — `TestEnumsExhaustive` now iterates the generated `fleetgrpc.InstanceStatus_name` / `BackendType_name` maps, skipping UNSPECIFIED and asserting `FromProto` yields a non-empty domain value that maps back to the same proto member. This goes further than suggested (the literal list is gone entirely), and the failure mode is exactly right: a proto member added without converter support hits `StatusFromProto`'s `default: return ""` and fails by name. The one residual gap — a new domain constant added without a `.proto` member — is now explicitly documented in the test comment as out of scope (such a value can't reach the wire's authoritative member list); that's a reasonable, recorded trade-off.

**Verification at `edc2a56`:** `go build ./...`, `go vet ./...`, and the full `go test ./...` (40 packages) all pass locally.

No new issues in the delta. Nothing blocking from my side.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the full diff at `7537ccb` (8 commits, 40 files), with focus on the two commits since the last review round (`edc2a56`, `7537ccb`).

**Verified locally:**
- `go build ./...`, `go vet ./...`, and the full unit suite (40 packages) pass at head.
- The `Trigger.Disabled` data-loss fix is real and covered: the old CLI converters (`protoTriggersToFleet`/`fleetTriggersToProto`) omitted `Disabled` on both sides; the consolidated converters carry it, and the reflection filler sets bools to `true` so the round-trip suite now pins it.
- Prior review feedback is properly addressed: `TestEnumsExhaustive` now derives its member lists from the generated `InstanceStatus_name`/`BackendType_name` maps, so a new proto enum member with a missing domain mapping fails CI without touching the test.
- The depguard carve-out works as described: `protoconv` sits inside the client rule (purity machine-checked), and the `!**/*_test.go` exclusion legitimately covers the test file's direct `internal/state` import.
- Every delegation move (server/TUI/CLI converters, `state`→`fleetpaths`, `fleetclient` env constants, `shellquote` call sites, `CountInstanceStatuses`) is behavior-preserving — including the subtle ones (TUI read path's `DefaultConfig` base matches the old `protoConfigToLegacy` exactly; `counts.Total()` equals the old `len(f.GetInstances())`). No stray copies of the `'\''` idiom remain outside `internal/shellquote`.
- The one deliberate divergence (`ConfigFromProto(pc, base)`) is well-documented and locked by `TestConfigFromProtoBases`.

**Findings (2 inline):**
1. `internal/protoconv/protoconv_test.go` — the round-trip suite has a blind spot the deleted `TestConfigToProtoEncodesFullConfig` covered: symmetric encode/decode bugs. I verified empirically that swapping `Codespaces.Machine`↔`IdleTimeout` on the wire **in both directions** passes the entire suite. Details + suggested fix inline.
2. `internal/shellquote/shellquote.go` — mangled `'\''` in the doc comment (cosmetic, but this is now the canonical documentation of the idiom).

Overall: a careful, well-tested consolidation. The trap-killer test design is sound, the data-loss fix is genuine, and the depguard reasoning holds up under inspection. Finding 1 is worth closing since wire placement is exactly the contract this package now owns.

internal/protoconv/protoconv_test.go
```diff
@@ -0,0 +1,282 @@
...
+	in := filled[state.State](t)
+	requireEqual(t, StateFromProto(StateToProto(&in)), &in)
+}
+
+func TestConfigRoundTripAllFields(t *testing.T) {
+	in := filled[state.Config](t)
+	// DefaultBackend is a plain string carrying enum semantics (the one field
+	// the generic filler can't know about): only valid backend names survive.
+	in.DefaultBackend = string(fleet.BackendCodespaces)
+	requireEqual(t, ConfigFromProto(ConfigToProto(&in), &state.Config{}), &in)
```
BenjaminBenetti, line 141: 🔍 **CodeReview Agent**

The round-trip suite guards field **presence** but not wire **placement** — it's blind to symmetric encode/decode swaps. Verified empirically: swapping `Codespaces.Machine`↔`IdleTimeout` in *both* `ConfigToProto` and `ConfigFromProto` passes every test in this package, because the same inverted mapping decodes what it encoded. The deleted `TestConfigToProtoEncodesFullConfig` (tui/client_test.go) would have caught this — it asserted actual proto field values (`pc.GetCodespaces().GetMachine() == "m1"`, etc.).

This matters more now, not less: with one shared implementation, a symmetric swap is self-consistent end-to-end within one binary version, so nothing surfaces at runtime either — but the wire contract is silently wrong for any cross-version client/server pair (remote daemon/gateway setups) and for the proto field semantics themselves.

Suggested fix: one wire-side assertion test per top-level converter (Config, FleetSettings, Instance, Trigger) that fills a struct with distinct values and asserts a handful of proto getters directly — essentially resurrect the deleted test's shape inside this package. Cheap to write, and it makes the "one source of truth" also the one source of *correctly placed* truth.

internal/shellquote/shellquote.go
```diff
@@ -0,0 +1,24 @@
...
+// internal/backend, internal/dotfiles, internal/tui and internal/fleet —
+// a quoting bug is shell injection, so the rule lives once, in a leaf package
+// with no dependencies that every layer (backend code and depguard-restricted
+// client code alike) may import.
+package shellquote
+
+import "strings"
+
+// Single wraps value in single quotes with embedded single quotes escaped via
+// the '\” idiom (close the quote, emit an escaped quote, reopen the quote),
```
BenjaminBenetti: 🔍 **CodeReview Agent**

Nit: the doc comment reads `the '\” idiom` — the intended `'\''` got mangled into `'\` + U+201D (curly right-double-quote). Inherited from the old copies (backend/dotfiles had the same corruption), but since this package is now the canonical home of the idiom, its doc should spell it correctly: `'\''`. Same mangled sequence appears in the package comment of internal/backend/copy_file.go's delegate and internal/fleet/automation.go history — a one-pass `grep -rn '\\\”'` cleanup would catch them all.

BenjaminBenetti:
Both round-3 findings addressed in 4696987:

1. **Wire placement** — new `internal/protoconv/wire_test.go` asserts proto getters directly for every converter (Trigger, Agent, Instance, FleetSettings, Config, GroupLayout): distinct values per string field, and a set-one-assert-its-getter loop for the bool toggles since two bool values can't be made mutually distinct. I confirmed the suite now catches the exact swap you demonstrated (Machine↔IdleTimeout inverted in both directions fails `TestConfigWirePlacement`). Good find — the cross-version framing is exactly right.
2. **Mangled idiom doc** — `'\''` restored in shellquote's doc comment; a repo-wide grep for the U+201D corruption shows that was the last one (the backend/fleet copies that carried it were deleted by this PR).

internal/protoconv/wire_test.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+// The round-trip suite (protoconv_test.go) guards field PRESENCE but is blind
+// to symmetric encode/decode swaps: a converter pair that writes Machine into
+// idle_timeout and reads it back from idle_timeout round-trips cleanly. These
+// tests pin wire PLACEMENT by asserting proto getters directly, so the wire
+// contract stays correct for cross-version client/server pairs — one distinct
+// value per string field, and a set-one-assert-its-getter loop for bools
+// (whose two values can't be made mutually distinct).
+
```
BenjaminBenetti, line 36: 🔍 **CodeReview Agent**

`Fleet` and `State` are the two converters without a placement test in this file, and `Fleet` is the one that matters: `Name` and `Remote` are two same-type strings in the same message. Verified empirically at `4696987`: swapping them symmetrically in both `FleetToProto` and `FleetFromProto` passes the **entire** protoconv suite (the round-trip suite is placement-blind by design, and no wire test covers them). A 5-line `TestFleetWirePlacement` closes it. `State`'s only scalar (`LastSeenVersion`) has no same-type sibling to swap with, so it's safe by construction — but a trivial assertion would keep the "every converter" invariant honest.

Minor, optional hardening: these literals are hand-maintained, so a *future* field added with symmetric misplacement would pass both suites (round-trip can't see placement; this file doesn't know the field exists). Same class as the enum-literal finding from round 1 — an arity pin per covered struct, e.g. `if reflect.TypeOf(fleet.Trigger{}).NumField() != 12 { t.Fatal("Trigger grew a field: extend the wire test") }`, would make this file self-enforcing the way the enum test now is.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed the update (`7537ccb..4696987`: wire-placement tests for every converter + shellquote doc fix). Both round-3 findings verified fixed — empirically, not just eyeballed:

1. **Wire placement** — I re-introduced the exact symmetric swap from last round (`Machine`↔`IdleTimeout` in both `ConfigToProto`/`ConfigFromProto`): `TestConfigWirePlacement` now fails immediately, naming the misplacement (`codespaces wire placement wrong: machine:"it" idle_timeout:"ma"`). The one-directional design is sound: the wire tests pin to-proto placement, and the round-trip suite forces from-proto to invert it, so both directions are pinned for every asserted field. Field coverage checked struct-by-struct against the domain types: Trigger 12/12, Agent 4/4, Instance 13/13, GroupLayout 5/5, CoderParameter 6/6, LayoutPreset 3/3, FleetSettings all scalars + all 7 bools via the set-one-per-iteration loop, Config all 6 sub-structs + `DefaultBackend` including the tri-state nil assertions. The bool-isolation technique is the right answer to value-indistinguishability.
2. **Shellquote doc** — the `'\''` idiom now renders correctly; the U+201D mojibake is gone.

`go build ./...` and the full unit suite pass at `4696987`.

**One finding** (inline comment on `wire_test.go`): the "every converter" claim misses `FleetToProto`/`FleetFromProto` — and `Fleet.Name`/`Remote` are exactly the swap-prone shape (two strings, same message). Verified: a symmetric Name↔Remote swap in both converters passes the entire protoconv suite at head. A tiny `TestFleetWirePlacement` closes the last hole; `State` is safe by construction (one scalar, no same-type sibling). Also left an optional note on making the file self-enforcing against future fields via per-struct arity pins, mirroring what the enum test already does.

Not blocking — the finding is a test-coverage gap in a hardening commit, not a product bug. Everything shipped in this round does what it claims.

BenjaminBenetti:
Round-4 finding closed in 133c435:

- **`TestFleetAndStateWirePlacement`** pins `Fleet.Name`/`Remote` (the swap you demonstrated), the instance-list placement, and `State`'s scalar + both map keys.
- Took the optional hardening too: every wire test now opens with an **arity pin** (`pinArity[T](t, n)`) on the domain struct(s) it asserts — a struct growing a field fails the pin with instructions to extend the placement assertions, so the hand-maintained literals can no longer silently fall behind. That closes the last "hand-maintained list" class in the package: enums derive from the proto maps, presence derives from reflection fill, and placement is now arity-guarded.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

Manual, behavior-level QA of `4696987` inside an isolated Docker-in-Docker container (`ghcr.io/benjaminbenetti/fleet-man/qa`), driving the real `fleet` binary (CLI + TUI-in-tmux) against real devcontainers. No automated tests run.

| Test | Result | Notes |
|------|--------|-------|
| Build `fleet` from PR head | ✅ Pass | `go build ./cmd/fleet` clean |
| CLI lifecycle: `up` ×2, `ls`, `stop`, `start`, `down` | ✅ Pass | Both instances provisioned, state transitions correct |
| `fleet status` counts (empty / 2 running / 1+1 / totals) | ✅ Pass | `CountInstanceStatuses`+`Merge` bucketing correct at every step |
| Automation CLI CRUD (`agent`/`trigger` create·edit·rm·ls) | ✅ Pass | New protoconv path; `${PROMPT}`-style values survive intact |
| **Disabled-trigger preservation (headline fix)** | ✅ Pass | TUI-disabled trigger survives `trigger edit`, `agent edit`, `trigger create`+`rm` via CLI |
| Bug reproduction on `main`-built binary | ✅ Pass | Same edit on main **wipes** `disabled` — confirms this PR fixes real data loss |
| TUI: startup, automation view, `s` toggle, refresh, quit | ✅ Pass | `· disabled` marker renders; persists to state.json |
| TUI settings: defaults render, explicit-false persists | ✅ Pass | `tmux_vim_keys:false` round-trips; untouched fields keep defaults (tri-state) |
| `fleet cp` round-trip, quote-hostile name+content | ✅ Pass | `it's a "tricky" $file \`name\`.txt` — byte-identical both directions (shellquote) |
| `fleet exec` sanity | ✅ Pass | Command runs in instance, output/exit code correct |
| `FLEET_SERVER` remote-endpoint contract | ✅ Pass | Routes to remote, clear dial error, no local fallback/auto-spawn |
| `fleet trigger logs` empty state | ✅ Pass | Friendly "No events recorded" message |

**Overall: ✅ PASS — 12 of 12 scenarios passing**

---

### Detailed findings

**1. Headline fix verified end-to-end (disabled-trigger data loss).**
Steps: `fleet up alpha` → `fleet agent create itest-fleet builder` → `fleet trigger create nightly` + `weekly` → in the TUI (`m` → cursor on `nightly` → `s`) disabled `nightly`; confirmed `"disabled": true` in `~/.fleet/state.json` and the dim `· disabled` row. Then ran each CLI read-modify-write path with the PR binary:
- `fleet trigger edit itest-fleet weekly --prompt …`
- `fleet agent edit itest-fleet builder --system-prompt …`
- `fleet trigger create` + `rm` of a third trigger

After every write, `nightly` remained `"disabled": true` (state.json and TUI after `r` refresh). ✅

**2. Regression demonstrated on `main`.** Built a second binary from `main` in the same container and ran the identical `fleet trigger edit … weekly` against the same daemon: `nightly`'s flag was silently dropped (`"disabled": null` → trigger re-enabled). This is exactly the data-loss bug the PR description claims; the PR closes it.

**3. Settings converters (`ConfigFromProto`/`ConfigToProto`) behave per the documented divergence.** Fresh state renders all defaults (`Tmux vim keys [ on ]`, dotfiles unset, browser toggles off). Toggling `Tmux vim keys` off persisted an explicit `"tmux_vim_keys": false` in `config.json` (write path did not mask the cleared field), and reopening rendered `[ off ]` while untouched settings kept their defaults (read path fills absent fields from `DefaultConfig()`).

**4. `shellquote` consolidation holds under hostile input.** Uploaded and downloaded `it's a "tricky" $file \`name\`.txt` containing single/double quotes, `$dollar`, backticks and spaces via `fleet copy`; `diff` confirmed byte-identical round-trip, and `fleet exec … cat` inside the instance showed unmangled content.

**5. `fleet status` bucketing (`fleetgrpc.CountInstanceStatuses` + `StatusCounts.Merge`).** Verified empty state ("No fleets"), `2 running, 0 stopped`, `1 running, 1 stopped` after `fleet stop`, and the fleet-level + `Total:` lines agreeing at every step.

**6. Endpoint contract (`fleetclient` env constants).** `FLEET_SERVER=127.0.0.1:9 fleet ls` attempted the remote target and failed with a clear `fleet server not reachable at 127.0.0.1:9 … connection refused` error — no silent fallback to the local socket and no local auto-spawn. Unsetting it restored normal local operation.

No regressions observed in any exercised flow. Environment torn down after the run.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed the update (`4696987..133c435`: Fleet/State wire-placement test + arity pins). The round-4 finding is verified fixed — empirically, not just eyeballed:

1. **Fleet.Name/Remote swap now caught.** I re-introduced the exact symmetric swap from last round (`Name`↔`Remote` in both `FleetToProto` and `FleetFromProto`): `TestFleetAndStateWirePlacement` fails with `fleet wire placement wrong: name:"rm" remote:"nm"` where previously the entire suite passed. `State`'s scalar and both map keys are asserted too. (Direction coverage is sound: the wire tests pin the ToProto side, and the round-trip suite — domain→proto→domain — transfers that to FromProto.)
2. **Arity pins are complete.** I cross-checked `pinArity` coverage against every converter in the package: all 16 domain structs protoconv maps (Trigger, Agent, Instance, Fleet, State, FleetSettings + LayoutPreset/CoderParameter, Config + its 6 nested settings structs, GroupLayout) are pinned in the test that asserts them; the two enums are separately self-enforced via the generated `_name` maps. Pin counts match the current structs, and `reflect.TypeFor` is fine on the module's go 1.25.
3. **Suite health at head:** `go build ./...`, `go vet ./...`, and the full unit suite pass at `133c435`.

I also did a residual sweep of the parts of the diff earlier rounds spent less time on — no new findings:

- `state.FleetDir/WorkspacesDir/WarnPath` → `fleetpaths` delegation is byte-identical to the old bodies (`fleetpaths.Dir()` is the same `/home/vscode/.fleet` join), so no path behavior change.
- `fleetgrpc.CountInstanceStatuses`/`StatusCounts.Merge` reproduce both deleted bucketing loops exactly; the per-fleet line's switch from `len(f.GetInstances())` to `counts.Total()` is equivalent since the `default` bucket makes counting total.
- `fleetclient` env constants and `admiralmcp`'s `IsRemote()` guard are drop-in equivalents of the replaced literals.

Two micro-observations, neither blocking:

- The arity pin is a tripwire, not a proof — a future author can bump the pin without adding the placement assertion. The failure message does say to add assertions first, which is about as much as a test can enforce; just noting the residual gap is now human-discipline-shaped rather than coverage-shaped.
- `NumField()` counts unexported fields, so a non-wire unexported field addition would also trip a pin. Arguably a feature (forces a conscious decision + pin bump); flagging so it's a known behavior, not a surprise.

No new issues. From the code-review side this looks ready: four rounds of findings all verified closed, and the trap-killer suite now genuinely self-enforces presence (round-trip), placement (wire tests), enum membership (generated maps), and growth (arity pins).

PR #235, opened 2026-07-21
fix(tui): portable placeholder sleep so session-group restore works on macOS

shawn-vincent-well:
Fixes #231

## What

Phase 1 of restoreGroupCmd creates placeholder panes with 'sleep infinity' to establish layout geometry before respawning each pane with its real session. 'sleep infinity' is a GNU coreutils extension: macOS/BSD sleep rejects it and exits immediately, so on a Mac every placeholder pane died the moment it was created. Follow-up splits then targeted dead pane IDs, select-layout had nothing to lay out, and the respawn phase found no slots. Net effect: every session-group switch or restore broke on macOS.

This swaps the placeholder to 'sleep 2147483647' (a named constant with a comment explaining why). Both GNU and BSD sleep accept a plain seconds count, and the placeholder only needs to outlive the respawn that happens moments later, so behavior on Linux is unchanged.

## Testing

- Minimal repro from #231: 'tmux split-window ... -- sleep infinity' loses the pane within a second on macOS; with the numeric sleep the pane survives. Verified on macOS (arm64), tmux 3.5a.
- Full manual pass on macOS in the TUI: two session groups (one with 2 panes, one with 1), switching via Ctrl+PageUp/PageDown lands correctly every time, pane positions and a custom resize survive round trips, rapid switching (~10 fast flips) ends in a consistent state, Ctrl+Q closes cleanly, sessions and scrollback survive.
- go build ./... clean, internal/tui tests pass.

### DIFF
--- a/internal/tui/splitpane.go
+++ b/internal/tui/splitpane.go
@@ -202,6 +202,13 @@ func unbindDefaultSplitKeys() {
 	_ = exec.Command("tmux", "bind-key", `"`, "display-message", msg).Run()
 }
 
+// placeholderSleepSecs keeps Phase-1 placeholder panes alive until they are
+// respawned with their real session. It must be a plain seconds count:
+// `sleep infinity` is a GNU coreutils extension, and macOS/BSD sleep rejects
+// it and exits immediately — the pane then closes before the layout apply and
+// respawn phases run, breaking every group restore/switch on macOS.
+const placeholderSleepSecs = "2147483647"
+
 // splitWindowWithRetry runs `tmux split-window` with the given args,
 // retrying up to 3 times with a 250ms pause between attempts. tmux
 // occasionally fails a split mid-layout under rapid pane churn — a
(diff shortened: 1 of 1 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary:** Fix looks correct, minimal, and complete. No blocking issues.

**Verified:**
- `sleep 2147483647` is portable: POSIX `sleep` only guarantees a plain non-negative integer operand, and GNU, macOS/BSD, and busybox implementations all accept it. The value (INT_MAX seconds, ~68 years) comfortably outlives the Phase-3 respawn, so Linux behavior is unchanged.
- Both `sleep infinity` call sites in `restoreGroupCmd` were replaced, and a grep of the head commit confirms no other `sleep infinity` usages remain in the codebase.
- The named constant with a why-comment (`placeholderSleepSecs`) is well-placed next to its consumers and matches the file's existing comment conventions. Keeping it a string constant is right for the `exec` argv usage.
- `go build ./...`, `go vet ./internal/tui/`, and `go test ./internal/tui/` all pass on the PR head.

**Non-blocking observations:**
1. **Test coverage / regression risk:** this bug class (GNU-ism in a shelled-out command) is invisible to Linux CI, and the fix ships with no automated coverage — the manual macOS pass in the PR description is the only guard. Not actionable within this diff given the tmux dependency, but if macOS breakage recurs, a macOS runner exercising the group restore path in `integration/` would be the durable fix.
2. **Optional hardening:** the placeholder panes' survival still depends on the placeholder command not exiting. Setting `remain-on-exit` on the placeholder panes (and clearing it after respawn) would make Phase 1 immune to the placeholder command failing for any reason, not just this one. The current approach is simpler and fine as-is — just noting the alternative if this area grows.

Nice, thorough PR description and repro notes — made this easy to review.

PR #236, opened 2026-07-21
fix(devcontainer): mount the Docker VM's SSH agent path on macOS

shawn-vincent-well:
Fixes #234

## What

sshUpArgs bind-mounts the host's $SSH_AUTH_SOCK into new containers. On macOS that is a launchd socket under /private/tmp/com.apple.launchd.*/, which exists on the mac side but is invisible to Docker Desktop's Linux VM, so 'devcontainer up' fails with "bind source path does not exist" and instance creation never succeeds on a Mac.

Docker Desktop exposes the host's SSH agent inside the VM at a fixed path, /run/host-services/ssh-auth.sock (OrbStack is path-compatible). On darwin this change keeps the existing "is an agent actually running" gate on $SSH_AUTH_SOCK, but uses the VM-side path as the bind source. Linux behavior is unchanged.

Also fixes two ssh_test.go cases that failed on macOS independent of this bug: they bound unix sockets at t.TempDir() paths, which on macOS live under /var/folders/... and exceed the platform's 104-byte sun_path limit. A shortSocketPath helper creates them under /tmp instead, and the mount assertion is now platform-aware.

## Testing

- On macOS (Docker Desktop, arm64): docker run with the VM-side mount succeeds and the agent socket is live inside the container; end-to-end 'fleet up' of a devcontainer instance now works (it failed with the mount error before).
- go test ./internal/backend/devcontainer/ passes on macOS (both pre-existing failures gone).

### DIFF
--- a/internal/backend/devcontainer/ssh.go
+++ b/internal/backend/devcontainer/ssh.go
@@ -90,10 +90,12 @@ func sshUpArgs() []string {
 }
 
 // sshExecArgs returns additional devcontainer exec arguments to set
-// SSH_AUTH_SOCK inside the container. Returns nil if SSH_AUTH_SOCK
-// is not set on the host.
+// SSH_AUTH_SOCK inside the container. It uses the same gate as the mount
+// (sshAgentMountSource) so exec sessions export the variable exactly when
+// the socket is actually mounted — e.g. FLEET_SSH_AGENT_SOCK=off suppresses
+// both, and an override path with no host agent enables both.
 func sshExecArgs() []string {
-	if os.Getenv("SSH_AUTH_SOCK") == "" {
+	if sshAgentMountSource() == "" {
 		return nil
 	}
 	return []string{
--- a/internal/backend/devcontainer/ssh.go
+++ b/internal/backend/devcontainer/ssh.go
@@ -1,11 +1,21 @@
 package devcontainer
 
-import "os"
+import (
+	"os"
+	"runtime"
+)
 
 // containerSSHSocketPath is the target path for the SSH agent socket inside
 // managed containers. Uses /run instead of /tmp because some devcontainer
 // features (e.g. docker-in-docker) mount a tmpfs on /tmp that shadows bind mounts.
 const containerSSHSocketPath = "/run/ssh-agent.sock"
+
+// dockerDesktopSSHAuthSock is the fixed path at which Docker Desktop (and
+// OrbStack, which is path-compatible) exposes the host's SSH agent inside its
+// Linux VM. On macOS the host's real SSH_AUTH_SOCK (a launchd socket under
+// /private/tmp) is not visible to the VM, so bind-mounting it fails with
+// "bind source path does not exist" — this VM-side path must be used instead.
+const dockerDesktopSSHAuthSock = "/run/host-services/ssh-auth.sock"
 
 // hostSSHAuthSock returns the host's SSH_AUTH_SOCK path if the environment
 // variable is set and the socket exists. Returns empty string otherwise.
@@ -31,6 +41,12 @@ func sshUpArgs() []string {
 	sock := hostSSHAuthSock()
 	if sock == "" {
 		return nil
+	}
+	// The macOS agent socket only exists on the mac side; the Docker VM
+	// exposes the agent at its own fixed path. The host check above still
+	// gates on an agent actually running.
+	if runtime.GOOS == "darwin" {
+		sock = dockerDesktopSSHAuthSock
 	}
 	return []string{
 		"--mount", "type=bind,source=" + sock + ",target=" + containerSSHSocketPath,
--- a/internal/backend/devcontainer/ssh.go
+++ b/internal/backend/devcontainer/ssh.go
@@ -11,11 +11,22 @@ import (
 const containerSSHSocketPath = "/run/ssh-agent.sock"
 
 // dockerDesktopSSHAuthSock is the fixed path at which Docker Desktop (and
-// OrbStack, which is path-compatible) exposes the host's SSH agent inside its
-// Linux VM. On macOS the host's real SSH_AUTH_SOCK (a launchd socket under
-// /private/tmp) is not visible to the VM, so bind-mounting it fails with
-// "bind source path does not exist" — this VM-side path must be used instead.
+// OrbStack, which is path-compatible; likewise `colima --ssh-agent`) exposes
+// the host's SSH agent inside its Linux VM. On macOS the host's real
+// SSH_AUTH_SOCK (a launchd socket under /private/tmp) is not visible to the
+// VM, so bind-mounting it fails with "bind source path does not exist" — this
+// VM-side path must be used instead. Two caveats, both escapable via
+// FLEET_SSH_AGENT_SOCK: default Colima, Podman machine, and Rancher Desktop
+// don't provide this path; and it forwards the agent Docker Desktop itself
+// sees (the default launchd one), which may differ from a custom
+// SSH_AUTH_SOCK agent such as 1Password's.
 const dockerDesktopSSHAuthSock = "/run/host-services/ssh-auth.sock"
+
+// sshAgentSockOverrideEnv overrides the bind source used for SSH agent
+// forwarding into instances: a path replaces the computed source verbatim
+// (no host-side existence check — the path may only exist inside the Docker
+// VM), and "off" or "none" disables the agent mount entirely.
+const sshAgentSockOverrideEnv = "FLEET_SSH_AGENT_SOCK"
 
 // hostSSHAuthSock returns the host's SSH_AUTH_SOCK path if the environment
 // variable is set and the socket exists. Returns empty string otherwise.
@@ -34,19 +45,43 @@ func hostSSHAuthSock() string {
 	return sock
 }
 
+// sshAgentMountSource returns the bind source to use for the SSH agent
+// socket mount, or "" when agent forwarding should be skipped. Shared by
+// devcontainer up and clone so every container-creation path applies the
+// same gate, darwin remap, and override.
+func sshAgentMountSource() string {
+	return sshAgentMountSourceFor(os.Getenv(sshAgentSockOverrideEnv), hostSSHAuthSock(), runtime.GOOS)
+}
+
+// sshAgentMountSourceFor is the pure core of sshAgentMountSource, split out
+// so both the linux and darwin branches are testable on any platform.
+func sshAgentMountSourceFor(override, hostSock, goos string) string {
+	switch override {
+	case "off", "none":
+		return ""
+	case "":
+	default:
+		return override
+	}
+	if hostSock == "" {
+		return ""
+	}
+	// The macOS agent socket only exists on the mac side; the Docker VM
+	// exposes the agent at its own fixed path. The hostSock gate still
+	// requires an agent to actually be running.
+	if goos == "darwin" {
+		return dockerDesktopSSHAuthSock
+	}
+	return hostSock
+}
+
 // sshUpArgs returns additional devcontainer up arguments to bind-mount the
 // host SSH agent socket and set SSH_AUTH_SOCK inside the container.
 // Returns nil if SSH agent forwarding is not available.
 func sshUpArgs() []string {
-	sock := hostSSHAuthSock()
+	sock := sshAgentMountSource()
 	if sock == "" {
 		return nil
-	}
-	// The macOS agent socket only exists on the mac side; the Docker VM
-	// exposes the agent at its own fixed path. The host check above still
-	// gates on an agent actually running.
-	if runtime.GOOS == "darwin" {
-		sock = dockerDesktopSSHAuthSock
 	}
 	return []string{
 		"--mount", "type=bind,source=" + sock + ",target=" + containerSSHSocketPath,
--- a/internal/backend/devcontainer/ssh_test.go
+++ b/internal/backend/devcontainer/ssh_test.go
@@ -132,13 +132,29 @@ func TestSSHUpArgs_OverrideOff(t *testing.T) {
 
 func TestSSHUpArgs_NoSocket(t *testing.T) {
 	t.Setenv("SSH_AUTH_SOCK", "")
+	t.Setenv(sshAgentSockOverrideEnv, "")
 	if args := sshUpArgs(); args != nil {
 		t.Errorf("expected nil, got %v", args)
 	}
 }
 
-func TestSSHExecArgs_WithEnvVar(t *testing.T) {
-	t.Setenv("SSH_AUTH_SOCK", "/some/path")
+// liveAgentSocket binds a real unix socket and points SSH_AUTH_SOCK at it,
+// with the override cleared — sshExecArgs shares sshUpArgs's gate, which
+// stats the socket, so a bare env var is no longer enough.
+func liveAgentSocket(t *testing.T) {
+	t.Helper()
+	sockPath := shortSocketPath(t)
+	ln, err := net.Listen("unix", sockPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { ln.Close() })
+	t.Setenv("SSH_AUTH_SOCK", sockPath)
+	t.Setenv(sshAgentSockOverrideEnv, "")
+}
+
+func TestSSHExecArgs_WithAgent(t *testing.T) {
+	liveAgentSocket(t)
 	args := sshExecArgs()
 	if len(args) != 2 {
 		t.Fatalf("expected 2 args, got %d: %v", len(args), args)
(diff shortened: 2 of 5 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Summary:** The fix is correct for #234 — on Docker Desktop/OrbStack the VM-side `/run/host-services/ssh-auth.sock` is the right bind source on macOS, keeping the host-side agent gate is sensible, and Linux behavior is untouched. The `sun_path`-limit test fixes are legitimate and self-contained. A few concerns, most notably one more call site with the same bug:

**1. `Clone` still bind-mounts the raw host path — same failure on macOS** (`internal/backend/devcontainer/clone.go:304`)
`buildCloneRunArgs` emits `--mount type=bind,source=<hostSSHAuthSock()>` for cloned instances. On macOS that's the same mac-side launchd path this PR fixes for `devcontainer up`, so cloning an instance on a Mac will still fail (or mount a dead socket). Consider centralizing the gate + darwin remap in one shared helper used by both `sshUpArgs` and `buildCloneRunArgs`, rather than inlining the remap in `sshUpArgs` only.

**2. The darwin branch is never exercised in CI**
Unit tests run on ubuntu only, and the Windows integration job runs inside WSL — so the new branch (and its test assertion) only ever executes on a contributor's Mac. The test also reproduces the implementation's own `runtime.GOOS == "darwin"` conditional, so it can't catch a regression in the branch logic. Suggested fix: parameterize the OS (`sshUpArgsForOS(goos string)`, with `sshUpArgs()` a thin wrapper passing `runtime.GOOS`) and add explicit tests for both `"linux"` and `"darwin"` that run on every platform.

**3. darwin ≠ Docker Desktop**
The remap keys off the OS, but the fixed path is a Docker Desktop convention (OrbStack and `colima --ssh-agent` are path-compatible; default Colima, Podman machine, and Rancher Desktop are not). Since launchd sets `SSH_AUTH_SOCK` in every macOS session, the gate essentially always passes, so users on those backends will hit a hard `up` failure referencing a path they never configured. An escape hatch (e.g. an override/opt-out env var) or at least a README note would soften this. Not blocking — those setups were equally broken before this PR.

**Minor**
- On darwin the gate validates the *user's* agent (`$SSH_AUTH_SOCK`, e.g. a 1Password agent), but `/run/host-services/ssh-auth.sock` forwards the agent Docker Desktop itself sees (typically the default launchd one). A user on a custom agent passes the gate yet may get a different key set inside the container — worth a note in the `dockerDesktopSSHAuthSock` comment.
- `shortSocketPath` hardcoding `/tmp` is fine given unit CI is Linux-only; the platform-aware mount assertion is correct as written.

Nice, focused PR overall — the comments explain *why* rather than *what*, consistent with the codebase's conventions.

internal/backend/devcontainer/ssh.go
```diff
@@ -32,6 +42,12 @@ func sshUpArgs() []string {
 	if sock == "" {
 		return nil
 	}
+	// The macOS agent socket only exists on the mac side; the Docker VM
+	// exposes the agent at its own fixed path. The host check above still
+	// gates on an agent actually running.
+	if runtime.GOOS == "darwin" {
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This remap is also needed at the other SSH-agent bind-mount call site: `buildCloneRunArgs` (clone.go:304) still uses `hostSSHAuthSock()` directly as the bind source, so `Clone` on macOS keeps failing exactly the way `up` did. Suggest extracting a shared helper (gate + darwin remap) and using it at both call sites instead of inlining the remap here.

internal/backend/devcontainer/ssh_test.go
```diff
@@ -63,7 +77,11 @@ func TestSSHUpArgs_WithValidSocket(t *testing.T) {
 	if args[0] != "--mount" {
 		t.Errorf("args[0] = %q, want --mount", args[0])
 	}
-	wantMount := "type=bind,source=" + sockPath + ",target=" + containerSSHSocketPath
+	wantSource := sockPath
+	if runtime.GOOS == "darwin" {
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This assertion recomputes the implementation's own `runtime.GOOS` conditional, so it cannot fail if the branch logic regresses — and CI (ubuntu unit tests, WSL-based Windows integration) never runs the darwin side at all. Consider parameterizing the OS — `sshUpArgsForOS(goos string)` with `sshUpArgs()` as a thin wrapper passing `runtime.GOOS` — so both the linux and darwin paths get explicit tests that run on every platform.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Re-review of 1b6794f.** All four findings from the previous review are properly addressed:

- ✅ **Clone path**: `buildCloneRunArgs` and `sshUpArgs` now share `sshAgentMountSource()`, so the gate + darwin remap + override apply to every container-creation path.
- ✅ **CI-testable darwin branch**: the pure core `sshAgentMountSourceFor(override, hostSock, goos)` with table tests covers both branches (and the override/off cases) on every platform.
- ✅ **Non-Docker-Desktop backends**: `FLEET_SSH_AGENT_SOCK` escape hatch with a thorough README entry.
- ✅ **1Password/custom-agent caveat**: documented in the `dockerDesktopSSHAuthSock` comment.

Verified locally: `go test ./internal/backend/devcontainer/` passes on linux.

**Two new issues, both introduced by the override env var:**

**1. `sshExecArgs` still gates on raw `SSH_AUTH_SOCK`, diverging from the new mount gate** (`internal/backend/devcontainer/ssh.go:96`)
Container creation now decides via `sshAgentMountSource()`, but every `devcontainer exec` decides via `os.Getenv("SSH_AUTH_SOCK")`. The two disagree in exactly the cases this PR adds:
- `FLEET_SSH_AGENT_SOCK=off` + host agent running → no socket is mounted, yet every exec session exports `SSH_AUTH_SOCK=/run/ssh-agent.sock` pointing at nothing. The user opted out, but ssh inside the instance now probes a dead socket instead of cleanly seeing no agent.
- `FLEET_SSH_AGENT_SOCK=<path>` + no host `SSH_AUTH_SOCK` (the "override wins without host agent" case your table test explicitly supports) → the socket **is** mounted, but exec sessions never export `SSH_AUTH_SOCK`, so the forwarded agent is invisible.

Suggested fix: gate `sshExecArgs` on `sshAgentMountSource() != ""` so exec matches creation. (Its pre-existing skip of the socket-liveness check would be fixed by the same change.)

**2. Test hermeticity** (`internal/backend/devcontainer/ssh_test.go:134`)
`TestSSHUpArgs_NoSocket` doesn't clear `FLEET_SSH_AGENT_SOCK`. On a machine where it's exported to a path — precisely the Colima/Podman users the README now instructs to export it — the override wins over the empty `SSH_AUTH_SOCK` and the test fails spuriously. `TestSSHUpArgs_WithValidSocket` already clears it; `NoSocket` should too.

Neither is blocking for the #234 fix itself, but #1 is worth fixing in this PR since it's the same feature surface. Inline comments at both spots.

internal/backend/devcontainer/ssh.go
```diff
@@ -24,11 +45,41 @@ func hostSSHAuthSock() string {
...
+	}
+	return hostSock
+}
+
 // sshUpArgs returns additional devcontainer up arguments to bind-mount the
 // host SSH agent socket and set SSH_AUTH_SOCK inside the container.
 // Returns nil if SSH agent forwarding is not available.
 func sshUpArgs() []string {
-	sock := hostSSHAuthSock()
+	sock := sshAgentMountSource()
```
BenjaminBenetti, line 82: 🔍 **CodeReview Agent**

The new gate applied here is not applied in `sshExecArgs` below (line 96), which still checks raw `os.Getenv("SSH_AUTH_SOCK")`. With `FLEET_SSH_AGENT_SOCK=off` exec sessions still export `SSH_AUTH_SOCK=/run/ssh-agent.sock` pointing at an unmounted path; with an override path and no host agent, the mount happens but exec sessions never export the variable. Suggest gating `sshExecArgs` on `sshAgentMountSource() != ""` so exec-time behavior matches creation-time behavior.

internal/backend/devcontainer/ssh_test.go
```diff
@@ -76,6 +115,21 @@ func TestSSHUpArgs_WithValidSocket(t *testing.T) {
...
+
+	t.Setenv("SSH_AUTH_SOCK", sockPath)
+	t.Setenv(sshAgentSockOverrideEnv, "off")
+	if args := sshUpArgs(); args != nil {
+		t.Errorf("expected nil with %s=off, got %v", sshAgentSockOverrideEnv, args)
+	}
+}
+
 func TestSSHUpArgs_NoSocket(t *testing.T) {
 	t.Setenv("SSH_AUTH_SOCK", "")
```
BenjaminBenetti, line 134: 🔍 **CodeReview Agent**

This test is no longer hermetic: if the developer's environment exports `FLEET_SSH_AGENT_SOCK` to a path (the setup the README now recommends for Colima/Podman/Rancher users), the override wins over the empty `SSH_AUTH_SOCK` and `sshUpArgs()` returns non-nil, failing the test spuriously. Add `t.Setenv(sshAgentSockOverrideEnv, "")` as `TestSSHUpArgs_WithValidSocket` does.

internal/backend/devcontainer/ssh.go
shawn-vincent-well: Fixed in 1b6794f: both devcontainer up and buildCloneRunArgs now go through a shared sshAgentMountSource(), so the gate, darwin remap, and FLEET_SSH_AGENT_SOCK override apply to every container-creation path.

internal/backend/devcontainer/ssh_test.go
shawn-vincent-well: Done in 1b6794f: the pure core is sshAgentMountSourceFor(override, hostSock, goos), with table tests covering the linux and darwin branches (plus override and off) that run on every platform.

internal/backend/devcontainer/ssh.go
shawn-vincent-well: Fixed in f148342: sshExecArgs now shares sshAgentMountSource as its gate, so exec sessions export SSH_AUTH_SOCK exactly when the socket is actually mounted (off suppresses both, override-without-host-agent enables both).

internal/backend/devcontainer/ssh_test.go
shawn-vincent-well: Fixed in f148342 and c8b467f: the override env is cleared in every test that assumes default behavior, and the exec / run-user-commands tests bind a live socket since the shared gate stats it.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-review at head `c8b467f` (branch received one more push after the triggering event's `f148342`; reviewed the latest state).

**Prior findings — all verified fixed at head:**

1. ✅ **Clone path** — `buildCloneRunArgs` (clone.go:304) now goes through the shared `sshAgentMountSource()`, so the gate, darwin remap, and override apply to every container-creation path.
2. ✅ **Testability** — pure core `sshAgentMountSourceFor(override, hostSock, goos)` with table tests exercising the linux and darwin branches, override precedence, and `off`/`none` on any platform.
3. ✅ **`sshExecArgs` gate** — now shares `sshAgentMountSource()`, so exec / run-user-commands sessions export `SSH_AUTH_SOCK` exactly when the socket is actually mounted.
4. ✅ **Test hermeticity** — every test relying on default behavior clears `FLEET_SSH_AGENT_SOCK`, and the exec/run-user-commands cases bind a live socket to satisfy the stat in the shared gate.

Independently verified: grep confirms no remaining ungated `SSH_AUTH_SOCK` call sites in the devcontainer backend; `go test ./internal/backend/devcontainer/` passes (linux); `GOOS=darwin GOARCH=arm64` compile of the package is clean. (The codespaces/coder backends read raw `SSH_AUTH_SOCK` but only to enable host-side ssh agent forwarding — unaffected by this bug, correctly left alone.)

**One minor new finding** (inline on ssh.go): `off`/`none` matching is case-sensitive and the override value is otherwise spliced verbatim into the `--mount` string — see comment.

**FYI, non-blocking:** Docker Desktop *for Linux* runs a VM too, with the same host-path invisibility, and exposes the same `/run/host-services/ssh-auth.sock`. The linux branch keeps using the host path there, so those users need the env override — the README's "set this if your Docker backend exposes the agent elsewhere" wording covers it, just flagging that the escape hatch isn't macOS-only.

Overall: solid, well-documented fix with genuinely cross-platform test coverage. No blocking issues.

internal/backend/devcontainer/ssh.go
```diff
@@ -24,11 +45,41 @@ func hostSSHAuthSock() string {
...
+// same gate, darwin remap, and override.
+func sshAgentMountSource() string {
+	return sshAgentMountSourceFor(os.Getenv(sshAgentSockOverrideEnv), hostSSHAuthSock(), runtime.GOOS)
+}
+
+// sshAgentMountSourceFor is the pure core of sshAgentMountSource, split out
+// so both the linux and darwin branches are testable on any platform.
+func sshAgentMountSourceFor(override, hostSock, goos string) string {
+	switch override {
+	case "off", "none":
```
BenjaminBenetti, line 60: 🔍 **CodeReview Agent**

Minor robustness: the `off`/`none` match is exact-case, so `FLEET_SSH_AGENT_SOCK=OFF` falls through to the default branch and is bind-mounted verbatim — `--mount type=bind,source=OFF,...` — failing container creation with a confusing Docker error far from the cause. Likewise any non-absolute value, or one containing a comma (e.g. `/run/foo.sock,readonly`), is silently spliced into the mount string as extra mount options. Consider `strings.EqualFold` for the kill-switch values and rejecting (or warn-and-ignoring) overrides that aren't absolute comma-free paths, so misconfiguration fails at the source rather than inside `devcontainer up`.

internal/backend/devcontainer/ssh.go
```diff
@@ -24,11 +45,41 @@ func hostSSHAuthSock() string {
...
+// devcontainer up and clone so every container-creation path applies the
+// same gate, darwin remap, and override.
+func sshAgentMountSource() string {
+	return sshAgentMountSourceFor(os.Getenv(sshAgentSockOverrideEnv), hostSSHAuthSock(), runtime.GOOS)
+}
+
+// sshAgentMountSourceFor is the pure core of sshAgentMountSource, split out
+// so both the linux and darwin branches are testable on any platform.
+func sshAgentMountSourceFor(override, hostSock, goos string) string {
+	switch override {
```
BenjaminBenetti, line 59: 🔍 **CodeReview Agent**

Minor, non-blocking: the repo's other `FLEET_*` env parsers (`FLEET_DEVCONTAINER_UPDATE_REMOTE_USER_UID` at devcontainer_backend.go:280, `FLEET_DEVCONTAINER_BUILDKIT` at :309) run `strings.TrimSpace` on the value before matching; this one dispatches on the raw string. `FLEET_SSH_AGENT_SOCK="off "` (stray whitespace from a shell rc) would silently fall through to the path branch and become a bind source verbatim, reproducing the exact "bind source path does not exist" failure class this PR fixes — with no invalid-value error to point at the cause, since any non-keyword string is a valid path here. Suggest trimming in `sshAgentMountSource()` before calling the pure core.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-review at head `c8b467f` (new commits since last review: `f148342`, `c8b467f`).

**All four prior findings are verified addressed at head:**

1. ✅ Clone path — `buildCloneRunArgs` (clone.go:304) and `sshUpArgs` both route through the shared `sshAgentMountSource()`, so the gate, darwin remap, and override apply to every container-creation path.
2. ✅ Platform testability — the pure core `sshAgentMountSourceFor(override, hostSock, goos)` has table tests exercising both the linux and darwin branches (plus override and off cases) on every platform.
3. ✅ Exec gate — `sshExecArgs` now shares `sshAgentMountSource()` as its gate, so exec sessions export `SSH_AUTH_SOCK` exactly when the socket is actually mounted.
4. ✅ Test hermeticity — I ran the full package suite with hostile ambient env (`FLEET_SSH_AGENT_SOCK=/vm/elsewhere.sock`, and `SSH_AUTH_SOCK= FLEET_SSH_AGENT_SOCK=off`): all pass, so the override-clearing is complete.

**Delta commits look correct.** `c8b467f` is the right fix for the `run-user-commands` test: `runUserCommandsArgs` → `sshExecArgs` now stats the socket, so a bare env var is no longer sufficient and binding a live agent socket via `liveAgentSocket(t)` is the correct setup. The "without ssh agent" subtest also clears the override, keeping it hermetic.

**Independently verified on this run:** `go test ./internal/backend/devcontainer/`, `go vet`, and a `GOOS=darwin GOARCH=arm64` cross-compile of the package all pass. Also confirmed the hardcoded `/tmp` in `shortSocketPath` is safe for every platform tests actually run on (ubuntu unit CI, WSL-based Windows CI, and darwin dev machines).

**One new minor finding** (inline, non-blocking): `sshAgentMountSourceFor` dispatches on the raw override value without `strings.TrimSpace`, unlike the repo's other `FLEET_*` env parsers — a value with stray whitespace (e.g. `"off "`) silently becomes a bind-source path and reproduces the exact mount failure this PR fixes. One-line fix at the impure call site.

No correctness, security, or error-handling issues found beyond that. Nice round of revisions — the shared-helper structure keeps all three consumers (up, clone, exec) provably in lockstep.

internal/backend/devcontainer/ssh.go
shawn-vincent-well: Fixed in #243: TrimSpace + EqualFold for off/none, and unusable values (relative paths, commas that would splice mount options) are now a hard error naming the env var at create/clone time, matching devcontainerUpArgs / devcontainerEnv.

internal/backend/devcontainer/ssh.go
shawn-vincent-well: Fixed in #243 along with the case-sensitivity finding -- the value is TrimSpace'd before matching, consistent with the other FLEET_* parsers.

PR #239, opened 2026-07-21
feat(ci): macOS CI without Docker - unit job + docker-free integration mode

shawn-vincent-well:
macOS CI as discussed on Slack: unit tests plus a docker-free integration mode on GitHub's hosted macos-latest runners. No self-hosted hardware needed.

Stacked on #236 (branched from it) -- merge that first and this diff shrinks to just the CI work.

## Why no Docker

GitHub's hosted macOS runners are Apple Silicon VMs with nested virtualization disabled, so Docker Desktop / colima / lima cannot run at all. Rather than a self-hosted Mac, this covers the host-side surface where macOS bugs actually live: BSD userland, the 104-byte sun_path limit, tmux, daemon/process management. (True e2e devcontainer-up on a Mac stays manual for now.)

## What's in it

1. `.github/workflows/macos.yml` -- two jobs on macos-latest: `unit-macos` (go vet + full go test, with tmux installed so the cli session tests actually run) and `integration-macos-no-docker`.

2. `FLEET_ITEST_NO_DOCKER=1` mode in `integration/run.sh`: skips the docker/devcontainer preflight and runs only tests marked `# itest: no-docker`. The marker is opt-in, so a test that quietly grows a docker dependency fails loudly in review instead of silently joining the set. run.sh is also bash-3.2 clean now (macOS ships bash 3.2; `declare -A` was a hard error).

3. Markers on the 11 docker-free tests: 070 cli help, 170-173 new-fleet dialogs, 230 fleet-delete-empty, 392 automation CRUD, and the 610-613 daemon lifecycle suite. (391 spawns a real devcontainer instance via the scheduler, so it stays docker-only.)

4. Portable `server_count` in common.sh: macOS pgrep has no `-c` flag, so the six per-test `pgrep -fc` copies printed nothing there and every daemon-count assertion failed. Replaced with a shared `pgrep -f | wc -l` form that behaves identically on Linux.

5. Unit-test fixes for macOS, all the same disease (104-byte sun_path limit on unix sockets under /var/folders/... TempDirs): tmux server sockets in internal/cli, control sockets under $HOME in internal/server, and direct binds in internal/control now use short /tmp dirs. The cli tmux env also points HOME at the tmux dir so zsh exit-time writes can't race TempDir cleanup (was a flake).

## Testing

- Full `go test ./...` green on macOS arm64 (with #236 merged; without it, only the one test #236 fixes is red).
- The no-docker subset passes 11/11 on macOS with docker *blocked* via a failing PATH shim, proving no hidden docker dependency.
- Linux behavior unchanged: preflight, sharding, FLEET_ITEST_SKIP, and the pgrep form all behave identically; existing ubuntu jobs untouched.

### DIFF
--- a/integration/run.sh
+++ b/integration/run.sh
@@ -32,31 +32,40 @@ say "Using fleet binary: ${FLEET_BIN}"
 "${FLEET_BIN}" --help >/dev/null || { printf "%b[fatal]%b fleet binary failed to run\n" "${RED}" "${RESET}" >&2; exit 2; }
 
 # 2. Preflight: docker + devcontainer CLI must be available and usable.
-if ! command -v docker >/dev/null 2>&1; then
-  printf "%b[fatal]%b docker CLI not on PATH\n" "${RED}" "${RESET}" >&2
-  exit 2
-fi
-if ! docker info >/dev/null 2>&1; then
-  printf "%b[fatal]%b docker daemon not reachable\n" "${RED}" "${RESET}" >&2
-  exit 2
-fi
-if ! command -v devcontainer >/dev/null 2>&1; then
-  printf "%b[fatal]%b devcontainer CLI not on PATH (npm i -g @devcontainers/cli)\n" "${RED}" "${RESET}" >&2
-  exit 2
-fi
-
-# 3. Pre-pull each fixture image once so individual tests are not waiting
-#    on image pulls. The agent fixture is used by the startup-script tests
-#    (claude / codex install) and is a different image from the default.
-for fixture_dir in "${INTEGRATION_DIR}/fixture" "${INTEGRATION_DIR}/fixture-agent"; do
-  dc_json="${fixture_dir}/.devcontainer/devcontainer.json"
-  [ -f "${dc_json}" ] || continue
-  fixture_image=$(grep -oE '"image":\s*"[^"]+"' "${dc_json}" | sed -E 's/.*"([^"]+)"$/\1/')
-  if [ -n "${fixture_image}" ]; then
-    say "Pre-pulling fixture image: ${fixture_image}"
-    docker pull -q "${fixture_image}" >/dev/null || true
+#    FLEET_ITEST_NO_DOCKER=1 skips the preflight and (below) restricts the
+#    run to tests marked "# itest: no-docker" — the host-side subset (CLI
+#    surface, TUI dialogs, daemon lifecycle, automation CRUD) that never
+#    creates an instance. This is what lets the suite run on GitHub's
+#    hosted macOS runners, whose Apple Silicon VMs cannot run Docker.
+if [ -z "${FLEET_ITEST_NO_DOCKER:-}" ]; then
+  if ! command -v docker >/dev/null 2>&1; then
+    printf "%b[fatal]%b docker CLI not on PATH\n" "${RED}" "${RESET}" >&2
+    exit 2
   fi
-done
+  if ! docker info >/dev/null 2>&1; then
+    printf "%b[fatal]%b docker daemon not reachable\n" "${RED}" "${RESET}" >&2
+    exit 2
+  fi
+  if ! command -v devcontainer >/dev/null 2>&1; then
+    printf "%b[fatal]%b devcontainer CLI not on PATH (npm i -g @devcontainers/cli)\n" "${RED}" "${RESET}" >&2
+    exit 2
+  fi
+
+  # 3. Pre-pull each fixture image once so individual tests are not waiting
+  #    on image pulls. The agent fixture is used by the startup-script tests
+  #    (claude / codex install) and is a different image from the default.
+  for fixture_dir in "${INTEGRATION_DIR}/fixture" "${INTEGRATION_DIR}/fixture-agent"; do
+    dc_json="${fixture_dir}/.devcontainer/devcontainer.json"
+    [ -f "${dc_json}" ] || continue
+    fixture_image=$(grep -oE '"image":\s*"[^"]+"' "${dc_json}" | sed -E 's/.*"([^"]+)"$/\1/')
+    if [ -n "${fixture_image}" ]; then
+      say "Pre-pulling fixture image: ${fixture_image}"
+      docker pull -q "${fixture_image}" >/dev/null || true
+    fi
+  done
+else
+  say "FLEET_ITEST_NO_DOCKER set: skipping docker preflight, running docker-free tests only"
+fi
 
 # 4. Iterate tests.
 shopt -s nullglob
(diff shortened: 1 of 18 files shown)


integration/run.sh
```diff
@@ -90,6 +99,24 @@ if [ -n "${FLEET_ITEST_RANGE:-}" ]; then
...
+# marker line. Opt-in (not inference) so a test that quietly grows a docker
+# dependency fails loudly in review rather than silently joining this set.
+if [ -n "${FLEET_ITEST_NO_DOCKER:-}" ]; then
+  nodocker=()
+  for test_file in "${tests[@]}"; do
+    if grep -q "^# itest: no-docker" "${test_file}"; then
+      nodocker+=("${test_file}")
+    fi
+  done
+  tests=("${nodocker[@]}")
```
BenjaminBenetti: 🔍 **CodeReview Agent**

**bash 3.2 + `set -u`: empty-array expansion crashes before the graceful exit below.**

In bash < 4.4, expanding an empty array with `"${nodocker[@]}"` under `set -u` is an `unbound variable` error — and macOS `/bin/bash` 3.2 is exactly the interpreter this mode adds. If no tests carry the marker (e.g. `FLEET_ITEST_NO_DOCKER=1` combined with a `FLEET_ITEST_RANGE` shard that contains no marked tests), the script dies with `nodocker[@]: unbound variable` on this line instead of reaching the intended `exit 0` two lines down.

Length expansion is safe on 3.2, so checking before assigning fixes it:

```bash
say "No-docker subset: running ${#nodocker[@]} test(s)"
if [ "${#nodocker[@]}" -eq 0 ]; then
  say "No no-docker tests in shard — nothing to do"
  exit 0
fi
tests=("${nodocker[@]}")
```

Two other spots share the latent pattern if they ever run under bash 3.2, worth fixing while you're here: `tests=("${shard[@]}")` (line 98, empty shard + no-docker on macOS) and `for row in "${result_rows[@]}"` (line 210, reachable only if every test crashes before emitting a row — it would mask the "crashed before itest_begin" report with an unbound-variable error).

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed at head `7a9c083`. Solid PR overall — the opt-in `# itest: no-docker` marker is the right design (a test that grows a docker dependency fails loudly in review), and the sun_path/`/tmp` fixes are well-commented with the *why* at each site. One real bug found, plus two minor notes.

### Bug (1 inline comment)

- **`integration/run.sh:112`** — `tests=("${nodocker[@]}")` with an empty `nodocker` is an `unbound variable` error under `set -u` on bash < 4.4, i.e. exactly the macOS bash 3.2 this PR targets. It defeats the graceful "nothing to do" `exit 0` two lines later (reachable via `FLEET_ITEST_NO_DOCKER` + a shard with no marked tests). Fix is small — check `${#nodocker[@]}` before assigning; details and two sibling spots in the inline comment.

### Minor

- **`integration/run.sh:108`** — `grep -q "^# itest: no-docker"` is a prefix match, so a hypothetical `# itest: no-docker-foo` marker would silently join the set. Anchoring it (`^# itest: no-docker[[:space:]]*$`) keeps the opt-in strict, which is the point of the marker.
- **`internal/backend/devcontainer/ssh.go`** — on darwin, any running host agent now yields the Docker Desktop VM path unconditionally; on backends without it (default Colima, Podman machine, Rancher Desktop) container create fails with docker's generic `bind source path does not exist`. That's not a regression (the old mac-side path was equally invisible to the VM) and the README documents the escape hatch, but the failure doesn't mention `FLEET_SSH_AGENT_SOCK`. A follow-up that remaps or hints at the env var when that specific mount fails would save the next Colima user a debugging session. Non-blocking.

### Verified

- `go vet` + `go test` green on the four touched packages (Linux); imports/compile all check out.
- Scanned `common.sh` and all 11 marked tests for bash-4isms (`declare -A`, `mapfile`, `${var,,}`, `local -n`, …): clean apart from the run.sh issue above. Docker references in the marked tests are comments only; teardown's docker sweeps are `command -v docker`-guarded.
- `server_count` portability swap is behavior-identical on Linux (procps `pgrep -c` prints `0`/exits 1 on no match → same `0` output; `|| true` correctly absorbs the pipefail). No stray `pgrep -fc` remains outside a comment.
- `FLEET_TUI_WAIT_SCALE` exists in `common.sh` and mirrors the WSL job's precedent; workflow triggers and action versions (`checkout@v5`, `setup-go@v6`) match the existing unit/integration workflows.
- `sshExecArgs` now sharing the mount-source gate is a behavior change on Linux too (env var set but socket dead → no longer exports `SSH_AUTH_SOCK` into exec) — looks intentional and strictly more consistent; test coverage for it is good.

Note: PR is stacked on #236, so I focused on the CI/no-docker/ssh work described as this PR's own diff.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-reviewed at new head `9d6ac52`. This push is a pure rebase — every file in the diff is byte-identical to the previously reviewed `7a9c083`. The diff has shrunk to just the 18 CI/test files as the description promised (the #236 ssh work is out of this PR now, so the earlier `ssh.go` follow-up note travels with #236, not this one). Re-verified the full diff at this head; no new issues introduced.

### Still open from the previous review

- **`integration/run.sh:112` (bug)** — `tests=("${nodocker[@]}")` with an empty `nodocker` is an `unbound variable` error under `set -u` on bash 3.2, defeating the graceful `exit 0` two lines below (reachable via `FLEET_ITEST_NO_DOCKER` + a shard with no marked tests). The existing inline thread has the small fix plus the two sibling spots (lines 98 and 210). This is the one correctness item worth fixing before undraft.
- **`integration/run.sh:108` (minor)** — the `# itest: no-docker` marker grep is a prefix match; anchoring keeps the opt-in strict.

### New minor note (not previously flagged)

- **`internal/cli/session_commands_test.go:96`** — `t.Cleanup(func() { os.RemoveAll(tmuxDir) })` discards the removal error, which is part of what makes the zsh exit-time race benign (with `t.TempDir()` that race failed the test). The flip side: if a straggler write lands mid-removal, the `/tmp/fmtmux*` dir leaks silently instead of failing loudly. Fine on ephemeral CI runners and a reasonable trade for de-flaking; just noting it's a deliberate silent-leak-over-flake choice in case you'd rather log the error. Non-blocking.

### Verified at this head

- Old→new head diff is empty for all 18 files (rebase only); base repo conventions still match (`checkout@v5`/`setup-go@v6`/`go-version-file`, PR-only triggers, no `-race` anywhere, so the macOS unit job is consistent — and omitting the golangci-lint import-boundary step there is right, it's platform-independent and already runs on ubuntu).
- `os` is imported in both `internal/control/control_test.go` and `internal/server/control_test.go`, so the new `os.MkdirTemp` call sites compile; `FLEET_TUI_WAIT_SCALE` exists in `common.sh` (default 1) and mirrors the WSL job's precedent.
- `server_count` in `common.sh` remains behavior-identical on Linux and correctly absorbs pgrep's exit-1-on-zero under pipefail; all 11 marked tests carry the marker, 391 correctly stays docker-only.

One cosmetic thing: the PR description still says "Stacked on #236 -- merge that first" — now stale, worth a quick edit when undrafting.

integration/run.sh
shawn-vincent-well: Fixed in 82b622a: the length check now happens before the expansion and the empty-subset exit-0 comes first. Also applied the same guard to the FLEET_ITEST_RANGE shard filter, whose empty case is documented as exit-0 but previously fell through (and would have hit the same bash-3.2 crash). Verified the empty-shard path exits 0 under macOS /bin/bash 3.2.

PR #240, opened 2026-07-21
feat(ci): BSD-userland shim job - full suite on Linux with macOS command semantics

shawn-vincent-well:
Layer 3 from the Slack thread: the full integration suite on cheap ubuntu runners, with macOS/BSD command behavior fronting the host PATH. Catches the #231 bug class (GNU-isms in host-side shell-outs, e.g. `sleep infinity`) across all 85 tests, zero Mac hardware. Credit where due: the technique is your QA agent's -- it validated the #231 fix with exactly this kind of sleep shim.

Stacked on #239 (which stacks on #236); merge those first and this diff shrinks to just the shim work. Draft because of the open-PR limit.

## How it works

- `integration/bsd-shims/` holds four fakes: `sleep` (rejects non-numeric operands), `stat` (rejects -c/--format), `date` (rejects -d/--date), `timeout` (always command-not-found -- macOS ships none). Each prints a BSD-style usage error for GNU-only usage, otherwise delegates to the next real binary on PATH.
- `FLEET_ITEST_BSD_SHIMS=1` in run.sh prepends the shim dir to PATH. Host shell-outs (fleet's exec.Command calls, the harness, the tests) see BSD semantics; everything inside containers keeps its real Linux userland, mirroring a real Mac + Docker Desktop split.
- `.github/workflows/integration-bsd.yml` mirrors the stock integration workflow: same 7-shard matrix, same aggregate gate (stable `integration-bsd` status name).

## First catch

025_copy asserted host-side file modes with `stat -c '%a'`, which BSD stat does not have (it spells it `-f '%Lp'`). That would have failed on any real Mac. Fixed with a portable `file_mode` helper in common.sh; the in-container `stat -c` call in the same test is correct as-is (containers are Linux) and untouched.

## Testing

- Shim unit behavior: `sleep infinity` exits 1 with BSD usage, numeric sleeps pass through; `stat -c` / `date -d` / `timeout` all reject as macOS would.
- On macOS: the docker-free subset passes 11/11 under shims, and the docker-backed 010 (up + ls) and 025 (copy, exercising the new file_mode) pass end-to-end under shims in a fresh sandbox HOME.
- Doubling of integration cost per PR is real (a second 7-shard matrix). If that is too spendy, easy dial-downs: run this workflow on a schedule instead of per-PR, or only when integration/ or internal/tui/internal/cli change. Happy to adjust.

One unrelated find from testing on my Mac, filed separately: Docker Desktop's VirtioFS intermittently reports "bind source path does not exist" for workspace paths that were just rm -rf'd and recreated (the suite's teardown/setup cycle). Purely macOS-local, cannot affect these ubuntu jobs.

### DIFF
--- a/integration/bsd-shims/stat
+++ b/integration/bsd-shims/stat
@@ -1,0 +1,46 @@
+#!/bin/sh
+# BSD stat has no -c/--format/--printf (it spells formatting -f).
+for arg in "$@"; do
+  case "$arg" in
+    -c|--format|--format=*|--printf|--printf=*)
+      echo "stat: illegal option -- c" >&2
+      echo "usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file ...]" >&2
+      exit 1
+      ;;
+    --) break ;;
+  esac
+done
+
+self_dir=$(cd "$(dirname "$0")" && pwd)
+real=""
+old_ifs=$IFS; IFS=:
+for d in $PATH; do
+  [ "$d" = "$self_dir" ] && continue
+  [ -x "$d/stat" ] && { real="$d/stat"; break; }
+done
+IFS=$old_ifs
+[ -n "$real" ] || { echo "stat: real binary not found" >&2; exit 1; }
+
+# Emulate the BSD spelling instead of delegating it: on the real GNU stat
+# behind this shim, -f means --file-system and the format string would be
+# read as a file operand. Without this, portable "try GNU, fall back to
+# BSD" probes have NO working spelling under shims — a false positive
+# relative to a real Mac. Translate the formats the suite needs; fail
+# loudly on the rest so new uses extend the map instead of silently
+# hitting GNU semantics.
+if [ "${1:-}" = "-f" ] && [ $# -ge 2 ] && "$real" --version 2>/dev/null | grep -q GNU; then
+  fmt="$2"
+  shift 2
+  case "$fmt" in
+    "%Lp") exec "$real" -c '%a' "$@" ;;
+    "%z")  exec "$real" -c '%s' "$@" ;;
+    *)
+      echo "bsd-shims/stat: BSD -f format '$fmt' not emulated; add it to the shim" >&2
+      exit 1
+      ;;
+  esac
+fi
+# (a real BSD stat behind the shim — e.g. running the shims on an actual
+# Mac — handles -f natively, so it falls through to plain delegation)
+
+exec "$real" "$@"
--- a/integration/common.sh
+++ b/integration/common.sh
@@ -601,3 +601,12 @@ tui_assert_not_contains() {
     fail "${msg}: needle=[${needle}]"
   fi
 }
+
+# server_count — number of live `fleet server` daemons for this suite's
+# binary. Shared here because macOS pgrep has no -c flag (the old per-test
+# `pgrep -fc` printed nothing there and every count assertion failed); the
+# pipe form behaves identically on Linux and macOS. `|| true` absorbs
+# pgrep's exit 1 on zero matches under pipefail.
+server_count() {
+  { pgrep -f "${FLEET_BIN} server" 2>/dev/null || true; } | wc -l | tr -d ' '
+}
--- a/integration/common.sh
+++ b/integration/common.sh
@@ -610,3 +610,13 @@ tui_assert_not_contains() {
 server_count() {
   { pgrep -f "${FLEET_BIN} server" 2>/dev/null || true; } | wc -l | tr -d ' '
 }
+
+# file_mode <path> — octal permission bits of a HOST file, portably. Not
+# stat: GNU spells it -c '%a' and BSD -f '%Lp', and no probe order survives
+# the BSD-shim job — the shim rejects -c like BSD, but a -f fallback lands
+# on the real GNU stat where -f means file*system* status. perl's stat has
+# one meaning everywhere. (In-container stat calls stay plain `stat -c` —
+# containers are Linux regardless of the host.)
+file_mode() {
+  perl -e 'printf "%o", (stat shift)[2] & 07777' "$1"
+}
--- a/.github/workflows/integration-bsd.yml
+++ b/.github/workflows/integration-bsd.yml
@@ -1,0 +1,82 @@
+name: Integration Tests (BSD userland shims)
+
+# The full integration suite on cheap Linux runners, but with macOS/BSD
+# command behavior fronting the host PATH (integration/bsd-shims). This
+# catches the most common class of macOS breakage — GNU-isms in host-side
+# shell-outs, e.g. the `sleep infinity` placeholder that broke every
+# session-group switch on a Mac (#231) — across all tests, with no Mac
+# hardware. Docker and the containers keep their real Linux userland;
+# only what fleet and the harness run on the HOST sees BSD semantics.
+#
+# The technique is borrowed from the repo's QA bot, which validated the
+# #231 fix on Linux with exactly this kind of `sleep` shim.
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  integration-bsd-shard:
+    name: integration-bsd ${{ matrix.range }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        range:
+          - "0-100"
+          - "100-200"
+          - "200-300"
+          - "300-400"
+          - "400-500"
+          - "500-600"
+          - "600-"
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - uses: actions/setup-node@v5
+        with:
+          node-version: "20"
+
+      - name: Install devcontainer CLI
+        run: npm install -g @devcontainers/cli
+
+      - name: Install tmux (used by TUI tests)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y tmux
+          tmux -V
+
+      - name: Verify docker
+        run: |
+          docker version
+          docker info
+
+      - name: Build fleet binary
+        run: go build -o /usr/local/bin/fleet ./cmd/fleet
+
+      - name: Run integration tests under BSD shims
+        env:
+          FLEET_BIN: /usr/local/bin/fleet
+          FLEET_ITEST_RANGE: ${{ matrix.range }}
+          FLEET_ITEST_BSD_SHIMS: "1"
+        run: ./integration/run.sh
+
+  integration-bsd:
+    name: integration-bsd
+    runs-on: ubuntu-latest
+    needs: [integration-bsd-shard]
+    if: ${{ always() }}
+    steps:
+      - name: Verify all shards passed
+        run: |
+          result="${{ needs.integration-bsd-shard.result }}"
+          echo "integration-bsd shard matrix result: ${result}"
+          if [ "${result}" != "success" ]; then
+            echo "::error::one or more integration-bsd shards failed (result=${result})"
+            exit 1
+          fi
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,71 @@
+name: macOS
+
+# macOS coverage without macOS Docker: GitHub's hosted macOS runners are
+# Apple Silicon VMs with nested virtualization disabled, so Docker (Desktop,
+# colima, lima) cannot run at all. Instead of a self-hosted Mac, this
+# workflow covers the host-side surface where macOS bugs actually live
+# (BSD userland, sun_path limits, launchd sockets, tmux, process mgmt):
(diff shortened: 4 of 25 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the shim commit (149e4e4); 9d6ac52 is #239's content, so only one note on it at the end. The approach is sound and the harness changes are careful — but there is one confirmed bug, already visible in this PR's own CI.

## Confirmed: `file_mode` fails under the shims (shard 0-100 is red)

`integration-bsd 0-100` on this head failed on `025_copy`:

```
/usr/bin/stat: cannot read file system information for '%Lp': No such file or directory
```

Root cause: `file_mode()` probes GNU first, then falls back to `stat -f '%Lp'`. Under shims on Linux, `-c` is rejected by the shim, and the `-f` fallback is *delegated to GNU stat* — where `-f` means `--file-system` and `%Lp` is parsed as a file operand. So on a shimmed Linux host there is **no working spelling of stat**: the shim rejects the GNU form and the underlying binary can't parse the BSD form. Your macOS runs passed because delegation there reaches real BSD stat; the ubuntu job — the whole point of this PR — is the one place it breaks.

This is an instance of a general gap worth fixing at the shim layer: the shims emulate BSD's **rejections** but not its **acceptances**, so any correctly-written portable fallback ("try GNU, fall back to BSD spelling") fails only under shims — a false positive w.r.t. a real Mac. Recommended fix: teach the stat shim to emulate `-f` by translating the format(s) the suite uses (`%Lp` → `-c '%a'`), and error loudly on untranslated formats. Details in the inline comments; `date` has the same latent issue with `-r`/`-v`.

## Minor

- **`timeout` shim fidelity**: on a real Mac, `command -v timeout` / `exec.LookPath` fails and capability-detecting code takes its fallback branch; under the shim the probe *succeeds* and then execution gets 127. That can only cause false failures (never false passes), but if fleet ever grows a LookPath-gated `timeout` call, this job will fail code that is actually correct on macOS. Worth a line in the README.
- **Shim self-exclusion** compares PATH entries against `$self_dir` as strings. A trailing slash or symlinked checkout in PATH would defeat the skip and the shim would `exec` itself in a loop. Cheap hardening: also skip when `[ "$d/<cmd>" -ef "$0" ]`.
- **Workflow duplication**: `integration-bsd.yml` is a byte-level near-copy of `integration.yml` — shard ranges and setup steps will drift. Consider folding shims into the stock workflow as a matrix dimension (`shims: ["", "1"]`, name suffix + env from the value) so both variants share one definition. On your cost dial-downs: fine either way, but note that if `integration-bsd` becomes a required status check, a `paths:` filter will leave PRs that don't touch those paths stuck on an "Expected" check — schedule-based is the safer dial.

## Looks good

- `server_count` consolidation is correct and genuinely portable (pipe form, `|| true` under pipefail, `tr -d ' '` for BSD wc padding), and removes 5 duplicate definitions.
- The in-container `stat -c` in 025 (line 51) correctly left alone.
- `INTEGRATION_DIR` is absolute, so the PATH prepend is safe regardless of test cwd changes.
- Opt-in `# itest: no-docker` markers over inference — right call, for exactly the reason the comment gives.

## Stacked (#239) note

In `run.sh`, `tests=("${nodocker[@]}")` with an empty array is an "unbound variable" error under `set -u` on bash 3.2 (macOS) — the `#tests -eq 0` guard below it would never be reached. Latent today (11 marked tests exist), but worth `tests=("${nodocker[@]+"${nodocker[@]}"}")` or a pre-check if #239 is still open for edits.

integration/common.sh
```diff
@@ -601,3 +601,20 @@ tui_assert_not_contains() {
...
+server_count() {
+  { pgrep -f "${FLEET_BIN} server" 2>/dev/null || true; } | wc -l | tr -d ' '
+}
+
+# file_mode <path> — octal permission bits of a HOST file, portably. GNU
+# stat spells it -c '%a', BSD/macOS -f '%Lp'; probing GNU first keeps Linux
+# behavior byte-identical. (In-container stat calls stay plain `stat -c` —
+# containers are Linux regardless of the host.)
+file_mode() {
+  stat -c '%a' "$1" 2>/dev/null || stat -f '%Lp' "$1"
```
BenjaminBenetti: 🔍 **CodeReview Agent**

This is the line failing `025_copy` in the `integration-bsd 0-100` shard on this head: under shims, `stat -c` is rejected by the shim (as intended), but the `stat -f '%Lp'` fallback is **delegated to GNU stat**, where `-f` means `--file-system` and `%Lp` becomes a file operand:

```
/usr/bin/stat: cannot read file system information for '%Lp': No such file or directory
```

So on a shimmed Linux host neither spelling works, and under `set -euo pipefail` the `mode=$(file_mode …)` assignment kills the test. The helper itself is correct for real Linux and real macOS — the fix belongs in the stat shim (see comment there). Alternative if you'd rather not extend the shim: sidestep stat entirely with a spelling identical on both platforms, e.g. `perl -e 'printf "%03o", (stat shift)[2] & 0777' "$1"` — but shim-side emulation keeps this helper exercising the actual macOS code path, which is the job's purpose.

integration/bsd-shims/stat
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# BSD stat has no -c/--format/--printf (it spells formatting -f).
+for arg in "$@"; do
+  case "$arg" in
+    -c|--format|--format=*|--printf|--printf=*)
```
BenjaminBenetti, line 5: 🔍 **CodeReview Agent**

The shim rejects the GNU spelling but doesn't *accept* the BSD spelling — `stat -f '%Lp'` falls through to GNU stat, where `-f` is `--file-system` and the format string becomes a file operand. Net effect: portable "try GNU, fall back to BSD" code (like the new `file_mode` in common.sh, currently failing shard 0-100) has no working spelling under shims, which is a false positive relative to a real Mac.

Suggest emulating `-f` for the formats the suite needs, and failing loudly on the rest, before the delegation block:

```sh
if [ "${1:-}" = "-f" ]; then
  case "${2:-}" in
    '%Lp') shift 2; set -- -c '%a' "$@" ;;   # then fall through to exec "$real"
    *) echo "stat shim: unhandled BSD format: ${2:-}" >&2; exit 1 ;;
  esac
fi
```

Same latent class in the `date` shim: BSD-only `-r <epoch>` / `-v` delegate to GNU date, where `-r` means `--reference=FILE` — nothing uses them today, but worth either emulating or rejecting explicitly so the failure is self-explanatory when something does.

integration/common.sh
shawn-vincent-well: Fixed in the current head: file_mode no longer uses stat at all -- it is a perl one-liner ((stat)[2] & 07777), which means the same thing on GNU, BSD, and under the shims. Verified plain and shimmed on macOS, and the integration-bsd run is green on the reworked head.

integration/bsd-shims/stat
shawn-vincent-well: Good catch -- fixed in 484b39a. The stat shim now emulates BSD -f for the formats the suite uses (%Lp -> -c %a, %z -> -c %s) when the real stat behind it is GNU, and fails loudly naming the format for anything unmapped. When the real stat is BSD (running the shims on an actual Mac) -f falls through to native delegation. Verified both branches: native passthrough on macOS, translated emulation in an ubuntu container.

BenjaminBenetti:
🧪 **QA Agent** — Manual QA

| Test | Result | Notes |
|------|--------|-------|
| Build + `go test ./...` at PR head | ✅ Pass | 40 packages ok, 0 fail; `go vet ./...` clean; modified `cli`/`control`/`server` test packages pass |
| Shim semantics (`sleep`/`stat`/`date`/`timeout`) | ✅ Pass | `date -d/--date=` → exit 1; `+%s`/`-u` delegate; `timeout` → exit 127 "command not found" (faithful to macOS having no `timeout(1)`) |
| Mechanism is a **real** test, not a no-op | ✅ Pass | Throwaway canary reproducing the #231 `sleep infinity` bug: **without** shims → PASS/exit 0 (Linux hides it); **with** shims → FAIL/exit 1 with correct attribution. Canary deleted, tree clean. |
| #231 fix present & shim-compatible | ✅ Pass | `internal/tui/splitpane.go:210` uses pure-numeric `placeholderSleepSecs = "2147483647"`; the `file_mode` helper (PR's first catch, replacing `stat -c`) returns correct octal (755/644/600) under shims |
| `run.sh` harness under shims | ✅ Pass | No-docker filter selects 11 tests → 11/11 pass under BSD shims (isolated HOME/TMUX); docker-backed 010/020 pass under shims locally |
| Workflow structure | ✅ Pass | `integration-bsd.yml` mirrors stock `integration.yml` (identical 7-shard matrix, same `needs`+`always()`+`result != success` aggregate gate); only delta is `FLEET_ITEST_BSD_SHIMS: "1"`; no shard gaps/overlaps |
| Real GitHub CI (strongest evidence) | ✅ Pass | All 7 `integration-bsd` shards + the gate pass, plus `integration-macos-no-docker` and `unit-macos` |
| macOS-breakage coverage completeness | ⚠️ Concern | Two definite macOS breaks (`sha256sum`, `mapfile -t`) are silently missed by the shim job while it shows green — see below |

**Overall: ✅ PASS (merge-safe) — 7/7 core scenarios pass, zero regressions; 2 MEDIUM coverage gaps flagged for non-blocking follow-up.** The mechanism is correct, proven, and green on real CI. The concerns are scope/coverage caveats, not reasons to hold the PR.

## Environment

Manual QA at head `484b39a` in a Linux devcontainer (GNU userland, like `ubuntu-latest`). Shims exercised exactly as `run.sh` activates them (`PATH=integration/bsd-shims:$PATH`); `go test ./...` run natively; and the live GitHub CI results for this head cross-checked (`mergeStateStatus: BLOCKED` is branch-protection/stacking, not a CI failure).

## ⚠️ MEDIUM — `sha256sum` unshimmed (test 380): green here, `command not found` on a real Mac

`integration/tests/380_tui_admiral_skill_install.sh:51` hashes a host file (`$HOME/.claude/skills/.../SKILL.md`) with `sha256sum`. macOS has no `sha256sum` — it's `shasum -a 256`. It isn't shimmed, and 380 is **not** marked `# itest: no-docker`, so it runs only under the new BSD-shim job (green on Linux) and never on the mac-native job → it would `command not found` on a real Mac while CI stays green.

## ⚠️ MEDIUM — `mapfile -t` unshimmed (test 060): bash-4 builtin on bash-3.2 macOS

`integration/tests/060_destroy.sh:13` uses `mapfile -t`, a bash-4 builtin; macOS ships bash 3.2. It's structurally un-shimmable (a builtin, not a PATH command), so the shim job can't catch it. This is inconsistent with `run.sh`'s own deliberate bash-3.2 avoidance (it dropped `declare -A` for exactly this reason).

Both MEDIUMs are **pre-existing test-portability bugs, not regressions from this PR** — but they matter because the deliverable's whole purpose is catching macOS breakage, and it silently misses two definite macOS breaks while presenting a green "full suite, macOS semantics" signal. Non-blocking follow-up: add a `sha256sum`→`shasum` shim + guard/replace `mapfile` (or mark those tests), and document that the shim job is a *common-class* gate, not a complete macOS gate.

## Lower-severity notes

- **LOW** — `pgrep`/`pkill` are unshimmed. Current flags (`-f`/`-x`/`-9`) are macOS-portable, but this surface already bit the repo (`pgrep -fc`) and has zero shim coverage — a future GNU-only flag would pass silently.
- **LOW** — the `sleep` shim accepts unit suffixes (`5s`/`5m`/`5h`/`5d`) that older BSD `sleep` rejects (a deliberate tradeoff to avoid false-positives on newer macOS). Harmless today: nothing host-side uses suffixed sleeps. It would also wrongly reject `sleep -- 5`, but nothing uses that host-side either.
- **INFO** — CI cost roughly doubles: a second 7-shard docker matrix per PR (~2–6 min/shard). The author flags this and offers dial-downs (schedule / path-filter).
- **INFO** — `025_copy` fails inside the nested-docker QA sandbox (fails **without** shims too — the copy invocation is byte-identical to `main`), so the `file_mode` fix was validated at the unit level instead of end-to-end. Not a PR issue; it passes on real CI.

## ✅ What works

- **Shim behavior** matches intent: BSD-incompatible invocations are rejected (correct exit codes/usage), compatible ones delegate to the real binary; delegation lookup correctly skips the shim dir.
- **Failures surface** — the canary proves the shim job genuinely fails (and attributes correctly) when a macOS-incompatible command is used, rather than passing as a no-op.
- **No regressions** — full `go test ./...` green, `go vet` clean.
- **Workflow** faithfully mirrors the stock integration matrix and aggregate gate; the only difference is the `FLEET_ITEST_BSD_SHIMS` env flag.
- **Real CI** is green across all `integration-bsd` shards, the gate, and the macOS jobs.

Solid technique — the shims are correct and the mechanism is proven. The only thing worth a follow-up is closing the two coverage gaps (`sha256sum`, `mapfile`) so the "macOS semantics" signal is complete, not just common-class.

PR #242, opened 2026-07-21
feat(release): build macOS binaries on tag + OS-aware installer

BenjaminBenetti:
Closes #230

## What
- **release.yml**: also build `fleet-darwin-amd64` and `fleet-darwin-arm64`
  alongside the Linux binaries. They flow through the existing `sha256sum
  fleet-*` and `gh release create dist/*` steps, so checksums and the release
  upload cover them with no extra wiring.
- **install.sh**: replace the Linux-only guard with a `uname -s` case
  (linux/darwin), so the installer downloads the matching `fleet-${OS}-${ARCH}`
  asset for the user's machine (Apple Silicon → darwin-arm64, Intel Mac →
  darwin-amd64). Unsupported OSes still fail with a clear message.
- **README.md**: Requirements now list macOS.

## Notes
- macOS binaries are unsigned; `install.sh` downloads via curl (no
  `com.apple.quarantine`), so Gatekeeper won't block running `fleet` from a
  terminal.
- Both darwin targets verified to cross-compile locally.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,8 +36,10 @@ jobs:
           version="${GITHUB_REF_NAME}"
           ldflags="-s -w -X github.com/BenjaminBenetti/fleet-man/internal/version.Version=${version}"
 
-          GOOS=linux GOARCH=amd64 go build -ldflags "${ldflags}" -o dist/fleet-linux-amd64 ./cmd/fleet
-          GOOS=linux GOARCH=arm64 go build -ldflags "${ldflags}" -o dist/fleet-linux-arm64 ./cmd/fleet
+          GOOS=linux  GOARCH=amd64 go build -ldflags "${ldflags}" -o dist/fleet-linux-amd64 ./cmd/fleet
+          GOOS=linux  GOARCH=arm64 go build -ldflags "${ldflags}" -o dist/fleet-linux-arm64 ./cmd/fleet
+          GOOS=darwin GOARCH=amd64 go build -ldflags "${ldflags}" -o dist/fleet-darwin-amd64 ./cmd/fleet
+          GOOS=darwin GOARCH=arm64 go build -ldflags "${ldflags}" -o dist/fleet-darwin-arm64 ./cmd/fleet
 
           cd dist
           sha256sum fleet-* > checksums.txt
--- a/install.sh
+++ b/install.sh
@@ -31,11 +31,18 @@ case "$ARCH" in
         ;;
 esac
 
+# Detect OS and map it to the release asset's platform token. Releases ship a
+# binary per OS/arch pair (fleet-linux-amd64, fleet-darwin-arm64, ...); pick the
+# one matching this machine.
 OS=$(uname -s | tr '[:upper:]' '[:lower:]')
-if [ "$OS" != "linux" ]; then
-    echo "Error: unsupported OS: $OS (only linux is supported)"
-    exit 1
-fi
+case "$OS" in
+    linux)         OS="linux" ;;
+    darwin)        OS="darwin" ;;
+    *)
+        echo "Error: unsupported OS: $OS (supported: linux, darwin)"
+        exit 1
+        ;;
+esac
 
 ASSET="fleet-${OS}-${ARCH}"
 
--- a/install.sh
+++ b/install.sh
@@ -77,8 +77,10 @@ chmod +x "$TMP"
 
 # Ensure install directory exists (fresh macOS installs may not have /usr/local/bin)
 if [ ! -d "$INSTALL_DIR" ]; then
-    echo "Creating ${INSTALL_DIR} (requires sudo)..."
-    sudo mkdir -p "$INSTALL_DIR"
+    if ! mkdir -p "$INSTALL_DIR" 2>/dev/null; then
+        echo "Creating ${INSTALL_DIR} (requires sudo)..."
+        sudo mkdir -p "$INSTALL_DIR"
+    fi
 fi
 
 # Install — use sudo if needed
--- a/install.sh
+++ b/install.sh
@@ -75,6 +75,12 @@ fi
 
 chmod +x "$TMP"
 
+# Ensure install directory exists (fresh macOS installs may not have /usr/local/bin)
+if [ ! -d "$INSTALL_DIR" ]; then
+    echo "Creating ${INSTALL_DIR} (requires sudo)..."
+    sudo mkdir -p "$INSTALL_DIR"
+fi
+
 # Install — use sudo if needed
 if [ -w "$INSTALL_DIR" ]; then
     mv "$TMP" "${INSTALL_DIR}/${BINARY}"
--- a/README.md
+++ b/README.md
@@ -490,7 +490,7 @@ location), `TMUX` (enables split-pane mode when run inside tmux), `SSH_AUTH_SOCK
 
 ## Requirements
 
-- Linux, including Ubuntu on WSL2
+- Linux (including Ubuntu on WSL2) or macOS
 - Docker
 - [devcontainer CLI](https://github.com/devcontainers/cli) (`npm install -g @devcontainers/cli`)
 


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed `8dfeb3a` (release.yml + install.sh + README). The change does what it says, and the plumbing claims check out:

- ✅ The darwin binaries flow through the existing `sha256sum fleet-*` and `gh release create dist/*` steps — checksums.txt and the release upload cover them with no extra wiring.
- ✅ Cross-compiling darwin from the ubuntu runner implicitly sets `CGO_ENABLED=0`; the repo has no cgo or `//go:build` OS-tagged files, so this is safe. Note the Go linker ad-hoc-signs `darwin/arm64` binaries at link time, which is why these unsigned cross-compiled binaries run at all on Apple Silicon — the Gatekeeper/quarantine note in the PR description holds for curl downloads.
- ✅ `uname -m`/`uname -s` mapping is correct for Intel (`x86_64`→amd64) and Apple Silicon (`arm64`).

**Findings (2, both actionable):**

1. **`install.sh` will fail on a fresh macOS install — `/usr/local/bin` may not exist.** Details inline. This is the main gap: the happy path for a brand-new Apple Silicon Mac (the audience of this PR) hits it.

2. **Darwin compile-ability is only exercised at tag time.** Details inline. A one-line guard in PR CI would catch darwin-breaking changes before they break a release.

**Minor (no action needed):** under Rosetta 2 (installer run from an x86_64 shell on Apple Silicon), `uname -m` reports `x86_64`, so the amd64 binary gets installed. It works via Rosetta, just not natively. If you ever care, `sysctl -in sysctl.proc_translated` detects this — fine to leave as-is.

README/Requirements update is consistent with the change. No security concerns beyond the pre-existing lack of checksum verification in `install.sh` (unchanged by this PR, and checksums.txt now covers darwin if you ever add it).

install.sh
```diff
@@ -31,11 +31,18 @@ case "$ARCH" in
...
+# binary per OS/arch pair (fleet-linux-amd64, fleet-darwin-arm64, ...); pick the
+# one matching this machine.
 OS=$(uname -s | tr '[:upper:]' '[:lower:]')
-if [ "$OS" != "linux" ]; then
-    echo "Error: unsupported OS: $OS (only linux is supported)"
-    exit 1
-fi
+case "$OS" in
+    linux)         OS="linux" ;;
+    darwin)        OS="darwin" ;;
```
BenjaminBenetti, line 40: 🔍 **CodeReview Agent**

Now that darwin is accepted here, the install step further down (`mv "$TMP" "${INSTALL_DIR}/${BINARY}"`) has a macOS edge case: **`/usr/local/bin` does not exist on a fresh macOS install** — on Apple Silicon, Homebrew lives in `/opt/homebrew` and never creates it. A nonexistent dir fails `[ -w "$INSTALL_DIR" ]`, so the script drops into the sudo branch and `sudo mv` dies with a confusing `No such file or directory`.

Suggest ensuring the directory exists before the move, e.g.:

```sh
if [ ! -d "$INSTALL_DIR" ]; then
    echo "Creating ${INSTALL_DIR} (requires sudo)..."
    sudo mkdir -p "$INSTALL_DIR"
fi
```

(`/usr/local/bin` is already in macOS's default PATH via `/etc/paths`, so once created everything works.)

.github/workflows/release.yml
```diff
@@ -36,8 +36,10 @@ jobs:
           version="${GITHUB_REF_NAME}"
           ldflags="-s -w -X github.com/BenjaminBenetti/fleet-man/internal/version.Version=${version}"
 
-          GOOS=linux GOARCH=amd64 go build -ldflags "${ldflags}" -o dist/fleet-linux-amd64 ./cmd/fleet
-          GOOS=linux GOARCH=arm64 go build -ldflags "${ldflags}" -o dist/fleet-linux-arm64 ./cmd/fleet
+          GOOS=linux  GOARCH=amd64 go build -ldflags "${ldflags}" -o dist/fleet-linux-amd64 ./cmd/fleet
+          GOOS=linux  GOARCH=arm64 go build -ldflags "${ldflags}" -o dist/fleet-linux-arm64 ./cmd/fleet
+          GOOS=darwin GOARCH=amd64 go build -ldflags "${ldflags}" -o dist/fleet-darwin-amd64 ./cmd/fleet
+          GOOS=darwin GOARCH=arm64 go build -ldflags "${ldflags}" -o dist/fleet-darwin-arm64 ./cmd/fleet
```
BenjaminBenetti, line 42: 🔍 **CodeReview Agent**

These darwin builds only run on tag push, so the first signal that a change broke darwin cross-compilation (a new dep with cgo, an OS-specific syscall, a `//go:build linux` file) would be a failed **release**. Unit CI (`unit.yml`) vets/tests on ubuntu only and never compiles for darwin.

Cheap insurance — one step in `unit.yml`:

```yaml
- name: Cross-compile check (darwin)
  run: |
    GOOS=darwin GOARCH=arm64 go build ./...
    GOOS=darwin GOARCH=amd64 go build ./...
```

Adds a few seconds per PR and moves the failure from tag time to PR time.

install.sh
BenjaminBenetti: @copilot+claude-opus-4.8  Address the /usr/loca/bin issue

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Re-review of the new commits (`4d4d9aa`, `0f7e9f1b`) at head `0f7e9f1b`.

**✅ Prior feedback resolved.** The new "ensure install directory exists" block in `install.sh` fixes the fresh-macOS case where `/usr/local/bin` doesn't exist, and the `0f7e9f1b` refinement (try unprivileged `mkdir -p` first, fall back to `sudo` only on failure) is the right ordering. Edge cases I checked all behave correctly:

- Fresh macOS: `/usr/local` is root-owned → unprivileged `mkdir` fails → sudo path with a clear message. After `sudo mkdir`, the dir is root-owned, so `[ -w ]` fails and the existing `sudo mv` branch handles the install consistently.
- Root in a minimal container (no `sudo` binary): unprivileged `mkdir -p` succeeds, sudo is never invoked.
- Linux default path: `/usr/local/bin` exists, block is skipped — no behavior change.
- With `set -e`, a failed `sudo mkdir` aborts the script instead of falling through to a confusing `mv` error.

**Nit (non-blocking, pre-existing):** if the script aborts at the `sudo mkdir`/`sudo mv` steps, the downloaded binary is left in `$TMP`. A one-liner near the `mktemp` would tidy that up (harmless no-op after a successful `mv`):

```sh
trap 'rm -f "$TMP"' EXIT
```

**Still open from my previous review (non-blocking):** `unit.yml` remains ubuntu-only with no darwin cross-compile check, so a change that breaks darwin builds will first surface as a failed tag release. The two-line `GOOS=darwin go build ./...` step is cheap insurance.

No new blocking issues — the change is in good shape.

install.sh
BenjaminBenetti: 🔍 **CodeReview Agent**

Confirmed resolved at head `0f7e9f1b`. The new block creates `$INSTALL_DIR` before the move, and the follow-up refinement (unprivileged `mkdir -p` first, `sudo` only on failure) improves on my original suggestion — it avoids an unnecessary sudo prompt when the parent is user-writable and works when running as root without a `sudo` binary. Post-`sudo mkdir` ownership (root, 0755) correctly routes the install through the existing `sudo mv` branch.

BenjaminBenetti: 🔍 **CodeReview Agent**

Verified fixed as of `0f7e9f1`. The flow is coherent end-to-end: unprivileged `mkdir -p` is attempted first, sudo only on failure — and when sudo creates the dir root-owned, the subsequent `[ -w "$INSTALL_DIR" ]` check fails and correctly routes the install through the existing `sudo mv` branch. The bare `mkdir` inside the `if !` guard is also safe under `set -e`. Resolving.

BenjaminBenetti (commented):
🔍 **CodeReview Agent**

**Re-review of head `0f7e9f1`** (new commits `4d4d9aa`, `0f7e9f1` since last pass)

✅ **Missing-`/usr/local/bin` fix verified.** The new block in `install.sh` is correct:
- Unprivileged `mkdir -p` is tried first; sudo is a fallback only when it fails — so root/container and user-writable-`/usr/local` cases never prompt.
- When `sudo mkdir` does create the dir (root-owned), the downstream `[ -w "$INSTALL_DIR" ]` check fails and the install correctly flows through the existing `sudo mv` branch.
- The bare `mkdir` sits inside an `if !` guard, so `set -e` won't abort prematurely; a genuine `sudo mkdir` failure still exits the script, as it should.

**Minor (non-blocking):** if `$INSTALL_DIR` somehow exists as a regular *file*, both mkdirs fail and the script dies with a slightly misleading `(requires sudo)` message. Vanishingly unlikely; fine to leave as-is.

**Carried over from previous review (still open, non-blocking):** darwin cross-compilation is only exercised on tag push. A `GOOS=darwin go build ./...` smoke step in `unit.yml` would surface darwin breakage (cgo deps, `//go:build linux` files, OS-specific syscalls) at PR time instead of at release time.

No issues with the `release.yml` additions (darwin targets flow through the existing checksum/upload steps as described) or the README update. Nothing blocking merge.

PR #243, opened 2026-07-22
fix(devcontainer): harden FLEET_SSH_AGENT_SOCK parsing (post-merge review of #236)

shawn-vincent-well:
Addresses the two review threads left on #236 after merge.

1. Kill-switch matching: the off/none check was exact-case and untrimmed, unlike the other FLEET_* env parsers (devcontainerUpArgs, devcontainerEnv both TrimSpace). FLEET_SSH_AGENT_SOCK=OFF, or 'off ' with stray whitespace from a shell rc, fell through to the path branch and was bind-mounted verbatim -- reproducing exactly the confusing 'bind source path does not exist' failure class #236 fixed. Now TrimSpace + EqualFold.

2. Malformed overrides: a relative path or a comma-containing value ('/run/foo.sock,readonly') was spliced verbatim into the --mount string, where a comma injects extra mount options. Unusable values are now a hard error naming the env var, raised at instance create/clone time, matching the repo's other env parser conventions. Exec paths degrade to no forwarding instead of blocking shells into an existing container whose creation already validated the value.

Tests cover OFF / ' None ' / 'off ' (disable), relative / comma / non-path values (error), and the existing linux/darwin/override matrix is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,7 +470,7 @@ Variables fleet **reads** (set them to configure behavior):
 | `FLEET_SERVER` | `host:port` | Drive a remote daemon over plain TCP (no gateway). |
 | `FLEET_DEVCONTAINER_BUILDKIT` | `auto` (default), `never` | BuildKit mode for Fleet-managed devcontainers. See [Devcontainer BuildKit](#devcontainer-buildkit). |
 | `FLEET_DEVCONTAINER_UPDATE_REMOTE_USER_UID` | `default`, `never`, `on`, `off` | Remote-user UID/GID rewrite mode. See [Devcontainer UID Rewrite](#devcontainer-uid-rewrite). |
-| `FLEET_SSH_AGENT_SOCK` | path, `off`, or `none` | Override the bind source for SSH agent forwarding into instances (`off`/`none` disables it). On macOS the default is Docker Desktop's VM-side `/run/host-services/ssh-auth.sock` (OrbStack and `colima --ssh-agent` are path-compatible); set this if your Docker backend exposes the agent elsewhere (default Colima, Podman machine, Rancher Desktop). |
+| `FLEET_SSH_AGENT_SOCK` | absolute path, `off`, or `none` (case-insensitive) | Override the bind source for SSH agent forwarding into instances (`off`/`none` disables it). On macOS the default is Docker Desktop's VM-side `/run/host-services/ssh-auth.sock` (OrbStack and `colima --ssh-agent` are path-compatible); set this if your Docker backend exposes the agent elsewhere (default Colima, Podman machine, Rancher Desktop). |
 | `CODER_URL` | URL | Coder deployment URL (Coder backend). |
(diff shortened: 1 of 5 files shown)


BenjaminBenetti (commented):
🔍 **CodeReview Agent**

Reviewed the post-merge hardening for `FLEET_SSH_AGENT_SOCK`. This is a tight, well-scoped fix — the parsing now matches the repo's other `FLEET_*` conventions (`TrimSpace` + `EqualFold`, and the `invalid %s value %q (valid: …)` hard-error phrasing is identical to `devcontainerUpArgs` / `devcontainerEnv`). The comma-injection and case/whitespace kill-switch bugs are both real and correctly closed, and the test matrix (`OFF` / `" None "` / `"off "` disable, relative/comma/non-path error) covers the important cases. The `docker rmi` cleanup added to the new `buildCloneRunArgs` error path in `clone.go` correctly mirrors the existing `docker run`-failure cleanup, so a failed clone doesn't leak the committed image. Nice work.

A few minor / optional observations — none blocking:

**1. `sshExecArgs` silently swallows the validation error (`ssh.go`).**
```go
sock, err := sshAgentMountSource()
if err != nil || sock == "" {
    return nil
}
```
The `err != nil` collapse into the "no forwarding" path is a deliberate, documented tradeoff (don't block a shell into an already-created container), which is reasonable. The papercut is that it's *silent*: if a user typos the env var between create and exec (e.g. `…/agent.sock,ro`, or a relative path), `up`/`clone` hard-fail with a clear message but `exec` just drops `SSH_AUTH_SOCK` with zero diagnostic — even though the container's mount (created when the value was valid) is physically fine. Consider a one-line `fmt.Fprintf(os.Stderr, …)` warning on the `err != nil` branch so the lost forwarding is debuggable. Behavior stays the same; only the silence changes.

**2. `filepath.IsAbs` is host-OS–specific (`ssh.go`).**
The override is explicitly documented as "a path that may only exist inside the Docker VM," i.e. a POSIX path. `filepath.IsAbs` uses the semantics of the OS fleet runs on, so on Windows a legitimate VM path like `/run/host-services/ssh-auth.sock` would be rejected as "not absolute." Almost certainly out of scope (the code only special-cases `linux`/`darwin` and I see no Windows support), but flagging since the validator's notion of "absolute" and the socket's notion aren't the same on Windows. `strings.HasPrefix(override, "/")` would sidestep it if Windows ever matters.

**3. Small positive-path test gap (`ssh_test.go`).**
`TestSSHUpArgs_OverrideCaseAndWhitespace` proves whitespace is trimmed on the *disable* path, but there's no assertion that a valid override *with* surrounding whitespace (e.g. `" /run/foo.sock "`) is trimmed and accepted as the mount source. Adding that one case would lock in the trim behavior for the branch users are most likely to hit.

Logic-wise the override/darwin precedence is correct (a valid override wins over the darwin remap), and the validation/`sshAgentMountSourceFor` split keeps the pure core testable. No correctness concerns — just the notes above.

